PR #1, opened 2016-02-25
Added devise rspec support

mphprogrammer:


mphprogrammer:
@plusjeff Travis does not seem to be working with this commit.. Tried configuring, but no dice...

PR #2, opened 2016-02-25
Added devise rspec support, configure travis take 2 ...

mphprogrammer:


mphprogrammer:
For some reason, travis is not appearing. Added config items to travis.yml, and issued new PR but no luck.

PR #4, opened 2016-02-25
Added shared controller specs and application helpers

mphprogrammer:


jeffmaher:
Looks good to me! Merged it with `master` and unit tests / Heroku are re-running. OK to merge if those come back fine.

PR #6, opened 2016-03-01
Sources

mphprogrammer:
Added sources resource and specs.


app/controllers/raw_file_sources_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+	end
+
+	def edit
+		respond_to do |format|
+			format.html
+		end	
+	end
+
+	def update
+    respond_to do |format|
```
plusjeff: Nitpick: Formatting seems off here. Is the linter running? (I think it'll auto correct this)

app/views/raw_file_sources/new.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1>Raw File Sources</h1>
+<p>
+	This are the allowed raw file sources for the GIBCT. These sources reflect
```
plusjeff: "These" are...

app/views/raw_file_sources/show.html.erb
```diff
@@ -0,0 +1,50 @@
+<div class="row">
+	<div class="col-xs-8">
+		This are the allowed raw file sources for the GIBCT. These sources reflect
```
plusjeff: "These" are ...

spec/controllers/raw_file_sources_controller_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+	describe "GET new" do
+		login_user
+		render_views
+
+		before(:each) do
+			create :raw_file_source
+			get :new
+		end
+
+		it "renders the new page" do
```
plusjeff: Does this test just check to see if the response type is HTML? Does it need to be checking other things to verify that data is being pulled correctly?

jeffmaher:
Could you update the README to describe setup? A possible template is:

https://github.com/department-of-veterans-affairs/roadrunner-rails/blob/master/README.template.md

jeffmaher:
From a user standpoint, what does someone do with this screen? Could you describe what will happen in the end system when someone adds a new source?

image

jeffmaher:
Bouncing back for feedback and responses. Toss back my way when done.

mphprogrammer:
@plusjeff That screen is for me to set the build order, for each data source type. The order in which the csv's are processed and built in the database will matter. It (the build order) may change as things change down the road.

mphprogrammer:
@plusjeff The "GET new" spec just checks whether or not a page is rendered and that it contains a form. There really is no other functionality. The new controller creates an empty raw_file_source instance for the form_for helper in the view.

mphprogrammer:
@plusjeff I will update the read me in a separate pull (coming in a few minutes). The seeds.rb has login information (it needs to create a devise logins for development uses on staging and locally, and eventually production) and cannot be put onto github. I will write a simple rake task to generate a shell seeds file without the login information an put instructions to run the task and add their own logins.

mphprogrammer:
@plusjeff I ran brakeman locally, and it complained about ActionPack 4.2.5.1, suggesting that I update to 4.2.5.2 - I will do so by using rails 4.2.5.2 (currently 4.2.5.1).

PR #9, opened 2016-03-01
Fixed issue with scanners not running

jeffmaher:
@mphprogrammer Thanks for catching this earlier today. Looks like the Travis configuration was not running the code scans.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,5 @@
 AllCops:
+  TargetRubyVersion: 2.3
   Include:
     - '**/config.ru'
     - '**/Rakefile'
@@ -15,6 +16,9 @@ Style/StringLiterals:
 Style/Documentation:
   Enabled: false
 
+Style/ClassAndModuleChildren:
+  Enabled: false
+
 Metrics/LineLength:
   Max: 100
 
@@ -24,3 +28,9 @@ Rails:
 Rails/Output:
   Exclude:
     - 'lib/**/*.rb'
+
+Metrics/MethodLength:
+  Max: 20
+
+Metrics/ClassLength:
+  Max: 400
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ branches:
 
 script:
   - git branch
-  - bundle exec rspec
+  - bundle exec rake
 
 bundler_args: "--without development"
 cache:
--- a/app/controllers/concerns/alertable.rb
+++ b/app/controllers/concerns/alertable.rb
@@ -1,24 +1,24 @@
 module Alertable
-	extend ActiveSupport::Concern
+  extend ActiveSupport::Concern
 
-	included do
-	end
+  included do
+  end
 
-	class_methods do 
-		###########################################################################
-		## pretty_error
-		## Wraps an array of messages (presumably errors) in a list.
-		###########################################################################
-		def pretty_error(label = "", errors = [])
-			msg = errors.inject("<ul>") do |m, error|
-				m += "<li>#{error}</li>"
-			end + "</ul>"
+  class_methods do
+    ###########################################################################
+    ## pretty_error
+    ## Wraps an array of messages (presumably errors) in a list.
+    ###########################################################################
+    def pretty_error(label = "", errors = [])
+      msg = errors.inject("<ul>") do |m, error|
+        m += "<li>#{error}</li>"
+      end + "</ul>"
 
-			pstr = ""
-			pstr += "<p>#{label}</p>" if label.present?
-			pstr += msg if errors.present?
+      pstr = ""
+      pstr += "<p>#{label}</p>" if label.present?
+      pstr += msg if errors.present?
 
-			pstr
-		end
-	end
+      pstr
+    end
+  end
 end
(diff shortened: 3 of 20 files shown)


jeffmaher:
The Travis configuration file is the only file that has substantially changed. The Ruby code that changed was auto-updated by Rubocop.

jeffmaher:
There are still issues with this in that JSHint is scanning bootstrap.

jeffmaher:
Here are the remaining offenses. Handing over to you @mphprogrammer 

```
Inspecting 35 files
...W.WW...............WC......C...C

Offenses:

app/controllers/concerns/alertable.rb:14:9: W: Useless assignment to variable - m. Use just operator +.
        m += "<li>#{error}</li>"
        ^
app/controllers/raw_file_sources_controller.rb:29:3: C: Method has too many lines. [12/10]
  def create
  ^^^
app/controllers/raw_file_sources_controller.rb:42:8: W: end at 42, 7 is not aligned with if at 34, 6.
       end
       ^^^
app/helpers/application_helper.rb:9:27: C: Use %r around regular expression.
    (fullpath || "").scan(/^[\w+|\/]+/)[0]
                          ^^^^^^^^^^^^
app/helpers/application_helper.rb:34:5: W: Useless assignment to variable - l. Use just operator +.
    l += "</a></li>"
    ^
app/helpers/application_helper.rb:42:3: C: Assignment Branch Condition size for draw_controller is too high. [27.37/15]
  def draw_controller(controller, *actions)
  ^^^
app/helpers/application_helper.rb:42:3: C: Method has too many lines. [19/10]
  def draw_controller(controller, *actions)
  ^^^
spec/controllers/raw_file_sources_controller_spec.rb:29:38: W: Ambiguous regexp literal. Parenthesize the method arguments if it's surely a regexp literal, or add a whitespace to the right of the / if it should be a division.
      expect(response.body).to match /file_source_/im
                                     ^
spec/controllers/raw_file_sources_controller_spec.rb:48:38: W: Ambiguous regexp literal. Parenthesize the method arguments if it's surely a regexp literal, or add a whitespace to the right of the / if it should be a division.
      expect(response.body).to match /create a new raw file source/im
                                     ^
spec/factories/raw_file_sources.rb:3:20: C: Prefer {...} over do...end for single-line blocks.
    sequence :name do |i| "file_source_#{i}" end
                   ^^
spec/rails_helper.rb:62:3: C: Use only ascii symbols in comments.
  # transactions won’t work, so this code overrides the setting and
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/rails_helper.rb:63:3: C: Use only ascii symbols in comments.
  # chooses the “truncation” strategy instead. (MPH)
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/support/shared_examples_for_authentication.rb:39:16: C: Add parentheses to nested method call get action.
        expect(get action).to render_template("#{destination}/#{action}")
               ^^^^^^^^^^

35 files inspected, 13 offenses detected
```

mphprogrammer:
@plusjeff - there seems to be a cyclical error with the linter -- Given this code, that uses a ruby inject to build a string: 

def pretty_error(label = "", errors = [])
  msg = errors.inject("<ul>") do |m, error|
    m += "<li>#{error}</li>" # Linter error to use +=
  end + "</ul>"

The linter produces the following ...

app/controllers/concerns/alertable.rb:14:9: W: Useless assignment to variable - m. Use just operator +.
       m += "<li>#{error}</li>" # Linter error to use +=
        ^
Fixing that to 

```
def pretty_error(label = "", errors = [])
  msg = errors.inject("<ul>") do |m, error|
    m = m + "<li>#{error}</li>" # Linter error to use +=
  end + "</ul>"
```

The linter produces  another error 

app/controllers/concerns/alertable.rb:14:9: C: [Corrected] Use self-assignment shorthand +=.
        m = m + "<li>#{error}</li>" # Linter error to use +=
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
and corrects it back to the original code

jeffmaher:
@mphprogrammer I committed a fix against the mentioned line. I don't think there needs to be any kind of assignment in that statement, since inject will take care of that:

``` ruby
      msg = errors.inject("<ul>") do |m, error|
        m + "<li>#{error}</li>"
      end
```

http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-inject

jeffmaher:
We're also likely to be updated the linting configuration based on Caseflow's revisions:

https://github.com/department-of-veterans-affairs/roadrunner/issues/34

Moved Caseflow's configuration into this, particularly for the lines per method restrictions (moves from 10 to 20 lines).

jeffmaher:
I'll be offsite tomorrow, but will pass back to you for the rest. Happy to review on Thursday morning.

mphprogrammer:
DOH! Sometimes I get locked in a train a thought and when I read what I coded it doesn't register!

jeffmaher:
Heh - no worries. Also, I think I lopped off your closing `</ul>`. D'oh! On my end.

jeffmaher:
This change looks like it found it's way in (?). Closing.

PR #83, opened 2016-05-25
Complaints fix

mphprogrammer:


mphprogrammer:
I suspect the bug patrick found was due to the DS using the complaints ope6 as the key, which he has stated is unreliable. I now first update the complaints ope with the crosswalk ope and then take ope roll up sums. This is done in the data_csv build so that csv load order independence is preserved.

PR #104, opened 2016-06-08
Removed most case conversions when storing data

mphprogrammer:


mphprogrammer:
All case conversions have been removed (except institution and facility code, which are uppercase). As the data appears in the input CSVs, that is how they will appear in the data.csv.

saintsoup52:
Thank you!

PR #111, opened 2016-06-24
Uppercase 1

mphprogrammer:
Added uppercase institution types to export csv and uppercase accreditation types to Accreditation.

### DIFF
--- a/app/models/accreditation.rb
+++ b/app/models/accreditation.rb
@@ -2,10 +2,10 @@ class Accreditation < ActiveRecord::Base
   include Standardizable
 
   ACCREDITATIONS = {
-    "regional" => ["middle", "new england", "north central", "southern", "western"],
-    "national" => ["career schools", "continuing education", "independent colleges", 
+    "REGIONAL" => ["middle", "new england", "north central", "southern", "western"],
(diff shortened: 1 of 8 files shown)


app/models/accreditation.rb
```diff
@@ -36,9 +36,16 @@ class Accreditation < ActiveRecord::Base
   ## Case insensitive inclusion validator
   #############################################################################
   def lowercase_inclusion_validator(attribute, collection, blank_ok = true)
-    return if (var = eval(attribute.to_s)).blank? && blank_ok
+    return if (var = self[attribute]).blank? && blank_ok
 
-    if !collection.include?(var.try(:downcase))
+    # if !collection.include?(var.try(:downcase))
+    #   errors.add(attribute, "#{var} not in [#{collection.join(', ')}]")
+    # end
```
ayaleloehr: Why is this in here but commented out? If it's still needed, can a comment go in about what it does and why it's commented?

mphprogrammer: In case there was an issue following Patricks test - I can remove it, but I normally leave comments like that in as a reference to what I did before.

ayaleloehr:
@mphprogrammer one question back to you. Other than that, LGTM, so you can merge once you address that one comment.

PR #112, opened 2016-06-24
convert \r and \r\n to \n in csvs

mphprogrammer:


mphprogrammer:
I'm converting all the \r and \r\n to \n. Its seems that (my MAC copy at least) Excel converts and XLS to a CSV that uses \r rather than \r\n or \n at the end of every line. This screwed up the ingestion of the CSVs. Now, all terminating \r and \r\n are converted to \n

ayaleloehr:
LGTM

PR #114, opened 2016-06-29
added ? and = to urls

mphprogrammer:


ayaleloehr:
Talked to @mphprogrammer in person re question around why 3 of these have `?` in them and one doesn't.  The `?`'s are for URLs, which the accreditation file doesn't have, and hence doesn't need them. 

LGTM

PR #123, opened 2016-07-07
Accreditation 1

mphprogrammer:
Broke the ugly accreditation sql into more manageable steps. Redid and simplified specs. @ayaleloehr please leave the commented out sql - should something prove wrong - I need to understand my intent. I will delete the comments in the morning should patrick have no issues.


ayaleloehr:
This SQL is both super impressive and terrifying, all at once. 

LGTM.

PR #129, opened 2016-07-08
Changed the way caution flags are updated

mphprogrammer:
@plusjeff @ayaleloehr 
The "Fix" is in. I've changed the way all relevant CSVs update the caution flags. Because some CSVs have all caution flags on a single row while others have a separate row per caution flag, it was necessary to ensure that (1) all rows were included and (2) that the caution flag text doesn't repeat - either for the same CSV or for related CSVs (e.g., Sec 702 and Sec 702 Schools).

There are only 3 changed files, but the changes are large, and the SQL is convoluted (for me), I will be reachable - so let me know if there are any issues!

### DIFF
--- a/app/models/data_csv.rb
+++ b/app/models/data_csv.rb
@@ -420,30 +420,6 @@ class DataCsv < ActiveRecord::Base
     query_str += "AND LOWER(accreditations.csv_accreditation_type) = 'institutional' "
     query_str += "AND LOWER(accreditations.accreditation_status) = 'show cause'; "
 
-    # query_str = 'UPDATE data_csvs SET '
-    # query_str += 'accreditation_type = accreditations.accreditation_type'
-    # query_str += ' FROM accreditations '
-    # query_str += 'WHERE data_csvs.cross = accreditations.cross '
-    # query_str += 'AND accreditations.cross IS NOT NULL '    
-    # query_str += %(AND LOWER(accreditations.periods) LIKE '%current%' )
-    # query_str += %(AND LOWER(accreditations.csv_accreditation_type) = 'institutional' )
-    # query_str += 'AND accreditations.accreditation_type IS NOT NULL '
-    # query_str += 'AND (data_csvs.accreditation_type IS NULL '
-    # query_str += "OR (LOWER(data_csvs.accreditation_type) != 'regional' "
-    # query_str += "AND LOWER(accreditations.accreditation_type) != 'hybrid')); "
-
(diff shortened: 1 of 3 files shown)


ayaleloehr:
TBR= @ayaleloehr. Assigning myself to come back and review. But fix needs to get in and I don't have context to judge these SQL changes this quickly.

app/models/data_csv.rb
```diff
@@ -453,15 +429,37 @@ def self.update_with_accreditation
...
-    query_str += 'UPDATE data_csvs SET '
-    query_str += 'caution_flag_reason = CONCAT(data_csvs.caution_flag_reason,'
-    query_str += "'Accreditation (', accreditations.accreditation_status, '), ')"
-    query_str += ' FROM accreditations '
-    query_str += 'WHERE data_csvs.cross = accreditations.cross '
-    query_str += 'AND accreditations.cross IS NOT NULL '
-    query_str += %(AND LOWER(accreditations.periods) LIKE '%current%' )
-    query_str += 'AND accreditations.accreditation_status IS NOT NULL '
-    query_str += "AND LOWER(accreditations.csv_accreditation_type) = 'institutional'; "
+    query_str += 'UPDATE data_csvs SET caution_flag_reason = '
```
plusjeff: @mphprogrammer Can you leave a comment in a follow-up PR that explains what the SQL here is doing. It's well formatted and clear from a code standpoint, but it's hard for someone following in your footprints to know the [business] purpose of the change and what it's trying to accomplish should someone ever need to maintain it.

mphprogrammer: Sure.

mphprogrammer: If it helps for review, the basic structure (for all changes today) is a bulk update with 2 nested subqueries. The inner-inner subquery is responsible for getting a distinct set of caution flag reasons. The outer-inner subquery is responsible for aggregating all the cautions for a given key together. The bulk query then concat's the new caution flags to the existing caution flag.

Hopefully this solves the problem of having multiple row caution flags in some csvs and single row caution flags in others.

I will add comments to the next PR.

ayaleloehr:
Came back and gave this a look today. The biggest thing to me (besides agreeing with @plusjeff's comments that explaining in words what's happening/why would be super useful) would be also deleting the commented out code. Now that that code has been checked into GitHub, it can be found in the future via the commit history and doesn't need to stay in the file "just in case we need to revert" or anything like that (❤️  version control).

Thanks @mphprogrammer !

PR #130, opened 2016-07-12
Added some comments to the models

mphprogrammer:
Added comments to the models, the Controllers and views are straight forward. There are also comments in the config files and initializers. However, these have been incorporated from the git-go.


app/models/data_csv.rb
```diff
@@ -1,5 +1,10 @@
 require 'csv'
 
+###############################################################################
+## DataCsv
+## Holds the merrged data from CSV-sourced data tables, as well as methods to
```
ayaleloehr: `sed 's/merrged/merged'`

app/models/data_csv.rb
```diff
@@ -299,6 +305,7 @@ def self.initialize_with_weams
 
     names = Weam::USE_COLUMNS.map(&:to_s).join(', ')
 
+    # Include only those schools marked as approved (c.f., Weam model)
```
ayaleloehr: what's c.f.?

app/models/data_csv.rb
```diff
@@ -575,11 +605,12 @@ def self.update_with_scorecard
 
   ###########################################################################
   ## update_with_ipeds_ic
-  ## Updates the DataCsv table with data from the scorecards table.
+  ## Updates the DataCsv table with data from the IC table.
```
ayaleloehr: Is IC a known acronym or defined elsewhere?

app/models/data_csv.rb
```diff
@@ -590,11 +621,12 @@ def self.update_with_ipeds_ic
 
   ###########################################################################
   ## update_with_ipeds_hd
-  ## Updates the DataCsv table with data from the scorecards table.
+  ## Updates the DataCsv table with data from the HD table.
```
ayaleloehr: Same as above about `IC`. Is `HD` defined elsewhere.

app/models/data_csv.rb
```diff
@@ -804,11 +837,17 @@ def self.update_with_settlement
...
   def self.update_with_hcm
+    # Sets caution flags if the corresponing approved school (by OPE6 id) 
+    # has an entry in the Hcm table.
     query_str = 'UPDATE data_csvs SET '
     query_str += 'caution_flag = TRUE'
     query_str += ' FROM hcms '
     query_str += 'WHERE data_csvs.ope6 = hcms.ope6; '
 
+    # Sets the caution flag reason for each school, by OPE6. The inner select
+    # encloses unique reasons within a lable while the outer select aggregates
```
ayaleloehr: lable -> label.

ayaleloehr:
@mphprogrammer only a few small nits, LGTM otherwise. Feel free to merge once you address those small issues. Nice work here, the more we can get out of your head about what you did (and why), the better it will be for this project's long term success, so thanks for taking the time to write it all out!

app/models/data_csv.rb
mphprogrammer: Its the latin abbreviation for "conferre" meaning refer elsewhere (in this case, the Weam model) for more information.

ayaleloehr:
LGTM.

Issue #135, opened 2016-08-26
GIDS needs updates to deploy to relative URL path


patrickvinograd:
To comply with govcloud standardization, we will serve GIDS from a path under dev.vets.gov. Right now the application is kind of baked to only serve out of the root path. Following changes are sufficient to get this to work:
1. Set `config.assets.prefix = ENV['RAILS_RELATIVE_URL_ROOT']` in production.rb. 
2. Add a wrapper around all the routes in routes.rb:

```
Rails.application.routes.draw.do
  scope ENV['RAILS_RELATIVE_URL_ROOT'] do
    <all existing routes>
    ...
  end
end
```
1. Not specifically related to relative paths but it would be nice to throw a reasonable puma.rb into config/ so we don't have to insert one at devops time:

```
#!/usr/bin/env puma

environment 'production'

preload_app!

tag 'gi-bill-data-service'
```

patrickvinograd:
With above, devops deployment can be responsible for setting RAILS_RELATIVE_URL_ROOT env variable.

If there's any concern that this will cause problems with current deployment in east-west, I will create a PR with these changes and we can wait to merge it until right when govcloud is ready to go.

patrickvinograd:
Actually this may not be necessary - apps like VEC do not have their route scoped nor assets prefixed. Thinking this may be solveable on the revproxy with correct X-Forwarded-\* headers.

patrickvinograd:
Above changes are not necessary, but we do need the following in config.ru, which makes GIDS consistent with how VEC and GIBCT deploy:

```
map Rails.application.config.relative_url_root || '/' do
  run Rails.application
end
```

patrickvinograd:
Closed with #136

PR #136, opened 2016-08-29
Add relative URL path mapping for production deployment

patrickvinograd:
For #135. Allows for devops-compliant deployment in govcloud.


ayaleloehr:
It looks like two other places where /gi-bill-comparison-tool is set are:

https://github.com/department-of-veterans-affairs/gi-bill-comparison-tool/blob/master/config/deploy.rb#L41

https://github.com/department-of-veterans-affairs/gi-bill-comparison-tool/blob/master/config/environments/production.rb#L5

Should either of those be changed (maybe removed?) as well?

patrickvinograd:
@ayaleloehr There will be a separate PR for gi-bill-comparison-tool. This one is just for GIDS.

ayaleloehr:
shoot, you are correct. Sorry for greping the wrong repo!  

This matches the config.ru in VEC now. LGTM.

Issue #143, opened 2016-09-13
405 Method Not Allowed


ericks24:
This is my first update to Comparison Tool since we switched Data Service to new URL. I updated the WEAMS file with new CSV file. I clicked "Build New Data.csv" and get 405 Method Not allowed. I attached a screen shot of error message.
![405 method not allowed](https://cloud.githubusercontent.com/assets/15673554/18487247/1af1e024-79ba-11e6-82e9-a8efc485274d.JPG)

patrickvinograd:
I can reproduce this in staging. It is due to moving the application from being deployed on its own under the <host>/ context path to being deployed under <host>/gids. All the other buttons on the dashboard have an href like "/gids/<foo>", but the href for the "build new data.csv" is just "/dashboard". 

My familiarity with rails .erb stuff is pretty minimal so I can puzzle through how to adjust this particular button, but if anybody else knows an easy way go ahead.

patrickvinograd:
This problem is fixed and deployed to production. Please retry at your convenience.

ericks24:
I am now getting a new message when I click "Build New Data.csv".
![something went wrong](https://cloud.githubusercontent.com/assets/15673554/18513504/61c8d7ac-7a53-11e6-9f8a-3149c76cfe26.JPG)

PR #144, opened 2016-09-13
Use helper path instead of hardcoded path for build csv link

patrickvinograd:
Fix for #143. Uses path helper instead of hard-coded path. Reproduced issue and verified with manual fix in staging environment.


ayaleloehr:
LGTM

PR #147, opened 2016-10-07
Scorecard Patch

saintsoup52:
We cannot update the GI Bill CT with Dept of ED Scorecard data till make this update.

The Department of Education on their College Scorecard no longer uses the variable "C200_L4_POOLED_SUPP' for graduation rates for schools that are less than 4 years.

Instead ED now uses a variable called "C150_L4_POOLED_SUPP". 

The big difference is that now 4 year and less than 4 year schools have the same period of time for students to graduate in order to be counted in the the school's graduation rate (150% normal time).

I did my best to find all the instances in the code where C200 was mentioned, but the search function on Git is not the best.


patrickvinograd:
It doesn't seem to me like a global search/replace on the column name is going to do the job - if we deploy this change the data for c150 is going to be empty, and we can't just use the c200 data in place of it (so a column rename is not appropriate). I think to make this seamless we need to:
- Add the empty c150 column to the model/database.
- Update GIDS to understand the new CSV format and populate the new c150 column.
- Push data GIDS -> GIBCT
- Update GIBCT to use the new column for calculations and display.
- (Optionally) drop the obsolete c200 column from the model/database.

If some downtime is tolerable, then this CL might work but there still needs to be a corresponding GIDS change to populate the data into the right column during a maintenance window. 

@saintsoup52 Do you have the complete scorecard csv file with the new c150 column that we can use to test this in our staging environment?

saintsoup52:
Can we set up a time to chat about this by phone?

The C200 column no longer exists and we will be using the C150 column in the exact same way we were using the C200 column. So I don't understand why this wouldn't work, but I am not a coder so... I looking forward to hearing why.

Here is where the file lives:
https://ed-public-download.apps.cloud.gov/downloads/Most-Recent-Cohorts-Scorecard-Elements.csv

Since we last used this file they changed one data point we were using (C200 --> C150) and removed one we were not using (HIGHDEG). Just remember there are two C150's and we will be using them both.

patrickvinograd:
Ok, I was under the mistaken impression that this field made it straight through to the GIBCT institution model; if that were the case it would be more difficult because we'd either need to maintain both fields temporarily, or have some downtime while we could rename the column and inject the new data. Since this field is only on the GIDS side and is used to derive a field in GIBCT that's not changing, we're in much better shape. 

As it is, I think we only need one additional piece beyond what @saintsoup52 has done, which is to add a rails db migration script to update the column name in the postgres database itself. This will get run via rake db:migrate when we redeploy.

patrickvinograd:
Assigning to @ayaleloehr for review. I had to tweak @saintsoup52's commits a bit - in particular, the previous migration script should not have been touched. I verified in local env that I could do a rake db:migrate, restart the app, then upload the new scorecard file. After that I could create a new data.csv and push to GIBCT.

ayaleloehr:
@patrickvinograd we seem to have one version of each input file in /csvs, but this PR adds a new Scorecard CSV without removing the old one. Can we remove the old one to keep it consistent (only having one file of each type)?

patrickvinograd:
@ayaleloehr Done.

ayaleloehr:
LGTM

PR #156, opened 2016-11-04
Add config/puma.rb

### LINKED ISSUE #140: Add a puma.rb file
GIDS should have a puma.rb file under config. It's required for the application to run in govcloud. Right now devops is installing one during our deployment workflow, which works fine, but for consistency with other apps (GIBCT/VEC/et al) it should really come from the application. 

Not a blocker, rather a clean-up/consistency item.

patrickvinograd:
Fixes #140 

Values are aggregated from GIBCT/VEC/devops-GIDS versions, for this app specifics are not too crucial. After this is merged we can remove the injection of puma.rb that happens in the devops role for this application, although having both won't break anything.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,0 +1,30 @@
+#!/usr/bin/env puma
+
+# Set the environment in which the rack's app will run. The value must be a string.
+#
+# The default is "development".
+#
+environment 'production'
+
+# Configure "min" to be the minimum number of threads to use to answer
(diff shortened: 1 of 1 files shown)


ayaleloehr:
LGTM. 

At some point we should look at if having 1-5 threads and 3 workers is actually a good idea or not; I just made that up one day and it seems to have stuck. Could most likely be optimized.

PR #157, opened 2016-11-04
Production release 20161104

jkassemi:


jkassemi:
will open another PR once production brought back into master

PR #159, opened 2016-11-04
Production release 20161104

jkassemi:


jkassemi:
had to update master with production changes. I believe the above commit references are due to a prior squash merge into production, but changed files accurately reflect the differences between current prod and master.

PR #164, opened 2016-11-29
Changed ope justification from left to right

mphprogrammer:
Despite a few lines that had trailing blanks (that were deleted), the only real change was in the pad method of 
```
app/models/concerns/standardizable.rb:128
```


ayaleloehr:
LGTM.

PR #170, opened 2016-12-09
Update scorecard_test.csv

knkski:
Found this branch hanging around. @ayaleloehr, @saintsoup52, does this still need merging?


ayaleloehr:
I don't know what this is. going to close, but if @saintsoup52 has some reason it needs to get in or knows more about it, please comment here and we can reconsider.

saintsoup52:
So this pull request was because ED changed one of the headers in their college scorecard. We fixed this in the data service but we did not fix it in the example.

The change is hard to see because the line is so long but it is there:
C200_L4_POOLED_SUPP --> C150_L4_POOLED_SUPP

Specifically this changes means we (ED/VA) are now displaying the graduation rate for 2 year schools based on 150% of normal time (3 years) not 200% of normal time (4 years).

ayaleloehr:
Cleaning out my inbox and tagging @ben-USDS-damman and @mphprogrammer that this was probably supposed to have changed but I don't think did. will potentially want to account for in the work you're doing now.

mphprogrammer:
The file in question is no longer being used, but I will have to make sure it isn't repeated in the new test files. I will have to account for this -> This column is used by the DS to compute grad rates, but not by GIBCT

PR #175, opened 2017-01-05
Initial Commit - In Progress

mphprogrammer:
Added the Loadable and Exportable modules, responsible for loading all CSVs, and the export module responsible for exporting data to CSVs. Added the Weam model as proof of concept for loading and exporting. The Version model is also provided to give some idea of how the data will be stored. Versions of the data used by the GIBCT can be manipulated by adding new records to the versions table. The most recent (as determined by `created_at`) version record where `production: true` is the production version, while the most recent version where `production: false` is the preview version. In this way, we can switch version simply be adding records, and can determine which version was in force on any given date/time.


app/models/version.rb
```diff
@@ -0,0 +1,60 @@
...
+    query = 'SELECT * FROM versions '\
+            'INNER JOIN '\
+              '(SELECT max(created_at) AS max_created_at FROM versions WHERE production=? AND created_at<?) as mv '\
+              'ON created_at=mv.max_created_at '\
+            'WHERE production=?;'
+
+    Version.find_by_sql([query, production, time, production]).first
+  end
+
+  def self.find_max_version(production = true)
```
patrickvinograd: I think this would be better called `find_most_recent_version` or something since it's not finding the max numerical version, it's finding the one with the latest created_at timestamp.

app/models/version.rb
```diff
@@ -0,0 +1,60 @@
...
+    query = 'SELECT * FROM versions '\
```
patrickvinograd: I believe this could be done without SQL and without subquery via the Rails query API. Something like: `Version.production.order(created_at).first`

app/models/version.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def self.preview_version_by_time(time)
+    find_version_by_time(time, false)
+  end
+
+  def self.find_version_by_time(time, production = true)
+    time = Time.zone.parse(time).to_datetime if time.is_a?(String)
+
+    query = 'SELECT * FROM versions '\
```
patrickvinograd: This could also be done using Rails query API, a little more awkwardly than the case below though.

app/models/user.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 class User < ActiveRecord::Base
   has_many :versions, inverse_of: :user
```
patrickvinograd: I'm confused by the model relationship between User and Version - why does a version belong to an individual user? Is it just to track who happened to create that version, or is there some limited ownership/visibility of specific versions?

config/csv_file_defaults.yml
```diff
@@ -1,5 +1,6 @@
 
 Weam:
-  skip_lines_before_header: 0
-  skip_lines_after_header: 1
-  delimiter: ','
+  skipped_first_lines: 0
+  col_sep: ','
+  row_sep: "\r\n"
```
patrickvinograd: Is there any way to make cvs upload work irrespective of line termination? Or are there CSVs that have `\n` within a line or something?

app/models/version.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def self.production_version_by_time(time)
+    find_version_by_time(time)
+  end
+
+  def self.preview_version_by_time(time)
+    find_version_by_time(time, false)
+  end
+
+  def self.find_version_by_time(time, production = true)
```
patrickvinograd: Pass in the scopes you created above instead of a boolean.

app/models/version.rb
mphprogrammer: Point taken, in the next PR today I'll rename it.

config/csv_file_defaults.yml
mphprogrammer: Unfortunately, no. Windows uses a different termination than does Unix/OSX. Most cases, the windows termination will be used. However, this is not guaranteed. This option will be presented to the user when uploading. We should make the option understandable for a user though.

app/models/user.rb
mphprogrammer: Every version action is initiated by a User, which then can be tracked. So if some question about who did what arises, we can determine it.

app/models/version.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+class Version < ActiveRecord::Base
+  belongs_to :user, inverse_of: :versions
```
typesend: Related to @patrickvinograd's User model comment...you might consider naming it `:created_by` instead of `:user` to reveal the intent better (here and in the User model.)

db/migrate/20161231003416_create_weams.rb
```diff
@@ -0,0 +1,41 @@
+class CreateWeams < ActiveRecord::Migration
+  def change
+    create_table :weams do |t|
+      t.string :facility_code, null: false
+    	t.string :institution, null: false
+    	t.string :city
+    	t.string :state
+    	t.string :zip
+    	t.string :country
+      t.integer :bah
```
typesend: goofy indentation of lines 5-12

typesend (changes requested):

db/migrate/20161231003416_create_weams.rb
typesend: you've got tabs mixed with your spaces

typesend (approved):

mphprogrammer:
Thanks @patrickvinograd @typesend

PR #176, opened 2017-01-07
Csv Loading and Exporting

mphprogrammer:
* Weam
* Crosswalk
* Accreditation
* EightKey


app/models/eight_key.rb
```diff
@@ -0,0 +1,22 @@
...
+    'institution of higher education' => { institution: InstitutionConverter },
+    'city' => { city: BaseConverter },
+    'state' => { state: StateConverter },
+    'opeid' => { ope: OpeConverter },
+    'ipeds_id' => { cross: CrossConverter },
+    'notes' => { notes: BaseConverter }
+  }.freeze
+
+  def validate_ope_and_cross
+    return if ope.present? || cross.present?
```
typesend: Should this method have a different name? ope_and_cross vs ope || cross logic. Perhaps `validate_either_ope_or_cross` or similar would be more descriptive of what's actually happening.

app/models/weam.rb
```diff
@@ -42,18 +42,16 @@ class Weam < ActiveRecord::Base
   validates :institution, presence: true
   validates :bah, numericality: true, allow_blank: true
 
-  validate :validate_derived_fields
+  validate :derive_fields
 
   # Computes all fields that are dependent on other fields. Called in validation because
   # activerecord-import does not engage callbacks when saving
-  def validate_derived_fields
+  def derive_fields
```
typesend: Could be mistaken, but it looks like you can enable callback calling. https://github.com/zdennis/activerecord-import/wiki/Callbacks

Either way, I do think it would be best to have a separate method for mutating those dependent attributes.

Outside the scope of this PR, but I do wonder if we could eliminate some of these dependent attributes–at least the ones that are equivalent to one other attribute. Similar things are happening in the frontend that are probably best simplified in the backend first.

app/models/converters/integer_converter.rb
```diff
@@ -4,6 +4,8 @@
 class IntegerConverter < BaseConverter
   def self.convert(value)
     value = super(value)
+
+    return value unless value.respond_to? :to_i
     value.blank? ? nil : value.to_i
```
typesend: Integers do respond to `:to_i` so this line is unnecessary / buggy.

typesend (dismissed):

app/models/converters/integer_converter.rb
typesend: Curious: what was the Class of a value that we want to return that doesn't have a `#to_i`?

mphprogrammer:
@typesend I made a whole bunch of changes, could you review?

typesend (approved):
LGTM and you've made a number of improvements I like, esp the use of variables instead of constants for dynamic class names.

PR #179, opened 2017-01-17
Rewrite 1 csvs

mphprogrammer:
* Scorecard
* ArfGiBill
* P911Tf
* P911Yr
* Vsoc
* Sva
* Sec702
* Sec702School
* Mou


mphprogrammer:
@typesend This is not a large a PR as it seems. Most of the changes are cookie cutter, with a lot of CSVs added in to the MX.

typesend (approved):
Looks like more of the same and LGTM

PR #180, opened 2017-01-17
Adding more csvs

mphprogrammer:
* Hcm
* Settlement
* IpedsIC
* IpedsHd
* IpedsIcAy
* IpedsIcPy
* Complaint
* Outcome


mphprogrammer:
@typesend This should have an in-sync schema, can't imagine how it happened...

mphprogrammer:
@typesend @saneshark - I know this seems like a HUGE PR, it really isn't. A lot of the files are migrations/Csvs/Fixtures. Specs too. The models however are cookie cutter. Its mostly a mapping from the csv file to the activemodel (MAP) along with a before_validation_callback. Each model is pretty much the same as the other...

app/controllers/complaints_controller.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+class ComplaintsController < ApplicationController
+end
```
saneshark: @mphprogrammer why are these stubbed controllers necessary for this PR? Can they wait until these are actually implemented in a separate PR?

app/models/complaint.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+# Complaint data appears in its CSV as list of individual student veteran complaints keyed by an optional facility code.
```
saneshark: Can we put this comment in a Markdown within docs/ folder and then just link to it here? I think someone reading the code could figure this out without the comment actually, but if anything having this really long comment before getting to the code is not desirable.

app/models/complaint.rb
```diff
@@ -0,0 +1,119 @@
...
+  validates :closed_reason, inclusion: { in: CLOSED_REASONS }, allow_blank: true
+
+  before_validation :derive_dependent_columns
+
+  def self.update_ope_from_crosswalk
+    find_each do |complaint|
+      crosswalk = Crosswalk.find_by(facility_code: complaint.facility_code)
+      complaint.update(ope: crosswalk.ope, ope6: crosswalk.ope6) if crosswalk
+    end
+  end
```
saneshark: This is going to do a database lookup for each item, and then do an update for each one? This is not good. It results in N+1, we should investigate optimizing this.

app/models/converters/boolean_converter.rb
```diff
@@ -5,6 +5,6 @@ class BooleanConverter < BaseConverter
 
   def self.convert(value)
     value = super(value)
-    value.blank? ? nil : TRUTHS.include?(value.downcase)
+    value.blank? ? nil : TRUTHS.include?(value.to_s)
```
saneshark: So I believe @typesend introduces virtus into this project with his PR, would love to maybe see this type of thing replaced with a custom coercion class. We can revisit this in another PR.

app/models/ipeds_ic_py.rb
```diff
@@ -0,0 +1,142 @@
...
+
+  validates :cross, presence: true
+  validates :chg1py3, numericality: true, allow_blank: true
+  validates :books, numericality: true, allow_blank: true
+
+  before_validation :derive_dependent_columns
+
+  def derive_dependent_columns
+    self.tuition_in_state = chg1py3
+    self.tuition_out_of_state = chg1py3
```
saneshark: Are we setting a class variable based on an instance variable? Why do we need to do "self." here, isn't self the instance? I think you can just do "tuition_in_state = " without the self

app/models/ipeds_ic.rb
```diff
@@ -0,0 +1,149 @@
...
+  validates :calsys, inclusion: { in: [-2, 1, 2, 3, 4, 5, 6, 7] }
+  before_validation :derive_dependent_columns
+
+  def derive_dependent_columns
+    self.credit_for_mil_training = IpedsIc.coded_to_boolean(vet2)
+    self.vet_poc = IpedsIc.coded_to_boolean(vet3)
+    self.student_vet_grp_ipeds = IpedsIc.coded_to_boolean(vet4)
+    self.soc_member = IpedsIc.coded_to_boolean(vet5)
+    self.online_all = IpedsIc.to_online_all(distnced)
+    self.calendar = IpedsIc.to_calendar(calsys)
```
saneshark: Can we remove the "self." ?

config/routes.rb
```diff
@@ -1,4 +1,12 @@
 Rails.application.routes.draw do
+  resources :outcomes
+  resources :complaints
+  resources :ipeds_ic_pies
+  resources :ipeds_ic_ays
+  resources :ipeds_hds
+  resources :ipeds_ics
+  resources :settlements
+  resources :hcms
```
saneshark: Are these routes actually exposed? Looking at the corresponding controller stubs above, lets not add / expose these routes until the controller logic is actually implemented.

db/migrate/20170118010335_create_ipeds_ics.rb
```diff
@@ -0,0 +1,125 @@
...
+      t.boolean :online_all
+      t.integer :calsys, null: false
+      t.integer :distnced, null: false
+      t.integer :vet2, null: false
+      t.integer :vet3, null: false
+      t.integer :vet4, null: false
+      t.integer :vet5, null: false
+
+      # Not used in building DataCsv, but used in exporting source csv
+      t.integer :peo1istr
```
saneshark: all of these t.integers, is it OK if they end up being null? or should you have null: false on these as well?

db/migrate/20170122220228_create_ipeds_ic_ays.rb
```diff
@@ -0,0 +1,245 @@
...
+    create_table :ipeds_ic_ays do |t|
+      # Used in the building of DataCsv
+      t.string :cross, null: false
+      t.integer :tuition_in_state
+      t.integer :tuition_out_of_state
+      t.integer :books
+
+      # Not used in building DataCsv, but used in exporting source csv
+      t.string :xtuit1
+      t.integer :tuition1
```
saneshark: Should some of these also have null: false or default 0 ?

db/migrate/20170123202716_create_ipeds_ic_pies.rb
```diff
@@ -0,0 +1,136 @@
...
+      t.integer :tuition_out_of_state
+      t.integer :books
+
+      # Not used in building DataCsv, but used in exporting source csv
+      t.integer :prgmofr
+      t.float :cipcode1
+      t.string :xciptui1
+      t.integer :ciptuit1
+      t.string :xcipsup1
+      t.integer :cipsupp1
```
saneshark: Should some of these also have `null: false` or `default: 0`?

spec/controllers/complaints_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe ComplaintsController, type: :controller do
+end
```
saneshark: Let's not add any of these controller spec stubs until you've written up the controllers. No point merging these changes in as they provide no new functionality.

spec/controllers/hcms_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe HcmsController, type: :controller do
+end
```
saneshark: Let's not add any of these controller spec stubs until you've written up the controllers. No point merging these changes in as they provide no new functionality.

spec/controllers/ipeds_hds_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe IpedsHdsController, type: :controller do
+end
```
saneshark: Let's not add any of these controller spec stubs until you've written up the controllers. No point merging these changes in as they provide no new functionality.

spec/controllers/ipeds_ic_ays_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe IpedsIcAysController, type: :controller do
+end
```
saneshark: Let's not add any of these controller spec stubs until you've written up the controllers. No point merging these changes in as they provide no new functionality.

spec/controllers/ipeds_ic_pies_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe IpedsIcPiesController, type: :controller do
+end
```
saneshark: Let's not add any of these controller spec stubs until you've written up the controllers. No point merging these changes in as they provide no new functionality.

spec/controllers/settlements_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe SettlementsController, type: :controller do
+end
```
saneshark: Let's not add any of these controller spec stubs until you've written up the controllers. No point merging these changes in as they provide no new functionality.

saneshark (changes requested):
The only reason I'm requesting changes is because that N+1 query is not good. All the other changes recommended are minor.

mphprogrammer:
@saneshark Thanks for the speedy review! Regarding the N+1 query issue. GIDS has no relations between tables partly because we do not know in what order CSVs (tables) are loaded (populated), and because there are many schools in different CSVs (foreign_key) that are not in Weams or Crosswalks (primary_key), in some cases the foreign_key (ope in the comment in this case) in the CSV is just plain wrong. Anyway, when the institution table is being built, one of the things that needs to be done is to find the correct ope based on facility code, which is then used to "roll-up" complaint counts to each campus (using facility code) and each school (using ope6).

mphprogrammer:
@saneshark Regarding the defaults - I've only set defaults on those fields used in further calculations. However, there is a need (perhaps desire would be a better term) to differentiate between null (meaning no value supplied - in some cases) and 0 or false.

mphprogrammer:
@saneshark I will get rid of the not-as-of-yet used controllers and routes

app/models/complaint.rb
mphprogrammer: Sure - in this case, we need to replace the existing (incorrect) ope id in every complaint record with the correct one from the Crosswalk table. This is done by taking the presumably correct facility_code in per complaint record and referencing the Crosswalk table for the associated ope id. There is no guarantee that there will be a Crosswalk record for that facility code. We need the ope id to derive the ope6 id which identifies all campuses for a given institution. We can then sum all the complaints in each category (finances, admissions, etc) on a campus by campus basis, as well as on an institution basis.

mphprogrammer: All that begin said, I think can do this as an update/join.

mphprogrammer: @saneshark I think this is what I am looking for
```
UPDATE complaints
  SET ope = crosswalks.ope, ope6 = crosswalks.ope6
  FROM crosswalks
  WHERE complaints.facility_code = crosswalks.facility_code AND crosswalks.ope IS NOT NULL;
```

saneshark: @mphprogrammer I don't know if you can do this. Shouldn't there be a relationship between complaints and crosswalks? Maybe not, I'm not sure. If there is a relationship I would think you need to do a join statement no? Maybe I'm confused on the intended purpose of this still.

mphprogrammer: @saneshark I've tried this and it seems to work. I really liked the idea of a single update statement, and it seems to be very fast. In this usage the from acts as the join. I've tried it on my SQL client, but of course it'll have specs. Not sure given the update why AREL would be needed though, but I have used it before.

saneshark: AREL allows you to write queries in a way that is similar to raw SQL. It is bad practice to write something like Database.connection.execute(raw_sql)

With AREL you are still writing ruby vs sending a raw SQL statement. 

It's bad practice for a number of reasons, but most notably, there isn't a mechanism to prevent against things like SQL injection automatically. Raw sql is dangerous. But equally as important, the errors rendered with raw sql when a query fails are not as good as the ones when AREL query fails.

saneshark (approved):

saneshark:
LGTM @mphprogrammer

PR #182, opened 2017-01-24
Addition of v0 JSON API

typesend:
- JSON:API compliant endpoints for:
  - autocomplete
  - calculator constants (w/out versioning)
  - institution search results (w/out versioning, excludes unneeded attributes)
  - institution details (w/out versioning, selectively excludes many attributes that are often empty)
- Common Exceptions
- Rudimentary specs
- Includes gids:load_csv rake task to bootstrap your `institutions` table for now (until @mphprogrammer implements it for realz)
- Includes migration for `data_csvs` table for comparison to `institutions`. I picture one of these tables going away as our work converges.
- Removed jbuilder gem, added virtus and serializers

There *might* be a couple more attributes we need to add to a serializer(s) that are needed by the frontend.

Also note that `HOSTNAME` and `SECRET_KEY_BASE` env variables are now required by Figaro.

Last but not least, we should begin discussing how the `calculator_constants` table gets updated. Might be a new UI requirement for GIDS? Historically these constants were updated by modifying frontend source code.

Example:
```js
///////////////////////////////////////////////////////////////////////////////
// Constants
///////////////////////////////////////////////////////////////////////////////

Calculator.prototype.TFCAP = 21970.46;
Calculator.prototype.AVGBAH = 1611;
Calculator.prototype.BSCAP = 1000;
Calculator.prototype.BSOJTMONTH = 83;
Calculator.prototype.FLTTFCAP = 12554.54;
Calculator.prototype.CORRESPONDTFCAP = 10671.35;

Calculator.prototype.MGIB3YRRATE = 1789;
Calculator.prototype.MGIB2YRRATE = 1454;
Calculator.prototype.SRRATE = 368;

Calculator.prototype.DEARATE = 1021;
Calculator.prototype.DEARATEOJT = 745;

Calculator.prototype.VRE0DEPRATE = 605.44;
Calculator.prototype.VRE1DEPRATE = 751.00;
Calculator.prototype.VRE2DEPRATE = 885.00;
Calculator.prototype.VREINCRATE = 64.50;
Calculator.prototype.VRE0DEPRATEOJT = 529.36;
Calculator.prototype.VRE1DEPRATEOJT = 640.15;
Calculator.prototype.VRE2DEPRATEOJT = 737.77;
Calculator.prototype.VREINCRATEOJT = 47.99;
```

department-of-veterans-affairs/vets-website#3960


saneshark:
@typesend I noticed that a lot of the code, particularly with ActiveModelSerializers (including initializer monkey patch), Common components, etc are basically a duplicate of what's in vets-api.

I know previously we discussed that vets-api would just become a pass through, but I didn't anticipate this much code overlap. I'm going to investigate moving some of that stuff into a gem so that we don't have this code duplication. That being said, I don't think we need it to be a complete pass through like this. In other words, if the API has its own format, the client in vets-api can deserialize objects / classes just as well I think. 

Let me ponder this a bit more and perhaps we can discuss on a phone call to arrive at better solution than having this code duplication.

saneshark (commented):
Overall this looks great. I have to think through the code duplication stuff and determine if the error handling logic is even necessary for this or not.

I imagine anyone consuming this api would do so through vets-api and not with this service directly, as such some of the logic in api controller might also be a bit heavy handed and we can get by with a more simple approach.

app/serializers/institution_profile_serializer.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+class InstitutionProfileSerializer < ActiveModel::Serializer
+  include OptionalSerializedAttributes
+
+  attribute :institution
+  attribute :facility_code
+  attribute :institution_type_name, key: :type
```
typesend: Specifies a key in the JSON response that differs from the db field name.

app/serializers/institution_profile_serializer.rb
```diff
@@ -0,0 +1,105 @@
...
+  attribute :city
+  attribute :state
+  attribute :zip
+  attribute :country
+  attribute :bah
+  attribute :cross
+  attribute :ope
+  attribute :pred_degree_awarded
+  attribute :locale
+  attribute :gibill, key: :student_count
```
typesend: Specifies a key in the JSON response that differs from the db field name.

app/serializers/institution_profile_serializer.rb
```diff
@@ -0,0 +1,105 @@
...
+  attribute :online_all
+  attribute :p911_tuition_fees
+  attribute :p911_recipients
+  attribute :p911_yellow_ribbon
+  attribute :p911_yr_recipients
+  attribute :accredited
+  attribute :accreditation_type
+  attribute :accreditation_status
+  attribute :caution_flag, if: -> { object.caution_flag.present? }
+  attribute :caution_flag_reason, if: -> { object.caution_flag.present? }
```
typesend: Why would it be of any value for our responses to contain mostly null values?

mphprogrammer:
A heads up: Since we've removed institution type from the DB, the data_csv table and the institutions table is really one and the same. On the GIBCT its the institutions table, on the GIDS its data_csv table. We should standardize around one or the other...

Also, to which ever table name we choose, I will be adding a version_number column so that multiple versions may coexist.

db/migrate/20160224011900_create_data_csvs.rb
```diff
@@ -0,0 +1,226 @@
...
+      t.string :facility_code, null: false
+      t.string :institution, null: false
+      t.string :city
+      t.string :state
+      t.string :zip
+      t.string :country
+      t.string :va_highest_degree_offered
+      t.string :type # BEWARE!!!!
+      t.integer :bah
+      t.boolean :poe
```
mphprogrammer: There is a data dictionary that comes with the Data Service that defines each field and sources the information.

app/serializers/concerns/optional_serialized_attributes.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+module OptionalSerializedAttributes
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def attribute_if_positive(optional_attribute)
+      attribute optional_attribute, if: -> { object.send(optional_attribute).positive? }
+    end
```
typesend: I disagree. This was done advisedly. I fail to see how sending bloated responses to clients would help anyone.

app/serializers/institution_profile_serializer.rb
saneshark: ahh, ok, I was hoping that's what it does. That makes sense.

app/serializers/institution_profile_serializer.rb
mphprogrammer: For consistency, I have to keep field names the same as the existing Data.csv as used by the VA EDU folks.

app/serializers/institution_profile_serializer.rb
saneshark: Because JSON conforms to an object. And object has attributes that are either present or null, but the key should always be present or else its not the same object every-time.

app/models/institution.rb
```diff
@@ -0,0 +1,130 @@
...
+  def self.autocomplete(search_term, limit = 6)
+    search_term = search_term.to_s.strip.downcase
+    Institution.select('id, facility_code as value, institution as label')
+               .where('lower(institution) LIKE (?)', "#{search_term}%")
+               .limit(limit)
+  end
+
+  # Finds exact-matching facility_code or partial-matching school and city names
+  #
+  scope :search, lambda { |search_term|
```
mphprogrammer: If someone autocompletes on "UNI" there will be thousands.

app/models/concerns/tristateable.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# Reads a column, bypassing activerecord field accessors, so that we can
+# differentiate between nil values and boolean false (by default nils in
+# activerecord are converted to false). Naturally, booleans can only hold
+# true/false, so these fields should be created as strings.
+#
+module Tristateable
```
mphprogrammer: Not sure we need the Tristateable stuff, the DS reads in strings as booleans now and will produce T/F/nil for any column used in the institutions table (data_csv).

app/models/institution.rb
```diff
@@ -0,0 +1,130 @@
...
+  validates :institution_type_name, presence: true
+  validates :institution, presence: true
+  validates :country, presence: true
+
+  self.per_page = 10
+
+  def credit_for_mil_training
+    tristate_boolean(:credit_for_mil_training)
+  end
+
```
mphprogrammer: I don't think that the tristate_boolean stuff is required anymore...

app/models/concerns/tristateable.rb
mphprogrammer: The tristate stuff arose because the data was originally sourced in a rake'd csv that had 10 ways to express T/F in the same spreadsheet.

app/models/institution.rb
patrickvinograd: We did have issues with the original GIBCT design for both autocomplete and search operations when they returned a lot of results. We fixed the autocomplete by limiting number of results returned, as Ben has done above. 
For search, originally the problem was with how it was querying to generate the filter UI counts - i.e. being able to refine search by state, etc. The original implementation was fetching all matching records so that it could generate the filter counts, which for a search term like "a" would fetch thousands of institutions and cause memory pressure on the application. I updated that to do "select count" queries which eliminated the memory pressure, but in some sense probably pushed the issue back on to Postgres. 
I'm still groking this implementation to see how it compares but I agree all these `LIKE` queries are something to keep an eye on.

app/serializers/concerns/optional_serialized_attributes.rb
saneshark: It's not just a matter of bloated responses. 

1. You have a schema that acts as a form of documentation for what the object describes. This schema should be consistent so that data consumer knows how to build an object and what attribute will be present. Having attributes that are sometimes present, sometimes not complicates this process.

2. There's a debugging aspect too, which is you don't know if the property is ommitted due to a bug or processing error vs intentionally being left off.

3. Almost all Public Facing APIs will provide null value keys. There are many reasons why this is useful to a consumer of a public facing API that goes beyond the benefits to be gained by marginally less bloated responses.

If anything, JSONAPI allows for a consumer of the data to specify what attributes they want returned. If we want to mitigate bloated responses we should follow that approach.

saneshark: Not including an attribute implies this item doesn't have this, but it could additionally imply that all items in the set of items of this type do not have this attribute.

Whereas a null value implies this value for this item is null, all other items in the set of items having this type could or could not have a value for the attribute.

We want to convey the latter, not the former.

typesend:
@mphprogrammer I don't believe data_csvs and institutions are strictly identical. You can see other differences in this PR.

app/models/institution.rb
patrickvinograd: I wonder if something like pg_search, which lets us use Postgres's native fulltext search/indexing, would be worthwhile. Seems like not quite the intended use case (seems more geared towards documents), but as-is we're doing full table scans on every search due to the leading wildcard in the LIKE query.

saneshark: Yea, this is why i'm leaning towards sphinx or elastic search actually, the strings in question would be properly indexed allowing for very fast lookup. That being said, pg_search might do something very similar to this. I don't have as much experience with that as I do with sphinx and es

saneshark: @aub tagging you here as you should be aware of what we're discussing as well. I am thinking for now we can probably stick to what we have here, doing our best to enhance and optimize the query using native postgres and perhaps revisit Sphinx and / or ES. But wanted you to add any thoughts as well.

aub: I haven't used pg_search either. For the filter counts thing, ES would definitely help with its faceted searching, but my inclination here would be to keep it as simple as possible until we identify a problem. I'll take a look at pg_search now... am curious what it's about.

patrickvinograd: pg_search is used in VEC FWIW. I don't see any fulltext indexes created but it is used for queries in the veterans model.

typesend:
I've made all the optional keys *non-optional* and added json schema specs. Was able to simplify and normalize some rough json edges in the process. I've started working with pg_search but that will be a separate PR.

app/models/institution.rb
saneshark: Opening a separarte issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/950 To track this. @aub if it is OK with you, we will address this in a separate PR.

saneshark (commented):
@typesend this looks good to me, but I'm going to tag @mphprogrammer in for a quick review as well. I opened an issue for tracking search / autocomplete you can reference in your PR when it is ready.

mphprogrammer:
I am wondering if 

* we should fold in the error infrastructure used by the API to the VA EDU front end as well.
* we should consolidate some constants (e.g., LOCALE)

app/models/institution.rb
```diff
@@ -0,0 +1,125 @@
...
+    0 => nil,
+    'ncd' => 'Certificate',
+    1 => 'Certificate',
+    '2-year' => 2,
+    2 => 2,
+    3 => 4,
+    4 => 4,
+    '4-year' => 4
+  }.freeze
+  TYPES = [
```
mphprogrammer: Since school types are populated by the DS, and if we remove the validations (which may not be necessary - see below) then TYPES should not be required.

app/models/concerns/tristateable.rb
mphprogrammer: Are we sure that we still need the Tristateable? The DS will populate all booleans with True/False/Nil - so you only need to differentiate between nil and a boolean T/F.

typesend:
Regarding constants I think only constants the frontend needs belong in the calculator_constants table.

PR #183, opened 2017-02-01
Adding sample constants to seeds.rb

ben-USDS-damman:
These are the current values sourced from production.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -4,4 +4,32 @@ User.destroy_all
 puts 'Add new users ... '
 User.create(email: ENV['ADMIN_EMAIL'], password: ENV['ADMIN_PW'])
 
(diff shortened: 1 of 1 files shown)


saneshark:
@ben-USDS-damman since this is just an update to the constants, I'm going to give it the LGTM, and we can try to address the code coverage issues separately in a different PR.

saneshark (approved):
approving. OK to merge despite failing travis (it's failing for code coverage only).

saintsoup52:
Sorry for the dumb question... does this mean we will be able to update the constants on the GI Bill CT the same way we update the data on the tool?

Because that would be awesome sauce!

ben-USDS-damman:
@saintsoup52 Yes, that's the general idea. 💯

PR #185, opened 2017-02-01
Fixing linting violations

knkski:


ben-USDS-damman:
Compared this to our `rewrite-1` branch where there are no linting violations.

The `TargetRubyVersion: 2.3` line was missing from `rewrite-1` but all other settings were already identical. Still no violations in `rewrite-1`.

image

ben-USDS-damman:
I think this is no longer relevant

Issue #139, opened 2016-08-31
Devise sessions table looks like it grows without bound


patrickvinograd:
While doing db migration from e/w to govcloud, was debugging a db restore and observed the following:

```
select count(*) from sessions;
 count
--------
 411715
```

It seems like devise never culls sessions from that table. Not sure if this is a matter of devise configuration or a bug.

We can clear that table with a rake:db command but we should also ensure this doesn't continue to happen.

ben-USDS-damman:
@patrickvinograd I would imagine our half a dozen users rarely log out. Can we easily configure a `rake db:sessions:clear RAILS_ENV=production` cron job to run every weekend?

patrickvinograd:
We are now up to:
```
select count(*) from sessions;
  count
---------
 1905709
(1 row)
```
so I kind of suspect that every get of a GIDS url results in a new entry in the sessions table?

patrickvinograd:
E.g. the automated monitoring.

ben-USDS-damman:
Perhaps pointing it to an API endpoint (no auth) would help in the future?

patrickvinograd:
Yeah, right now all GIDS controllers are authenticated but once the API endpoints are available I'll point at that. Would like to keep this open to track verifying that we aren't still generating infinite session rows and also to track clearing the existing sessions.

patrickvinograd:
If for any reason we decide we still need a cron job (I think we can live without it given we have less than 10 users) I'll set up `whenever` with a rake task. I know we use sidekiq in vets-api but that seems like overkill for this.

ben-USDS-damman:
Closing this issue because it's tracked elsewhere

PR #187, opened 2017-02-02
Session cleaning proposal

ben-USDS-damman:
As a simple solution to our cancerous sessions problem (#139) I propose we call this method to clear inactive sessions every time new data is published.

This avoids adding more gems, cron jobs, or other overhead to our application while still ensuring that old sessions get swept every couple of weeks.

```
class DashboardsController
  include SessionCleaning
  ...

  def publish
    delete_inactive_sessions!
    ...
  end
end
```

### DIFF
--- a/app/controllers/concerns/session_cleaning.rb
+++ b/app/controllers/concerns/session_cleaning.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+module SessionCleaning
+  extend ActiveSupport::Concern
+
+  def delete_inactive_sessions!
+    ActiveRecord::SessionStore::Session.delete_all ["created_at < ?", 1.week.ago]
+  end
+end
--- a/app/controllers/concerns/session_cleaning.rb
+++ b/app/controllers/concerns/session_cleaning.rb
@@ -3,6 +3,6 @@ module SessionCleaning
   extend ActiveSupport::Concern
 
   def delete_inactive_sessions!
-    ActiveRecord::SessionStore::Session.delete_all ["created_at < ?", 1.week.ago]
+    ActiveRecord::SessionStore::Session.delete_all ['created_at < ?', 1.week.ago]
   end
 end


patrickvinograd:
In principle if the user doing the publish operation logged in more than a week ago, this immediately kills their session, which is probably a scary and not good user interaction. 

The sessions table has an updated_at field, so we could instead delete sessions that haven't been updated in some amount of time (I'm assuming the updated_at field gets refreshed as user interacts with the tool).

patrickvinograd:
Other than that though, I like this idea. Low-overhead. I think we will avoid most of this problem by changing our health-check endpoint to something that doesn't spawn a session, but this is a nice backstop.

saneshark:
Also Can we do this maybe using sidekiq worker? I can't help to think that this should happen asynchronously. That would maybe require adding Redis though. What about delayed_job? Should work with postgres. (NOTE: I am not a big fan of delayed job).

Additional alternative to sidekiq and delayed_job: https://github.com/brandonhilkert/sucker_punch

patrickvinograd:
I want to keep it as operationally simple as possible - no additional processes to launch, etc. If Sucker Punch fits the bill, that's fine. 

But, we are only spewing sessions because our monitoring health check is going to `/users/log_in`. Once we have an unauthenticated API endpoint that monitoring can visit instead, we will have on the order of ~10 sessions to delete at any given time and I think clearing them synchronously would be workable. (And we can manually delete the current crop of 1M+ sessions)

saneshark:
@ben-USDS-damman @patrickvinograd 

lets put together:

1) A service class (app/services/session_cleaner.rb) that invokes removal of N number of sessions that are older X number of hours. ie. WHERE (updated_at < X.hours.ago.utc) LIMIT (N)

2) create a class method for the session cleaner that invokes it for N = 10, and X = 336; This will be used for what you are calling delete_inactive_sessions!

3) create a class method for the session cleaner that invokes it for N = Session.size, and X = 336; This will be invoked using a rake task

4) create a rake task that calls the method created in 3)

Note: 336 hours, is 2 weeks. I think this is reasonable amount of time to expire a session and require login again.

Thoughts?

The benefit of having this as a service object is that later we can very easily wrap it up in an async call, and of course it helps facilitate having a rake task all in one DRY place.

Also, X should ideally follow @patrickvinograd 's suggestion of being based on updated_at, assuming updated_at is in fact touched with subsequent user interactions.

ben-USDS-damman:
5. Verify / implement sessions update such that updated_at is actually a source of inactivity truth

saneshark:
Closing this due to inactivity.

PR #188, opened 2017-02-03
Rewrite 1 institution

mphprogrammer:
* Implements versioning in institutions table
* Converted string to Boolean in institutions table where appropriate
* Added flight, correspondence, and ope6 to institutions
* Removes DataCsv which is now redundant
* Added InstitutionBuilder to load data from CSV tables to Institution.
* Add Weam, Crosswalk, Sva, Vsoc to InstitutionBuilder
* Added Accreditation to InstitutionBuilder
* Adjust api controller specs to use version
* Added ArfGiBill to InstitutionBuilder
* Added P911Tf to InstitutionBuilder
* Added P911Yr to InstitutionBuilder
* Added Mou to InstitutionBuilder
* Added Scorecard to InstitutionBuilder
* Added IpedsIc to InstitutionBuilder
* Added IpedsHd to InstitutionBuilder
* Added IpedsIcAy to InstitutionBuilder
* Added IpedsIcPy to InstitutionBuilder
* Added Sec702 to InstitutionBuilder
* Added Sec702School to InstitutionBuilder
* Added Settlement to InstitutionBuilder
* Added Hcm to InstitutionBuilder
* Added Complaint to InstitutionBuilder
* Added Outcome to InstitutionBuilder


app/models/version.rb
```diff
@@ -34,14 +34,14 @@ def self.preview_version_by_time(time)
...
+    if number.present? && Version.find_by(number: number).nil?
+      errors.add(:number, "Version number #{number} doesn't exist")
     end
 
     true
   end
 
   def increment_version
-    self.version = (Version.maximum(:version) || 0) + 1 if version.nil?
+    self.number = (Version.maximum(:number) || 0) + 1 if number.nil?
```
saneshark: should this maybe be done in an initializer? I can't help but think that this is an anti-pattern. You're using a getter method to set an attribute.

app/models/ipeds_ic_ay.rb
```diff
@@ -2,6 +2,8 @@
 class IpedsIcAy < ActiveRecord::Base
   include Loadable, Exportable
 
+  USE_COLUMNS = [:tuition_in_state, :tuition_out_of_state, :books].freeze
```
saneshark: It looks like these constants are sometimes duplicated, in any case, even when they are not. It would be nice if they were stored somewhere more centrally.

Can you create a lib/gibct_constants.rb and then define these there? Then you can invoke them with GIBCT_Constants::SOME_MORE_DESCRIPTIVE_NAME_THAN_USE_COLUMNS

This will have the added benefit of not having this code duplicated.

app/models/ipeds_ic_ay.rb
```diff
@@ -2,6 +2,8 @@
...
+
   MAP = {
```
saneshark: Same can be said about MAP

app/models/weam.rb
```diff
@@ -40,15 +46,15 @@ class Weam < ActiveRecord::Base
...
+  validates :facility_code, :institution, :institution_type_name, presence: true
   validates :bah, numericality: true, allow_blank: true
 
   before_validation :derive_dependent_columns
 
   # Computes all fields that are dependent on other fields. Called in validation because
   # activerecord-import does not engage callbacks when saving
   def derive_dependent_columns
-    self.institution_type = derive_type
+    self.institution_type_name = derive_type
```
saneshark: I think it would be better to do this in an initializer or a class method on the object that initializes a new version of the model or at the very least this should be a private method invoked by an initializer.

brakeman.log
```diff
@@ -0,0 +1,25 @@
+
++BRAKEMAN REPORT+
```
saneshark: I don't think you want this checked in to source control. This should be in gitignore

lib/common/exceptions/base_error.rb
```diff
@@ -11,7 +11,7 @@ def status_code
         errors&.first&.status&.to_i
       end
 
-      private
+      # private
```
saneshark: Why is this commented? These methods should be private.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -5,7 +5,7 @@
   context 'autocomplete results' do
     it 'returns collection of matches' do
       7.times { create(:institution, :contains_harv) }
-      get :autocomplete, term: 'harv'
+      get :autocomplete, term: 'harv', version: 1
```
saneshark: I think in routes.rb the version should also maybe have a default / fallback for when it is not provided?

app/models/institution_builder.rb
```diff
@@ -0,0 +1,416 @@
...
+        SELECT "cross", array_agg(DISTINCT(accreditation_type)) AS at_types
+        FROM accreditations
+        WHERE "cross" IS NOT NULL
+          AND accreditation_type IS NOT NULL
+          AND periods LIKE '%current%'
+          AND csv_accreditation_type = 'institutional'
+        GROUP BY "cross") AS cross_type_arr
+      WHERE cross_type_arr.cross = institutions.cross
+        AND institutions.version = #{version_number};
+    SQL
```
saneshark: This query seems dangerous at first glance. I will want us to make sure all of the columns in questions have appropriate indexes. Is there anyway to optimize these queries? Why do we need the CASE When? can that not be done with ruby ahead of time?

app/models/institution_builder.rb
```diff
@@ -0,0 +1,416 @@
...
+        WHERE "cross" IS NOT NULL
+          AND accreditation_type IS NOT NULL
+          AND (accreditation_status = 'probation' OR accreditation_status = 'show cause')
+          AND periods LIKE '%current%'
+          AND csv_accreditation_type = 'institutional'
+        GROUP BY "cross", accreditation_type) AS cross_status_arr
+      WHERE institutions.cross = cross_status_arr.cross
+        AND institutions.accreditation_type = cross_status_arr.accreditation_type
+        AND institutions.version = #{version_number};
+    SQL
```
saneshark: Similar comment, as above. We need to look at these queries more closely. In particular, there is a ruby gem that helps with identifying database optimizations that can be made. I can't remember the name, but I will research and get back to you on that. That being said, these variety of optimizations could happen in a separate PR.

saneshark (commented):
Mostly concerned with institution_builder.rb

I think this class is doing too much. It might all be necessary, but I think we should consider refactoring it into small components, that way rubocop won't complain either.

app/models/ipeds_ic_ay.rb
patrickvinograd: I disagree...the only exact duplication I see is between the two `ipeds` models. Anyway this seems like something that benefits from code locality - being able to look at the model and see in the same file which columns are actually consumed into the institutions table.

I do think maybe the way these constants are used in institution_builder could be refactored to better encapsulate this repeated pattern:
`columns = Crosswalk::USE_COLUMNS.map(&:to_s).map { |col| %("#{col}" = crosswalks.#{col}) }.join(', ')`

patrickvinograd: E.g. maybe a mixin that generates `columns`, assuming the table naming is consistently discoverable from the ActiveModel class, which I think it should be.

saneshark: I guess my biggest problem with this is USE_COLUMNS and MAP are less than ideal names for the constants. But if you are going to keep them outside of a single constants.rb, then it makes sense to have generic enough that you can use consistently across the various models.

app/models/weam.rb
saneshark: Not suggesting over-riding #initialize. It looks like you're using `before_validation` right now, I'm not sure if that is better than using `after_initialize.` What if I want to manually override an attribute before saving? I might not be able to because before_validation would possibly change my manual override. 

Just something to think about. In any case, I think making `derive_dependent_columns` a private method makes sense.

db/migrate/20170122220228_create_ipeds_ic_ays.rb
```diff
@@ -3,9 +3,9 @@ def change
     create_table :ipeds_ic_ays do |t|
       # Used in the building of DataCsv
       t.string :cross, null: false
-      t.integer :tuition_in_state
-      t.integer :tuition_out_of_state
-      t.integer :books
+      t.float :tuition_in_state
+      t.float :tuition_out_of_state
+      t.float :books
```
ben-USDS-damman: Are these actually float values? I thought they were always whole integers.

db/migrate/20170123202716_create_ipeds_ic_pies.rb
```diff
@@ -3,9 +3,9 @@ def change
     create_table :ipeds_ic_pies do |t|
       # Used in the building of DataCsv
       t.string :cross, null: false
-      t.integer :tuition_in_state
-      t.integer :tuition_out_of_state
-      t.integer :books
+      t.float :tuition_in_state
+      t.float :tuition_out_of_state
+      t.float :books
```
ben-USDS-damman: (Are these actually float values? I thought they were always whole integers.)

db/schema.rb
```diff
@@ -312,13 +220,13 @@
...
     t.string   "vetsuccess_name"
     t.string   "vetsuccess_email"
-    t.string   "credit_for_mil_training"
-    t.string   "vet_poc"
-    t.string   "student_vet_grp_ipeds"
-    t.string   "soc_member"
+    t.boolean  "credit_for_mil_training"
+    t.boolean  "vet_poc"
+    t.boolean  "student_vet_grp_ipeds"
+    t.boolean  "soc_member"
```
ben-USDS-damman: Should all the booleans now be `default: false, null: false`?

db/schema.rb
```diff
@@ -337,15 +245,15 @@
...
+    t.boolean  "online_all"
     t.float    "p911_tuition_fees",                                   default: 0.0
     t.integer  "p911_recipients",                                     default: 0
     t.float    "p911_yellow_ribbon",                                  default: 0.0
     t.integer  "p911_yr_recipients",                                  default: 0
     t.boolean  "accredited",                                          default: false
     t.string   "accreditation_type"
     t.string   "accreditation_status"
-    t.string   "caution_flag"
+    t.boolean  "caution_flag",                                        default: false
```
ben-USDS-damman: +`null: false`?

db/schema.rb
```diff
@@ -337,15 +245,15 @@
     t.float    "tuition_in_state"
     t.float    "tuition_out_of_state"
     t.float    "books"
-    t.string   "online_all"
+    t.boolean  "online_all"
```
ben-USDS-damman: +`null: false, default: false`?

db/schema.rb
```diff
@@ -312,13 +220,13 @@
     t.boolean  "poe",                                                 default: false
     t.boolean  "eight_keys",                                          default: false
     t.boolean  "dodmou",                                              default: false
-    t.boolean  "sec_702",                                             default: false
+    t.boolean  "sec_702"
```
ben-USDS-damman: Is there a reason this one is losing its `default: false`? Should it not be `default: false, null: false`?

db/schema.rb
```diff
@@ -290,16 +194,20 @@
...
+    t.integer  "version",                                                             null: false
     t.string   "institution_type_name"
     t.string   "facility_code"
     t.string   "institution"
     t.string   "city"
     t.string   "state"
     t.string   "zip"
     t.string   "country"
+    t.boolean  "flight",                                              default: false
+    t.boolean  "correspondence",                                      default: false
```
ben-USDS-damman: +2 `null: false`?

db/schema.rb
```diff
@@ -337,15 +245,15 @@
     t.float    "tuition_in_state"
     t.float    "tuition_out_of_state"
     t.float    "books"
```
ben-USDS-damman: Especially here, are these truly float values or are you just making them float to match the other tables?

app/models/version.rb
ben-USDS-damman: The `:max_by_time` scope could be more simply named `:newest`.
The `:created_by` attribute alias should be moved adjacent to the `:user` relation at the top.

spec/models/institution_builder_spec.rb
```diff
@@ -0,0 +1,706 @@
...
+  before(:each) do
+    create :user, email: 'fred@va.gov', password: 'fuggedabodit'
+  end
+
+  describe '#buildable?' do
+    before(:each) do
+      InstitutionBuilder::TABLES.each { |t| create t.name.underscore.to_sym, :institution_builder }
+    end
+
+    it 'is true when all tables rae populated' do
```
ben-USDS-damman: are

app/models/institution_builder.rb
```diff
@@ -0,0 +1,433 @@
...
+    Sec702School, Sec702, Settlement, Sva, Vsoc, Weam
+  ].freeze
+
+  def self.buildable?
+    TABLES.map(&:count).reject(&:positive?).blank?
+  end
+
+  def self.default_timestamps_to_now
+    query = 'ALTER TABLE institutions ALTER COLUMN updated_at SET DEFAULT now(); '
+    query += 'ALTER TABLE institutions ALTER COLUMN created_at SET DEFAULT now();'
```
ben-USDS-damman: I couldn't easily verify that specs covered this. Are we good?

.gitignore
```diff
@@ -15,7 +15,7 @@
 /data/*
 
 # Ignore seeds
-/db/seeds.rb
+# /db/seeds.rb
```
saneshark: just remove this line altogether maybe, we won't run it in deployments.

app/models/accreditation.rb
```diff
@@ -35,10 +35,10 @@ class Accreditation < ActiveRecord::Base
   # if the accrediting agency is the "New England Medical Association", then
   # the accreditation is 'Regional'.
   ACCREDITATIONS = {
-    'REGIONAL' => [/middle/i, /new england/i, /north central/i, /southern/i, /western/i],
-    'NATIONAL' => [/career schools/i, /continuing education/i, /independent colleges/i,
+    'regional' => [/middle/i, /new england/i, /north central/i, /southern/i, /western/i],
+    'national' => [/career schools/i, /continuing education/i, /independent colleges/i,
```
ben-USDS-damman: instead of an array of regexes can this be compiled into a single regex?

ben-USDS-damman:
Per conversation just now all my concerns have been addressed.

app/models/accreditation.rb
saneshark: ignoring this for current PR.

.gitignore
saneshark: ignoring this for current PR.

spec/controllers/v0/institutions_controller_spec.rb
saneshark: can be addressed in future PR.

app/models/version.rb
mphprogrammer: Done.

app/models/weam.rb
mphprogrammer: Done.

lib/common/exceptions/base_error.rb
mphprogrammer: This was vets-api related, will uncomment next PR if ok with @ben-USDS-damman

spec/controllers/v0/institutions_controller_spec.rb
mphprogrammer: This was vets-api related @ben-USDS-damman

saneshark:
OK, LGTM, for now... @mphprogrammer is opening up a ticket or series of tickets to address some of the points brought up in future PRs.

saneshark:
@mphprogrammer please reference this PR, even though it will be closed in your issues that you create.

PR #190, opened 2017-02-13
Dashboards

saneshark:


.ruby-version
```diff
@@ -1 +1 @@
-2.3.0
+2.3.1
```
ben-USDS-damman: Are we going to bump vets-api also?

saneshark: Oh, I didn't realize vets-api was still so out of date on ruby version. It really shouldn't matter.

saneshark:
Started separate branch / seperate PR based on this branch.

PR #192, opened 2017-02-14
API enhancements and fixes

ben-USDS-damman:
* CORS support for local use with vets-website
* rm load_csv task and old csv
* changes to JSON responses, json schema changes
* seeded data is a production version
* refactor Version model
* sensible default versioning in InstitutionController
* README updates


saneshark:
@ben-USDS-damman can we get specs passing before I review?

Jenkinsfile
```diff
@@ -0,0 +1,57 @@
+pipeline {
```
saneshark: @patrickvinograd tagging you in on the Jenkins related changes here.

Jenkinsfile.deploy
```diff
@@ -7,7 +7,9 @@ def notify(message, color='good') {
 }
 
 pipeline {
```
saneshark: @patrickvinograd and here

app/controllers/v0/api_controller.rb
```diff
@@ -3,6 +3,10 @@ module V0
   class ApiController < ApplicationController
     skip_before_action :authenticate_user!
 
+    def cors_preflight
+      head(:ok)
+    end
```
saneshark: We might need to revisit this when we add authentication component between vets-api and this data service, but this is good for now.

Jenkinsfile
patrickvinograd: All good, these are both things Ben pulled in from master.

Jenkinsfile.deploy
patrickvinograd: All good, these are both things Ben pulled in from master.

db/seeds/02_data.rb
```diff
@@ -1,3 +1,9 @@
+puts 'Destroy previous users ... '
```
patrickvinograd: This looks like a mis-merge between what I added last night - the user-seeding part of this belongs only in `01_users.rb`

patrickvinograd (changes requested):
Please fix mis-merge of db/seeds/02_data.rb. May have been partially my fault - git showed that file as a copy of original seeds.rb...

Jenkinsfile
patrickvinograd: Actually I take that back -we no longer want the part of this that's cloning gi-bill-comparison-tool repo, and that is what is causing jenkins to fail - because its winding up running rubocop on that subdirectory.

patrickvinograd: Should take out all the stuff related to that subdirectory.

ben-USDS-damman: I'm confused...I did not pull anything from `master`; only rebased onto the latest`rewrite-1`. Is this a problem that should be fixed in `rewrite-1`?

ben-USDS-damman:
Just confirmed that specs are lint are passing locally.

db/seeds/02_data.rb
ben-USDS-damman: Whoops, no probably my mistake.

Jenkinsfile
ben-USDS-damman: corrections made

patrickvinograd (approved):

PR #193, opened 2017-02-14
Seed refactoring to enable devops use case

patrickvinograd:
Marc, I believe you may be reworking seeds.rb - if so can you incorporate this into your change. Or if you are just further updating CSVs then merging this would be helpful. Reason being that for AWS deploys we want to overwrite the user seeding, and it's much easier if it is in its own file rather than having to patch part of a larger seeds.rb

`rake db:seed` continues to work as before with this change.


mphprogrammer:
Looks good Patrick - I will merge this tomorrow after this next pr ... Thanks!

mphprogrammer:
LGTM

PR #198, opened 2017-02-16
Csvs and fixed seed

mphprogrammer:


mphprogrammer:
@ben-USDS-damman These are the production CSVs sorry it took so long, but somehow my branch inherited funky Jenkins files from mast, and I had to clear them out.

ben-USDS-damman:
I just chased my tail for almost an hour on a bizarre ENV var loading problem, so I know the feeling. _sigh_

LGTM!

PR #200, opened 2017-02-16
Rewrite 1 dashboards

mphprogrammer:
* Added Upload Table
* Added Upload Controller
* Added Upload Index View
* Adjusted seeds.rb to work with uploads


app/helpers/application_helper.rb
```diff
@@ -17,4 +17,26 @@ def li_active_class(path, method = 'GET')
...
+          end
+        end
+      )
+    end
+  end
+
+  # Temporary placeholder for InstitutionBuilder.buildable?
+  def buildable?
+    true
+  end
```
saneshark: should this be removed?

app/views/dashboards/index.html.erb
```diff
@@ -0,0 +1,39 @@
+<%= render partial: 'introduction',
+    locals: {
+      new_link: buildable?, export_link: buildable?, list_link: false,
+      push_staging_link: buildable?, push_prod_link: buildable?
+    }
+%>
```
saneshark: I see ApplicationHelper#buildable? being used here, but if it is always true, then maybe this can be cleaned up a bit. I'm cool leaving it for now though.

lib/csv_helper/loader.rb
```diff
@@ -6,26 +6,31 @@ module Loader
...
-    def load_from_csv(filename, options)
+    def load_from_csv(file, options)
       records = { valid: [], invalid: [] }
 
       # Since row indexes start at 0 and spreadsheets on line 1,
       # add 1 for the difference in indexes and 1 for the header row itself.
       row_offset = 2 + (options[:skip_lines] || 0)
 
-      SmarterCSV.process(filename, merge_options(options)).each.with_index do |row, i|
+      SmarterCSV.process(file, merge_options(options)).each.with_index do |row, i|
```
saneshark: So I see that here we're doing validations on each record of the CSV. That's good. Are we certain then that active-record-import is not running validations again on top of this and is in fact doing a bulk insert of the records array? Does it only save the valid records or does it save all of them?

saneshark (approved):
Mostly just questions I have, but everything looks good enough to merge.

app/helpers/application_helper.rb
mphprogrammer: With the Upload model, I can collapse this into a single query returning only those tables that are empty. Anyway I think I can ...

mphprogrammer: But come too think of it, with seeds, it will always be buildable, so I'll yank this out now ...

app/views/dashboards/index.html.erb
mphprogrammer: I'll yank it out now

lib/csv_helper/loader.rb
mphprogrammer: AR-import is explicitly told not to validate . It only saves the valid records, in the smartercsv loop:
```
  record.errors.any? ? records[:invalid] << record : records[:valid] << record
```

The instruction to use AR-import only uses records[:valid]
```
klass.import records[:valid], validate: false, ignore: true
```

saneshark: you can `tail -f log/development.log` while running `bundle exec rake db:seed` and hopefully we're doing a bulk insert, and not each record individually. This should speed up seeding the database dramatically.

saneshark:
LGTM. Assuming build passes, merge away!

mphprogrammer:
Merging per approval...

PR #202, opened 2017-02-16
Remove some unique index constraints

patrickvinograd:


saneshark:
@patrickvinograd LGTM, if you have strong confidence merge away. Would be nice to get Marc's input, but we can always add it back if needed. Would like to get a deploy soon so I can add the VCR cassettes to vets-api client.

mphprogrammer:
@patrickvinograd I think you've got them. I find find any others, I'll include those in the next pr

PR #203, opened 2017-02-16
Add no-auth status endpoint outside of v0/ namespace

patrickvinograd:
In support of devops monitoring. While there are some non-authenticated API endpoints already, I want to be able to hide all of the /v0/ namespace as accessible internally only (mediating all API operations through vets-api). Whereas the status endpoint needs to be accessible externally for blackbox monitoring.

### DIFF
--- a/app/controllers/status_controller.rb
+++ b/app/controllers/status_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+class StatusController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  def status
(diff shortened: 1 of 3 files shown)


saneshark:
LGTM, this makes sense. I wonder if it could just inherit from ApiController though and have the authenticate_user skipped via inheritance (minor nitpick).

patrickvinograd:
Seemed a little wonky to have something outside of the v0 module inheriting from something within, but feel free to refactor later if it's bugging you.

PR #204, opened 2017-02-16
Update header

patrickvinograd:
Actually changed the title rather than just fixing the typo - this is not the comparison tool, so this made more sense to me.


saneshark (approved):
LGTM

PR #206, opened 2017-02-22
Facet counts and other suboptimal enhancements

ben-USDS-damman:
Wherein all API responses return verbose data versioning metadata.
```json
...
"meta": {
  "version": {
    "number": 1,
    "created_at": "2017-02-16T05:23:49.937Z",
    "preview": false
  }
}
```

And search results include filter counts to support the search interface.
```json
...
"facets": {
  "type": {
    "school": 3,
    "employer": 0
  },
  "type_name": {
    "for profit": 2,
    "public": 1
  },
  "state": {
    "ga": 1,
    "mn": 2
  },
  "country": {
     "usa": 3
  },
  "caution_flag": {
    "true": 3
  },
  "student_vet_group": {
    "false": 3
  },
  "yellow_ribbon_scholarship": {
    "false": 3
  },
  "principles_of_excellence": {
    "false": 1,
    "true": 2
  },
  "eight_keys_to_veteran_success": {
    "false": 3
  }
}
```

Upon further reflection the facet objects should probably be arrays.

### DIFF
--- a/app/controllers/v0/api_controller.rb
+++ b/app/controllers/v0/api_controller.rb
@@ -13,7 +13,7 @@ module V0
     # Newest production data version assumed when version param is undefined
     def resolve_version
       v = params[:version]
-      version = v.present? ? Version.find_by_number(v) : Version.default_version
+      version = v.present? ? Version.find_by(number: v) : Version.default_version
       raise ActiveRecord::RecordNotFound, "Version #{v} not found" unless version.try(:number)
       @version = {
         number: version.number,
--- a/app/controllers/v0/api_controller.rb
+++ b/app/controllers/v0/api_controller.rb
@@ -14,7 +14,7 @@ module V0
     def resolve_version
       v = params[:version]
       version = v.present? ? Version.find_by(number: v) : Version.default_version
-      raise ActiveRecord::RecordNotFound, "Version #{v} not found" unless version.try(:number)
+      raise Common::Exceptions::InvalidFieldValue, "Version #{v} not found" unless version.try(:number)
       @version = {
         number: version.number,
         created_at: version.created_at,
--- a/app/controllers/v0/api_controller.rb
+++ b/app/controllers/v0/api_controller.rb
@@ -2,12 +2,25 @@
 module V0
   class ApiController < ApplicationController
     skip_before_action :authenticate_user!
+    before_action :resolve_version
 
     def cors_preflight
       head(:ok)
     end
 
     private
+
+    # Newest production data version assumed when version param is undefined
+    def resolve_version
+      v = params[:version]
+      version = v.present? ? Version.find_by_number(v) : Version.default_version
+      raise ActiveRecord::RecordNotFound, "Version #{v} not found" unless version.try(:number)
+      @version = {
+        number: version.number,
+        created_at: version.created_at,
+        preview: version.preview?
+      }
+    end
 
     rescue_from 'Exception' do |exception|
       log_error(exception)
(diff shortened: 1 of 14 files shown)


db/schema.rb
```diff
@@ -154,10 +154,10 @@
     t.datetime "updated_at",    null: false
   end
 
-  add_index "crosswalks", ["cross"], name: "index_crosswalks_on_cross", using: :btree
+  add_index "crosswalks", ["cross"], name: "index_crosswalks_on_cross", unique: true, using: :btree
```
mphprogrammer: We should take the unique constraint off the index - I have found cases where crosswalk information is not unique

ben-USDS-damman:
Made schema.rb match base branch w/r/t unique crosswalks indexes that were removed sans migration, addressing https://github.com/department-of-veterans-affairs/gibct-data-service/pull/206#discussion_r102633641

db/schema.rb
ben-USDS-damman: corrected

ben-USDS-damman:
`rake ci` passes locally with flying colors - what's upsetting jenkins?

ben-USDS-damman:
Updated PR description above and removed [WIP]

ben-USDS-damman:
FYI search requests against my `Rails.env.development` seem to hover around around 350ms

patrickvinograd:
The Jenkinsfile was screwy - fixed now v2-refactor, if you want to pull that in.

app/controllers/v0/api_controller.rb
```diff
@@ -2,13 +2,26 @@
...
       head(:ok)
     end
 
     private
 
+    # Newest production data version assumed when version param is undefined
+    def resolve_version
+      v = params[:version]
+      version = v.present? ? Version.find_by(number: v) : Version.default_version
+      raise ActiveRecord::RecordNotFound, "Version #{v} not found" unless version.try(:number)
```
patrickvinograd: This might be better as a Common::Exceptions::RecordNotFound, or even better Common::Exceptions::InvalidFieldValue since it's a parameter that can't be found as opposed to a top-level resource.

patrickvinograd (approved):
Overall LGTM, don't want this to block further progress. 

One minor suggestion about an exception type. 

I'm curious to see how the filter count stuff affects the queries and if we can optimize that, but easiest to do that as a follow-up change. 

You should rebase this on latest v2-refactor to get Jenkins to pass.

app/controllers/v0/api_controller.rb
ben-USDS-damman: Good point. I will rebase and change the exception type.

app/controllers/v0/institutions_controller.rb
```diff
@@ -42,18 +41,16 @@ def show
...
          yellow_ribbon_scholarship principles_of_excellence
-         eight_keys_to_veteran_success caution).each do |k|
+         eight_keys_to_veteran_success caution country).each do |k|
         params[k].try(:downcase!)
       end
-      %i(country state).each do |k|
-        params[k].try(:upcase!)
-      end
+      params[:state].try(:upcase!)
       params[:name].try(:strip!)
```
saneshark: So I don't think it's good practice to modify params. It would be better to use an instance variable. As a user, if I pass in state as lower case, if it fails validation, I should see the params exactly as I provided them.

ben-USDS-damman:
1. changed that exception type
1. `#search_results` now relies on a memoized clone of params instead of mutating request params directly
1. merging since jenkins and other checks are green

@saneshark @patrickvinograd

PR #207, opened 2017-02-22
V2 refactor uploads

mphprogrammer:
Completed uploading UI


patrickvinograd:
Marc - the title of this PR says "In Progress" - is that still true or are you looking for review/merge here?

mphprogrammer:
@patrickvinograd I still have a few things to add to the dashboard, but it can be reviewed now. This entire PR is about the Admin UI.

mphprogrammer:
@patrickvinograd Jenkins does not appear to be running after I had merged with v2-refactor (although that may be a coincidence). Have I screwed something up?

patrickvinograd:
I don't think it's any problem on your end. Looking into it.

patrickvinograd:
Jenkins build was triggered and passed for your latest commit, as you can see here:
http://jenkins.vetsgov-internal/blue/organizations/jenkins/department-of-veterans-affairs%2Fgibct-data-service/activity
Still figuring out why it's not showing up here on github but that should not prevent merging in the meantime once this is reviewed.

patrickvinograd:
LGTM.

app/controllers/concerns/flashable.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+module Flashable
+  extend ActiveSupport::Concern
+
+  def message_for_notice(invalid_records)
+    msg = "Uploading #{upload_params[:csv_type]} succeeded"
+    msg + (invalid_records.present? ? ' with warnings:' : '.')
+  end
```
saneshark: NITPICK: So there are, as I recall a couple different versions of flash messages. One is flash[:notice] and the other is flash[:warn], and they can be setup to have different styles around them depending on how bootstrap css classes are defined. Might this be a better alternative?

app/controllers/dashboards_controller.rb
```diff
@@ -1,4 +1,30 @@
...
 class DashboardsController < ApplicationController
-  def index; end
+  include Flashable
+
+  def index
+    @uploads = Upload.last_uploads
+  end
+
+  def build
+    @version = InstitutionBuilder.run(current_user)
```
saneshark: This operation is going to take a long time to run isn't it? This action is most likely going to timeout well before this action returns. I think you're going to need to run it in a thread and just return a status indicating that it is running, at least until such time we can setup a worker with id, and a polling mechanism of some sort to see when the job is complete.

@patrickvinograd might need to fine tune the timeout for this seeing as this operation could take a long time.

Also I don't see any rescue blocks or anything, what would result if this encounters an error?

app/controllers/dashboards_controller.rb
```diff
@@ -1,4 +1,30 @@
...
+    klass = csv_model(params[:csv_type])
+    csv_data = klass.export
+    raise StandardError, 'No data to export.' if csv_data.blank?
+
+    respond_to do |format|
+      format.csv { send_data csv_data, type: 'text/csv' }
+    end
+  rescue StandardError => e
+    redirect_to dashboards_path, alert: e.message
+  end
```
saneshark: So I'm not crazy about this pattern. I would have preferred using accepts nested attributes somehow, but I think that will have to happen in a future refactor. The reason being it is a generic controller action, not resourceful.

app/controllers/uploads_controller.rb
```diff
@@ -1,10 +1,60 @@
...
+    @upload.update(ok: data.present? && data.ids.present?)
+    data
+  end
+
+  def original_filename
+    @f ||= upload_params[:upload_file].try(:original_filename)
+  end
+
+  def defaults
+    @defaults ||= YAML.load_file(Rails.root.join('config', 'csv_file_defaults.yml'))
```
saneshark: Should this be a constant maybe? I'm not sure we need to initialize the csv file deaults every time this controller is initialized. Just load it at the application level and keep it in memory.

app/controllers/uploads_controller.rb
```diff
@@ -1,10 +1,60 @@
...
+    @upload.skip_lines = defaults[csv_type || 'generic']['skip_lines']
+  end
+
+  def create
+    @upload = Upload.create(merged_params)
+    data = load_csv
 
-    respond_to do |format|
-      format.html
+    if @upload.ok
```
saneshark: .ok?

app/controllers/uploads_controller.rb
```diff
@@ -1,10 +1,60 @@
...
+    redirect_to uploads_path, alert: ["Upload with id: '#{params[:id]}' not found"] unless @upload
+  end
+
+  private
+
+  def load_csv
+    return unless @upload.persisted?
+
+    file = @upload.upload_file.tempfile
+    data = @upload.csv_type.constantize.load(file, skip_lines: @upload.skip_lines.try(:to_i))
```
saneshark: what happens if skip_lines receives nil from that try(:to_i) ?

app/controllers/uploads_controller.rb
```diff
@@ -1,10 +1,60 @@
...
+
+    @upload.update(ok: data.present? && data.ids.present?)
+    data
+  end
+
+  def original_filename
+    @f ||= upload_params[:upload_file].try(:original_filename)
```
saneshark: should we have a default for the file name when original_filename is nil?

app/controllers/concerns/flashable.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def errors_for_alert(invalid_records)
+    error_list = invalid_records[0, 15].map do |record|
+      line = record.errors.delete(:base)
```
saneshark: why do you want to remove any errors? Shouldn't this be done at the model level maybe by not calling the validation that adds whatever the error is in the first place? I can't help to think that maybe this is a model level concern and not a controller one.

saneshark (commented):
mostly minor recommendations.

app/controllers/dashboards_controller.rb
patrickvinograd: We've tweaked timeouts for GIDS already at rev proxy and load balancer layers so unless this is vastly slower than the current production's export/data push operations I think it will succeed. If not, it will result in a "Request timed out" error (probably from nginx) but the operation will continue running in the GIDS app.

app/controllers/concerns/flashable.rb
mphprogrammer: Flashable is a poorly named mixin for formatting the error messages. I've been using flash.notice for success, or success with warnings, and flash.alert for errors. If there is only a notice it goes into flash.notice, if there are errors and notices they go into flash.alert. I could use flash.warn for success uploading.

As an aside, I looked into rolling back uploads on validation errors, but there is no way for us to easily differentiate between a valid error or an error caused by a free-text row embedded amongst data (see eight_key.csv for example). Unless we require VA Edu users to clean up the CSVs before using, we should accept the upload but display the errors.

app/controllers/concerns/flashable.rb
mphprogrammer: The errors are removed during display after uploading. I could move the display errors functionality into the model, but once displayed to the user, the error information and invalid records are disposable.

app/controllers/dashboards_controller.rb
mphprogrammer: This rewrite should be as fast or faster than the current queries, and should not pose any burden. The problem with building is that we have a design that does not require any particular order in uploading. Otherwise we would have to require the user to upload new Weam and Crosswalk CSVs (if present) prior to any other CSV. That would be a different functionality.

app/controllers/dashboards_controller.rb
mphprogrammer: 👍

app/controllers/uploads_controller.rb
mphprogrammer: ok is a column, but I forgot to provide a .ok? method will add.

app/controllers/uploads_controller.rb
mphprogrammer: SmarterCsv uses its default value of 0

app/controllers/uploads_controller.rb
mphprogrammer: I think that original_filename would only be nil if the upload_file (`ActionDispatch::Http::UploadedFile`), is itself not valid. I wouldn't know where in the system rails would place the temp uploaded file. Also, the names of the CSVs may differ on every upload for the same csv (some have date and time encoded in their names).

app/controllers/uploads_controller.rb
mphprogrammer: Yes it should, I will modify

app/controllers/concerns/flashable.rb
saneshark: ok. lets revisit that later. eventually we should require them to fix the CSV, and with the line number based messages I think this is a good path forward. Perhaps in the warning you can say something along the lines of: 'uploading a file with invalid data will no longer work in the future, please review the warnings so that file can be better prepared for uploading in the future.'

app/controllers/dashboards_controller.rb
saneshark: @mphprogrammer what is returned if the build fails? I don't see any error handling.

saneshark:
@mphprogrammer my comments are mostly nitpicks that can be addressed later. Even the error handling comment I made, it's just for admins so its not the end of the world if it isn't perfect. Eventually i would like to get us to a place where admins are forced to fix invalid records too, but we're not there yet. We can discuss later.

LGTM feel free to merge

app/controllers/dashboards_controller.rb
mphprogrammer: Will add a rescue in next PR.

db/migrate/20170124095318_create_complaints.rb
```diff
@@ -23,34 +23,6 @@ def change
...
-      t.integer :complaints_marketing_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_accreditation_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_degree_requirements_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_student_loans_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_grades_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_credit_transfer_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_jobs_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_transcript_by_ope_id_do_not_sum, default: 0
-      t.integer :complaints_other_by_ope_id_do_not_sum, default: 0
-
```
ben-USDS-damman: After this gets merged I think it's important we stop updating migrations in place and start producing actual migrations. It makes reviewing schema changes easier and creates fewer problems updating our local dev environments.

app/controllers/concerns/flashable.rb
mphprogrammer: For some files, like Eight-keys there would be 120 "Validation Errors" that aren't really errors. I limit the number of errors reported to the user on the results because 120 would be a lot to display, and because of the nature of the errors being 'free text' embedded in the data rows of the CSV. The error being removed contains the line number of the row causing it and is essentially for display purposes only, and not really part of the model.

ben-USDS-damman (approved):
LGTM other than comment about doing real migrations in the future

app/controllers/concerns/flashable.rb
mphprogrammer: I like that, will amend error messages in next PR.

PR #209, opened 2017-03-01
Rename arf.csv to arf_gi_bill.csv to match model

patrickvinograd:
db:seed was broken due to filename/model mismatch


mphprogrammer:
Thought I did that, but must have dropped the ball ... LGTM and merging

PR #210, opened 2017-03-02
V2 refactor export

mphprogrammer:
Adds Importing, Exporting, and push to production of Institution data.
Wraps build in a transaction


mphprogrammer:
TODO: Add specs

app/controllers/dashboards_controller.rb
```diff
@@ -22,9 +31,29 @@ def export
...
+    if version.blank?
+      flash.alert = 'No preview version available'
+    else
+      pv = Version.create(number: version.number, production: true, user: current_user)
+
+      if pv.persisted?
+        flash.notice = 'Production data updated'
+      else
+
+        flash.alert = "Production data not updated: #{version.inspect} #{pv.errors.inspect}"
```
saneshark: The admins are all trusted people obviously right? .inspect might include more data than necessary, perhaps we log this, but show a more generic message?

app/controllers/dashboards_controller.rb
```diff
@@ -22,9 +31,29 @@ def export
...
+      end
+    end
+
+    redirect_to dashboards_path
```
saneshark: do you really want to redirect to dashboards_path if there is an error? I would think you would want to render the same page with the specified params in case they want try again.

app/controllers/dashboards_controller.rb
```diff
@@ -22,9 +31,29 @@ def export
...
+  end
+
   private
 
   def csv_model(csv_type)
+    return Institution if csv_type == 'Institution'
```
saneshark: I think we're going to want specs around both of these just in case something changes. It might be a pain in the ass to write it as request specs, so this could be a code smell. Possibly consider putting this in a model that you can test at unit level.

app/models/institution_builder.rb
```diff
@@ -49,13 +49,21 @@ def self.run_insertions(version_number)
...
-    version
+        default_timestamps_to_now
+        run_insertions(version.number)
+        drop_default_timestamps
+      end
+    rescue StandardError => e
+      error_msg = e.message
+    end
+
+    { version: Version.preview_version, error_msg: error_msg }
```
saneshark: what will Version.preview_version be if error is raised?

app/models/institution_builder.rb
```diff
@@ -49,13 +49,21 @@ def self.run_insertions(version_number)
...
+    begin
+      Institution.transaction do
+        version = Version.create!(production: false, user: user)
 
-    version
+        default_timestamps_to_now
+        run_insertions(version.number)
+        drop_default_timestamps
+      end
+    rescue StandardError => e
```
saneshark: Do we have to rescue from StandardError? this makes debugging very difficult. Lets at least make sure we're logging something when this happens. But if we can get a more specific error to rescue from it will be better.

app/models/institution_builder.rb
```diff
@@ -117,49 +125,39 @@ def self.add_accreditation(version_number)
...
-        AND institutions.version = #{version_number};
+      UPDATE institutions SET
+        accreditation_type = accreditations.accreditation_type
+      FROM accreditations
+      WHERE institutions.cross = accreditations.cross
+        AND accreditations.cross IS NOT NULL
+        AND accreditations.periods LIKE '%current%'
+        AND accreditations.csv_accreditation_type = 'institutional'
+        AND institutions.version = #{version_number}
+        AND accreditations.accreditation_type =
```
saneshark: do we have some performance metrics around this query vs what it replaces? This has a lot more conditional branches from the looks of it.

app/models/institution_builder.rb
```diff
@@ -117,49 +125,39 @@ def self.add_accreditation(version_number)
...
-        AND institutions.version = #{version_number};
+      UPDATE institutions SET
+        accreditation_status = accreditations.accreditation_status
+      FROM accreditations
+      WHERE institutions.cross = accreditations.cross
+        AND institutions.accreditation_type = accreditations.accreditation_type
+        AND accreditations.cross IS NOT NULL
+        AND accreditations.periods LIKE '%current%'
+        AND accreditations.csv_accreditation_type = 'institutional'
+        AND institutions.version = #{version_number}
```
saneshark: same comment about performance.

lib/csv_helper/exporter.rb
```diff
@@ -18,10 +18,17 @@ def generate(csv_headers)
...
         csv << csv_headers.values
 
-        klass.find_each do |record|
+        set_scope.find_each do |record|
           csv << csv_headers.keys.map { |k| record[k] }
         end
       end
     end
+
+    def set_scope
```
saneshark: is scope the best name to use? It does make sense, but might be confusing seeing as scope has semantical meaning in Rails.

lib/csv_helper/loader.rb
```diff
@@ -18,20 +24,34 @@ def load(file, options = {})
...
-        record.errors.any? ? records[:invalid] << record : records[:valid] << record
+        save_record_to_records(records, record)
+      end
+
+      records
+    end
+
+    def load_from_csv_with_version(file, records, options)
+      version = Version.preview_version
+      row_offset = 2 + (options[:skip_lines] || 0)
```
saneshark: 2 appears to be a magic number, can we specify a constant and use the constant here?

saneshark (dismissed):

app/controllers/dashboards_controller.rb
mphprogrammer: Yup, this was a miss on my part, forgot to change it from a dev message to a user message.

app/controllers/dashboards_controller.rb
mphprogrammer: Actually, that was intentional. Push has no parameters, it merely makes the latest preview version into a production version. The versioning is really abstracted away from the user, and I wanted the push to be simply a pass fail thing for them

app/controllers/dashboards_controller.rb
mphprogrammer: I think that is covered by this spec:
```
    it 'redirects to index on error' do
      expect(get(:export, csv_type: 'BlahBlah', format: :csv)).to redirect_to(action: :index)
      expect(get(:export, csv_type: 'Weam', format: :xml)).to redirect_to(action: :index)
    end
```
Anything that is not in the InstitutionBuilder::TABLES or 'Institution' should return nil, and that will trip a`raise StandardError, 'No data to export.'`. I could specialize the error message to say something along the lines of Non-supported CSV...

app/models/institution_builder.rb
mphprogrammer: These would be POSTGRES DB returned errors, and I could look into seeing what is returned ...

app/models/institution_builder.rb
mphprogrammer: The current preview version, rather than a new preview version.

app/models/institution_builder.rb
mphprogrammer: This is the old way of processing accreditation, makes the overall build faster by about a minute and a half. The new way I had written collapsed the separate 3 queries into a single query of a few lines, but was horrendous time-wise. The case-whens in PG are not really useful timewise.

app/models/institution_builder.rb
mphprogrammer: See above...

lib/csv_helper/exporter.rb
mphprogrammer: Yes, I'll change that.

lib/csv_helper/loader.rb
mphprogrammer: Yes, I'll change that.

app/models/institution_builder.rb
saneshark: If you put that in a Transaction block, you can rescue from ActiveRecord::Rollback

app/controllers/dashboards_controller.rb
saneshark: yea, and don't shy away from creating your own error class that inherits from StandardError when doing rescues. It's a lot better than rescuing from StandardError when that need arises.

app/controllers/dashboards_controller.rb
```diff
@@ -7,7 +7,16 @@ def index
...
+    results = InstitutionBuilder.run(current_user)
+
+    @version = results[:version]
+    @error_msg = results[:error_msg]
+
+    if @error_msg.present?
+      flash.alert = "Preview Data not built: #{@error_msg}"
+    else
+      flash.notice = "Preview Data (#{@version.number}) built successfully"
+    end
```
saneshark: So this is one reason I am not crazy about custom actions like "build" and prefer just the standard resource based actions -- to me this is really unclear which template will be rendered. 

First of all, this should probably be a POST or PUT (is it?) Any time you are modifying data you should have POST or PUT (not GET). 

Does build have its own template that gets rendered? I would think you get redirected back to dashboards / index but that doesn't appear to be implicit. Generally speaking when a non-idempotent action results in an error, the template that called gets rendered again with the errors. When it is successful, then you get redirected.

You might want to use flash.now with render_and_return , and flash.notice redirect_to as the else case,

app/controllers/dashboards_controller.rb
```diff
@@ -22,9 +31,28 @@ def export
...
+      pv = Version.create(number: version.number, production: true, user: current_user)
+
+      if pv.persisted?
+        flash.notice = 'Production data updated'
+      else
+        flash.alert = 'Production data not updated, remains at previous production version'
+      end
+    end
+
+    redirect_to dashboards_path
```
saneshark: Similar comment here. Since push is a non-idempotent action, this should be a POST or PUT request. It is also unclear to me why we would redirect to dashboards path and essentially clear the params provided when we have an "error" -- I would think in your else clause when the error condition results you would do a render instead of redirect.

config/routes.rb
```diff
@@ -10,6 +10,7 @@
   get '/dashboards' => 'dashboards#index'
   get '/dashboards/build' => 'dashboards#build', as: :dashboard_build
   get '/dashboards/export/:csv_type' => 'dashboards#export', as: :dashboard_export, defaults: { format: 'csv' }
+  get '/dashboards/push' => 'dashboards#push', as: :dashboard_push
```
saneshark: this is a non-idempotent action -- it should not be a GET request.

config/routes.rb
```diff
@@ -10,6 +10,7 @@
   get '/dashboards' => 'dashboards#index'
   get '/dashboards/build' => 'dashboards#build', as: :dashboard_build
```
saneshark: this is a non-idempotent action and should not be a GET request

PR #211, opened 2017-03-03
Update rubyzip gem due to CVE

patrickvinograd:


patrickvinograd:
N/m, Marc already has this in his branch.

PR #218, opened 2017-03-07
Match up header conversion with SmarterCSV behavior

patrickvinograd:
SmarterCSV converts dashes to underscores in headers. We have a dash in the weams header non-college degree indicator. We were not mapping accordingly in csv_helper/loader.rb, causing this column to not get mapped in AWS dev environment - all entries were nil. This in turn caused thousands of institutions to have approved? = false.

I still can't account for why this didn't happen in our local environments, but pushing this fix to dev and re-uploading weams.csv made the database look correct.


mphprogrammer:
Yup LGTM

PR #222, opened 2017-03-10
V2 refactor fixes #221

### LINKED ISSUE #221: Only 12 schools DODMOU identified as DODMOU true...
@saintsoup52 Has identified a major difference between the old and new GIDS regarding the DODMOU column. 

1. Missing 5,126 DOD Approved schools in new GIDS
2. Only 12 schools are showing as DOD approved

I've checked the source CSV downloaded from the old GIDS in production. There are only 12 schools that would show as false. I inadvertently omitted the negation in the dodmou computation. The fix will be up shortly.

screen shot 2017-03-10 at 12 47 17 pm

@saintsoup52 Is there anyone I can contact for the actual files used to populate the old GIDS in production?

mphprogrammer:
Fixed missing ! that inverted the logic of to_dodmou.

### DIFF
--- a/app/models/mou.rb
+++ b/app/models/mou.rb
@@ -26,7 +26,7 @@ class Mou < ActiveRecord::Base
   end
 
   def to_dodmou
-    STATUSES.include?(status)
+    !STATUSES.include?(status)
   end
 
   def to_dod_status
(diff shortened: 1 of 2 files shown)


saneshark (approved):
LGTM

PR #223, opened 2017-03-13
Removed outcome defaults - Please Review

mphprogrammer:
1. Fixes issue with NULL values reported as 0 in outcome rates.
2. Fixes issue with Production GIDS and new GIDs - Some Accreditation caution flags overwritten by Sec702 caution flag logic (rather than ORing to existing caution flags...)


mphprogrammer:
@saintsoup52 @saneshark Did find one issue with caution flag reason Accreditation (Probation/Show Cause) and caution flag not being set. This exists in both Old and New GIDS

mphprogrammer:
@saintsoup52 @saneshark Disregard last remark - both old and new gids agree completely regarding caution flags and caution flag reasons.  Ordering of caution_flag_reasons and spaces after commas, and text case caused whatever differences were found - can find nothing else wrong at the moment...

saneshark:
@mphprogrammer This LGTM please feel free to merge when you are ready.

saneshark (approved):

saneshark (changes requested):
Dont change existing migration

saneshark:
@mphprogrammer I spoke too fast. Please don't modify existing migrations. Create a new migration to change defaults instead.

db/schema.rb
```diff
@@ -853,17 +853,17 @@
...
-    t.float    "persistance_rate_veteran_otb", default: 0.0
-    t.float    "graduation_rate_veteran",      default: 0.0
-    t.float    "transfer_out_rate_veteran",    default: 0.0
+    t.string   "facility_code",                null: false
+    t.float    "retention_rate_veteran_ba"
+    t.float    "retention_rate_veteran_otb"
+    t.float    "persistance_rate_veteran_ba"
+    t.float    "persistance_rate_veteran_otb"
+    t.float    "graduation_rate_veteran"
+    t.float    "transfer_out_rate_veteran"
```
saneshark: @mphprogrammer missing one here

saneshark (approved):

saneshark:
LGTM

saintsoup52:
Not sure where we are with this... but the latest "verify file" (see link below) is still very different then the one in the old DS.

mphprogrammer:
@saintsoup The old gids data used in verify I produced myself from my local copy of the current Data service. Along the errors you had mentioned, verify.xslx didn't find any differences between institution_new.csv and institution_old.csv, what differences in these csvs did you find?

mphprogrammer:
@saintsoup52 also, there was no "link below" :-)

saintsoup52:
I just used the data dashboard "data validation" worksheet to compare the final products and those screenshots I sent you were the results. 

I created the data validation worksheet (which can be downloaded with the dashboards) to run a side by side comparison and detect where there are differences. It was to make sure there were no major variances when we update the data week to week.

mphprogrammer:
Did you use the `institution_old.csv` and `institution_new.csv` files to do that comparison?

PR #227, opened 2017-03-17
Downcase search name parameter

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1694

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -41,6 +41,7 @@ module V0
       query = params.deep_dup
       query.tap do
         query[:name].try(:strip!)
+        query[:name].try(:downcase!)
         query[:state].try(:upcase!)
         %i(type_name student_veteran_group yellow_ribbon_scholarship principles_of_excellence
            eight_keys_to_veteran_success caution country).each do |k|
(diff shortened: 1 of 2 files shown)


app/controllers/v0/institutions_controller.rb
```diff
@@ -41,6 +41,7 @@ def normalized_query_params
       query = params.deep_dup
       query.tap do
         query[:name].try(:strip!)
+        query[:name].try(:downcase!)
```
saneshark: just making sure name is always a string and not an array of strings or hash.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -32,6 +32,13 @@
       expect(response.content_type).to eq('application/json')
       expect(response).to match_response_schema('institutions')
     end
+
+    it 'search returns case-insensitive results' do
+      get :index, name: 'CHICAGO', version: 1
```
saneshark: Should we maybe improve this to also test 'CHICAGO Illinois' or something like that? Which would return 'University of Illinois Chicago' for example, but not return 'University of Chicago'.

saneshark (approved):

spec/controllers/v0/institutions_controller_spec.rb
patrickvinograd: Added a test case with mixed case and space in the search term. Passes. Merging.

saintsoup52: ED actually publishes a great school alias list that catch situations like this. We use the alias list at CFPB and is already in the raw data ingested by the data service.

PR #229, opened 2017-03-17
Avoid 500 error on missing autocomplete term

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1386

Perhaps more proper to raise a `MissingParameter` Exception, but then would also need to plumb that through as a BackendService exception in vets-api, which seems high-touch for the autocomplete which is by its nature a fire-and-forget kind of API.

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -3,14 +3,17 @@ module V0
   class InstitutionsController < ApiController
     # GET /v0/institutions/autocomplete?term=harv
     def autocomplete
-      @search_term = params[:term].strip.downcase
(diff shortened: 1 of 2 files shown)


patrickvinograd:
Prioritized because even in current GIBCT this was occasionally causing alerts based on a spike in bogus queries.

saneshark (approved):
@patrickvinograd this looks good to me. I wonder if perhaps we would want to show the original params for the self / links though? 

User shouldn't be concerned about what we need to do on the backend to make this happen, and showing the link as @search_term vs the provided params is the equivalent to modifying params.

PR #230, opened 2017-03-20
Fixed lowercase ojt in API. Fixes #226

### LINKED ISSUE #226: facets method in the institution controller relies on lower case 'ojt'
Need to implement a quick fix to change hard coded 'ojt' to upper cased 'OJT'

mphprogrammer:


saneshark (approved):
LGTM

patrickvinograd:
I'm not clear on what the issue is that this is supposed to fix. In fact it seems like it breaks the "employer" count - I pulled down your branch and ran it locally, and the employer count shows as zero when ojt count = 71. On dev-api right now (without this change), the employer count matches the ojt count.

patrickvinograd (dismissed):
Please respond to my question because this seems broken at least when I run it locally.

patrickvinograd:
I guess it would be helpful to know - how is the institution type of "OJT" intended to be represented (uppercase. vs. lowercase) in:
1. the input CSV
2. the database
3. the Institution model
4. the API result for an individual institution
5. the API facets object

patrickvinograd:
Also even if this is the right fix, could presumably reference `Institution::EMPLOYER` instead of re-defining the magic string in the controller.

patrickvinograd:
Ok, I dropped and re-seeded the database on my machine and now I see the types are uniformly in uppercase, so this change makes sense. 

I can re-seed the dev and staging environments - but I need a heads up about when I should do that as its not standard procedure on any given redeployment.

mphprogrammer:
Patrick - Forgot completely about Institution::Employer, and I am that put that in last year ...

mphprogrammer:
@patrickvinograd In all those cases - facility code, institution, country, state, institution type should be in upper case ... It is that way in the DB - as a general policy we should store the data as it is required for reporting purposes, and not for display.

mphprogrammer:
@patrickvinograd replaced the controller constant with Institution::EMPLOYER. Could you give it the once over, and if OK - merge?

PR #231, opened 2017-03-21
Upcase country to match database format

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1805

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -45,9 +45,11 @@ module V0
       query.tap do
         query[:name].try(:strip!)
         query[:name].try(:downcase!)
-        query[:state].try(:upcase!)
+        %i(state country).each do |k|
+          query[k].try(:upcase!)
+        end
         %i(type_name student_veteran_group yellow_ribbon_scholarship principles_of_excellence
-           eight_keys_to_veteran_success caution country).each do |k|
+           eight_keys_to_veteran_success caution).each do |k|
           query[k].try(:downcase!)
         end
       end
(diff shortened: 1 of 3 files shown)


spec/factories/institutions.rb
```diff
@@ -11,42 +11,42 @@
     version 1
 
     trait :in_nyc do
-      city 'new york'
-      state 'ny'
-      country 'usa'
+      city 'NEW YORK'
```
patrickvinograd: Updated factory to make data look like what is in DB.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -56,6 +56,34 @@
...
+    end
+
+    it 'filter by lowercase country returns results' do
+      get :index, name: 'chicago', country: 'usa', version: 1
+      expect(JSON.parse(response.body)['data'].count).to eq(1)
+      expect(response.content_type).to eq('application/json')
+      expect(response).to match_response_schema('institutions')
+    end
+
+    it 'filter by uppercase state returns results' do
```
patrickvinograd: state was actually working ok, now have specs to ensure that's the case.

app/controllers/v0/institutions_controller.rb
```diff
@@ -45,9 +45,11 @@ def normalized_query_params
       query.tap do
         query[:name].try(:strip!)
         query[:name].try(:downcase!)
-        query[:state].try(:upcase!)
+        %i(state country).each do |k|
+          query[k].try(:upcase!)
+        end
         %i(type_name student_veteran_group yellow_ribbon_scholarship principles_of_excellence
```
patrickvinograd: type_name is handled specially "school/employer" mapped to "not equal/equal OJT". All other filter fields here are booleans.

mphprogrammer (approved):
LGTM

saneshark:
This is good for now, but i think we need investigate doing case insensitive collation on these table columns in a future iteration.

There is also Postgres `ILIKE`: more on that here: https://www.postgresql.org/docs/8.3/static/functions-matching.html

PR #234, opened 2017-03-22
Update country facet to avoid key munging between API and frontend

### LINKED ISSUE #224: Filter value requiring hyphen in requests is inconsistent with its facet key in responses
With the `/v0/gi/institutions/search` resource, we get a list of countries in the `meta.facets.country` object, and `bosnia_herzegovina` is one of them. However, trying to filter with it doesn’t return any results: `/v0/institutions?country=BOSNIA_HERZEGOVINA&version=1`.

Furthermore, we convert to camelCase on the front end, so it actually removes the underscore and would get us `bosniaHerzegovina`. It seems that the filter would only work with `country=bosnia-herzegovina` (with the hyphen).

So far, this is the only example I could find, but there could potentially be more.

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/gibct-data-service/issues/224.
though needs a corresponding fix on frontend to consume this structure.

Example facet object with this fix. Only country is embedded in this way since it's the only facet with particularly open-ended values. 
```
"facets": {
  "type":{"school":136,"employer":108},
  "type_name":{"flight":3,"for profit":32,"foreign":5,"ojt":108,"private":55,"public":41},
  "state":{"al":3,"ar":6,"az":4,"ca":10,"co":21,"ct":1,"fl":5,"ga":12,"ia":3,"id":8,"il":4,"in":1,"ky":2,"ma":5,"md":4,"me":4,"mi":2,"mn":4,"mo":4,"ms":2,"mt":7,"nc":25,"nd":2,"nh":4,"nj":17,"nm":1,"ny":13,"oh":9,"ok":1,"or":3,"pa":6,"sd":1,"tn":2,"tx":1,"ut":13,"va":5,"vt":7,"wa":7,"wi":2,"wv":5,"wy":3},
  "country":[{"name":"AUSTRALIA","count":1},{"name":"CANADA","count":4},{"name":"USA","count":239}],
  "caution_flag":{"false":234,"true":10},
  "student_vet_group":{"false":229,"true":15},
  "yellow_ribbon_scholarship":{"false":202,"true":42},
  "principles_of_excellence":{"false":190,"true":54},
  "eight_keys_to_veteran_success":{"false":222,"true":22}}
```

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -79,7 +79,7 @@ module V0
         },
         type_name: institution_types,
         state: search_results.filter_count(:state),
-        country: search_results.filter_count(:country),
+        country: embed(search_results.filter_count(:country)),
         caution_flag: search_results.filter_count(:caution_flag),
         student_vet_group: search_results.filter_count(:student_veteran),
         yellow_ribbon_scholarship: search_results.filter_count(:yr),
@@ -88,5 +88,14 @@ module V0
       }
     end
     # rubocop:enable AbcSize
+
+    # Embed search result counts as a list of hashes with "name"/"count"
+    # keys so that open-ended strings such as country names do not
+    # get interpreted/mutated as JSON keys.
+    def embed(group_counts)
+      group_counts.each_with_object([]) do |(k, v), array|
+        array << { name: k, count: v }
+      end
+    end
   end
 end
--- a/spec/controllers/v0/institutions_controller_spec.rb
+++ b/spec/controllers/v0/institutions_controller_spec.rb
@@ -84,6 +84,14 @@ RSpec.describe V0::InstitutionsController, type: :controller do
       expect(response.content_type).to eq('application/json')
       expect(response).to match_response_schema('institutions')
     end
+
+    it 'has facet metadata' do
+      get :index, name: 'chicago', version: 1
+      facets = JSON.parse(response.body)['meta']['facets']
+      expect(facets['state']['il']).to eq(1)
+      expect(facets['country'].count).to eq(1)
+      expect(facets['country'][0]['name']).to eq('USA')
+    end
   end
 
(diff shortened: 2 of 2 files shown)


U-DON (approved):
LGTM.

app/controllers/v0/institutions_controller.rb
```diff
@@ -88,5 +88,14 @@ def facets
...
     end
     # rubocop:enable AbcSize
+
+    # Embed search result counts as a list of hashes with "name"/"count"
+    # keys so that open-ended strings such as country names do not
+    # get interpreted/mutated as JSON keys.
+    def embed(group_counts)
+      group_counts.each_with_object([]) do |(k, v), array|
+        array << { name: k, count: v }
+      end
```
saneshark: I think we can do this in the database with group style query and it will be faster, but we can revisit that later.

saneshark:
LGTM other than nitpick comment (we can revisit as part of query optimizations we make in future iterations).

saneshark (approved):

app/controllers/v0/institutions_controller.rb
patrickvinograd: This is already done with a database group/count query by means of the filter_counts method in the facets method above. This new method is just re-structuring the result of that query.

patrickvinograd:
Holding off on merging this until frontend fix is in place.

app/controllers/v0/institutions_controller.rb
saneshark: In that case it would be a lot more readable if we used the active record methods than these custom query methods in my opinion. I'll have to demonstrate with an example later, but if active record is used to do the group directly, you can actually build the hash you are building in the embed method here automatically (no need for the additional iterating over each object).

U-DON:
The fix on the front end has been merged.

PR #235, opened 2017-03-22
Refactor overloaded type_name query parameter to type/category

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1814
requires corresponding front-end fix to use updated facets and query parameters. 

New parameters are:
`category` = school/employer
`type` = private/for profit/ojt/foreign/etc

That's both what should be sent as query parameters, and what will come back in the facets object for populating the search form.

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -221,17 +221,22 @@ class Institution < ActiveRecord::Base
   scope :filter, lambda { |field, value|
     return if value.blank?
     raise ArgumentError, 'Field name is required' if field.blank?
-    case value
-    when 'true', 'yes'
-      where(field => true)
-    when 'false', 'no'
-      where.not(field => true)
-    when 'school' # field: institution_type_name
-      where.not(field => EMPLOYER)
-    when 'employer' # field: institution_type_name
-      where(field => EMPLOYER)
+    if field == :category
+      case value
+      when 'school'
+        where.not(institution_type_name: EMPLOYER)
+      when 'employer'
+        where(institution_type_name: EMPLOYER)
+      end
     else
-      where(field => value)
+      case value
+      when 'true', 'yes'
+        where(field => true)
+      when 'false', 'no'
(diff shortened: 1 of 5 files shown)


app/models/institution.rb
```diff
@@ -221,17 +221,22 @@ def self.autocomplete(search_term, limit = 6)
...
-    case value
-    when 'true', 'yes'
-      where(field => true)
-    when 'false', 'no'
-      where.not(field => true)
-    when 'school' # field: institution_type_name
-      where.not(field => EMPLOYER)
-    when 'employer' # field: institution_type_name
-      where(field => EMPLOYER)
+    if field == :category
```
patrickvinograd: category is the only real "synthetic" query parameter that doesn't map straight through to a database value, so for clarity's sake I pulled it out of the single case statement that was here before.

U-DON (approved):
LGTM.

saneshark:
So correct me if I'm wrong but with category having its own if / else this means that it will take precedence over the other attributes. To me it seems more logical to have the filters update / refresh on the front-end in a dynamic manner, this way both have equal precedence vs one having greater precedence than the other. Not only would this simplify some of the logic on the backend for institution_type_name column, but it also makes for a more dynamic set of filters for veterans, does it not?

patrickvinograd:
No this does not affect precedence. The :filter scope is just turning query params into the where clauses of the SQL statement, all of which are ANDed together.

The old code was taking a shortcut by saying if `value == school`, generate `where institution_type_name != OJT`, without actually checking whether the query parameter in question was `type_name`. So in practice you could submit an API request with `?state=school` and it would generate sql like `where state != OJT`. Not particularly dangerous, but incorrect. The version in this PR makes the behavior more explicit by only treating "school" and "employer" as meaningful query parameter values when the query parameter = "category".

saneshark (approved):

PR #236, opened 2017-03-30
Removed false/0 defaults, fixes #232

### LINKED ISSUE #232: "False"s & Zeros should be Blanks in export csv
Was just reviewing the new export csv and y'all are doing an amazing job.

I did notice one small thing... the new export file has lots of "False" and "0" where they used to be blanks. Since we use this export csv to run the public facing data dashboards, we need those fields to be appear as blanks in the csv.

The following fields are showing "False" when it should just be blank:
* Yr
* Student_veteran
* Poe
* Eight_keys
* Dodmou
* Accredited
* Caution_flag

The following fields are showing "0" when it should just be blank:
* Gibill
* P911_tuition_fees	
* P911_recipients	
* P911_yellow_ribbon	
* P911_yr_recipients
* All the complaint field

Last note... in the export csv all the headers are lowercase... not sure why the headers in this file all have the first letter capitalized.

Old (top) vs new file (bottom)
image

mphprogrammer:
Removed institution defaults to correct exporting/displaying false and 0 when no data was supplied.


db/migrate/20170330161442_change_institution_defaults.rb
```diff
@@ -0,0 +1,48 @@
+class ChangeInstitutionDefaults < ActiveRecord::Migration
+  def change
+    change_table :institutions do |t|
+      t.change :flight, :boolean, default: nil
+      t.change :correspondence, :boolean, default: nil
+      t.change :gibill, :integer, default: nil
+      t.change :yr, :boolean, default: nil
```
patrickvinograd: What effect does removing the default for :yr, :student_veteran, and other fields have on the `facets` object returned by institution search API? Do we now wind up with counts for true, false, and null instead of just true and false?

patrickvinograd (commented):
Question about effect on facets.

db/migrate/20170330161442_change_institution_defaults.rb
mphprogrammer: The relevant search filter code could accept `true`, `false`, or `nil`. However, the functionality of the GIBCT should be only to search (using the checkboxes) for `true` values. There should be no mechanism to search for a boolean that is `nil` or `false`. In other words, filtering should only be `true` or not `true` - this is how the current tool works.

mphprogrammer: For some boolean values that we search on, a `true` value is determined simply by the inclusion of the school in a data file. There is no guarantee that if it is not in the list, the boolean is false. The best we can say is that it isn't true.

mphprogrammer: This presents an issue for the not `caution_flag` search. I didn't realize that was added to the functionality, did someone request that? In this case there would be an issue.

mphprogrammer: `caution_flag` should not be filterable, for a number of business-rule reasons.

mphprogrammer: That's a great question

patrickvinograd:
So I think from the facets perspective we are okay - I re-seeded the sample CSVs, and in the institutions table all the filterable fields are either "t/f" or "t/nil" - there aren't any that wind up as "t/f/nil". And hence no facets that have three values. 

There is a slight discrepancy because a field like poe that has "t/f" in the database winds up with a facet like: 
`"principles_of_excellence":{"false":29207,"true":5414}`
Whereas a field like student_vet_group that has "t/nil" in the database winds up with a facet like: `"student_vet_group":{"true":1858}` (missing the count for false). But that has no impact on the frontend since it only consumes the "true" counts anyway. So I think we are good to merge this.

patrickvinograd (approved):
LGTM.

PR #238, opened 2017-03-31
Downcase instead of capitalize exported CSV headers

### LINKED ISSUE #237: Exported CSV column headers should be lowercase
Mentioned in #232. The PR that fixes the main issue in #232 doesn't address the export column case, so filing a separate issue to track that so it doesn't get lost.

patrickvinograd:
Fixes #237. 

Note this affects export of other CSVs, not just the main data.csv. But other CSVs that we have examples of are all over the map in terms of lowercase/uppercase/title case, so if we're going to have a one-size-fits-all here it should be the right one for the most important exported CSV.


mphprogrammer (approved):
LGTM

PR #239, opened 2017-04-04
Caution flag reason text is displayed as input from CSV, along with other fields Fixes #5125

mphprogrammer:
Also removed ability to filter caution flags. Bad ju-ju.


spec/models/converters/base_converter_spec.rb
```diff
@@ -14,15 +14,16 @@
     end
   end
 
+  it 'does not convert strings containing more than forbidden words to nil' do
```
patrickvinograd: The spec description looks like it missing a word..."containing more than ??? forbidden words"

spec/models/converters/base_converter_spec.rb
```diff
@@ -14,15 +14,16 @@
...
+  end
+
   it 'converts nil to nil' do
     expect(subject.convert(nil)).to be_nil
   end
 
   it 'converts blanks to an empty string' do
     expect(subject.convert('      ')).to be_blank
   end
-
```
patrickvinograd: Instead of removing this test case, why not flip its logic to be "it preserves input case" and expect input to match output. That way if behavior changes later it is caught by the spec.

app/models/institution_builder.rb
```diff
@@ -383,7 +383,7 @@ def self.add_hcm(version_number)
     str = <<-SQL
       UPDATE institutions SET
         caution_flag = TRUE,
-        caution_flag_reason = concat_ws(',', caution_flag_reason, hcm_list.reasons)
+        caution_flag_reason = concat_ws(', ', caution_flag_reason, hcm_list.reasons)
```
patrickvinograd: Would be nice to add a spec covering this concatenation to make sure it looks right.

patrickvinograd (changes requested):
Just a few minor spec issues. See comments.

app/controllers/v0/institutions_controller.rb
```diff
@@ -60,7 +60,6 @@ def search_results
       @query ||= normalized_query_params
```
saneshark: @mphprogrammer Do the normalized query params list caution_flag as an available query param? I think we should also maybe remove it from that array constant, if applicable.

app/models/complaint.rb
```diff
@@ -75,7 +75,7 @@ class Complaint < ActiveRecord::Base
     'school state' => { column: :state, converter: BaseConverter },
     'submitted' => { column: :submitted, converter: BaseConverter },
     'closed' => { column: :closed, converter: BaseConverter },
-    'closed reason' => { column: :closed_reason, converter: BaseConverter },
+    'closed reason' => { column: :closed_reason, converter: DowncaseConverter },
```
saneshark: Will these changes affect the CSV export vs what we show on the Front-end?

app/models/converters/boolean_converter.rb
```diff
@@ -1,10 +1,10 @@
...
 
 class BooleanConverter < BaseConverter
-  TRUTHS = %w(true t yes ye y 1 on).freeze
-
   def self.convert(value)
     value = super(value)
-    value.blank? ? nil : TRUTHS.include?(value.to_s)
+
+    return nil if value.blank?
+    value.to_s.match(/\A(true|t|yes|ye|y|1|on)\z/i).present? ? true : false
```
saneshark: This is pretty awesome, but it might be something we would want to raise warnings around in the CSV message ultimately, maybe we can put a TODO comment or something in there. 

My ultimate goal would be to have the CSV that gets uploaded to be consistent. If there are this many freaking nuances to something characterized as true, the method whereby they are input is likely wrought with errors and the admins need to take a closer look at precisely what it is they are uploading.

Obviously this is just a nitpick for now, something to think about long-term.

app/models/hcm.rb
```diff
@@ -10,7 +10,7 @@ class Hcm < ActiveRecord::Base
     'country' => { column: :country, converter: BaseConverter },
     'institution type' => { column: :institution_type, converter: BaseConverter },
     'stop pay/monitor method' => { column: :hcm_type, converter: BaseConverter },
-    'method reason desc' => { column: :hcm_reason, converter: DisplayConverter }
+    'method reason desc' => { column: :hcm_reason, converter: BaseConverter }
```
saneshark: @mphprogrammer can you elaborate on why this change is necessary?

app/models/settlement.rb
```diff
@@ -7,7 +7,7 @@ class Settlement < ActiveRecord::Base
     'instnm' => { column: :institution, converter: InstitutionConverter },
     'school_system_code' => { column: :school_system_code, converter: NumberConverter },
     'school_system_name' => { column: :school_system_name, converter: BaseConverter },
-    'settlement_description' => { column: :settlement_description, converter: DisplayConverter },
+    'settlement_description' => { column: :settlement_description, converter: BaseConverter },
```
saneshark: Curious what this change helps facilitate.

saneshark (changes requested):

app/models/complaint.rb
mphprogrammer: No, it shouldn't. I think that the exported data should represent exactly what is in the database at the time of export, with the same formatting and casing.

app/models/converters/boolean_converter.rb
mphprogrammer: I agree. We need to evangelize the need for consistent norms with data types in these CSVs. It would speed up processing.

app/models/hcm.rb
mphprogrammer: Yes - Patrick's issue was with the case of letters in acroynyms being an issue with the Dept of Ed. The BaseConverter leaves the case intact, but does strip quote characters, `null`, `none` and `privacysuppressed`, as well leading and trailing blanks. We are now displaying caution flag reasons exactly as input.

app/models/settlement.rb
mphprogrammer: Patrick's issue was with the case of letters in acroynyms being an issue. The BaseConverter leaves the case intact, but does strip quote characters, null, none and privacysuppressed, as well leading and trailing blanks. We are now displaying caution flag reasons exactly as input.

spec/models/converters/base_converter_spec.rb
mphprogrammer: Got it.

spec/models/converters/base_converter_spec.rb
mphprogrammer: Got it.

app/controllers/v0/institutions_controller.rb
mphprogrammer: Missed that! Its removed now.

patrickvinograd (approved):

patrickvinograd:
My requested spec changes are accounted for, will let Kam take another pass.

app/models/institution_builder.rb
mphprogrammer: There is a spec in place for this that adds an accreditation caution first, and then an HCM. It uses a regex to match, which is why I missed the ',' rather than ', '. However, I did add a concatenation test for Accreditation - although it affects the caution flag reason first, if moved to another position in the build, the caution flag reason will still be concatenated properly.

saneshark (approved):

saneshark:
@mphprogrammer LGTM

Issue #213, opened 2017-03-06
Change routes for non-idempotent actions in dashboard


mphprogrammer:
Currently, non-idempotent actions are mapped to get requests. This is not per HTTP specification. Change non-idempotent routes (build, push) in dash from GET to POST or PUT as appropriate.

patrickvinograd:
I agree this can probably be deferred from the launch-blocking list, however it's not just an academic risk - there are various web accelerator plugins (not as popular as they used to) and even chrome itself has a prefetch function that pre-follows links as you browse. It will follow GETs but not PUTs/POSTs so could trigger an action in GIDS.

leannammiller:
@patrickvinograd @mphprogrammer - sounds like this is something that should happen this sprint if we're trying to stop work on GIBCT after this sprint. Is this correct?

patrickvinograd:
I think there's inevitably going to be some ongoing maintenance work/technical debt for GIBCT even if it's not our focus, but I have some cycles today and this is a well-contained change - impacts GIDS admin interface only - so I will get it squared away.

PR #240, opened 2017-04-07
Change build/push methods to POST

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/gibct-data-service/issues/213

Open to differing opinions about HTTP method - POST vs. PUT. These operations are not very resource-oriented so it seemed like a toss-up.

### DIFF
--- a/app/views/dashboards/_introduction.html.erb
+++ b/app/views/dashboards/_introduction.html.erb
@@ -31,7 +31,7 @@
       <div class="row row-space-6">
         <div class="col-xs-12">
           <% if uploads.length == InstitutionBuilder::TABLES.length %>
-            <%= link_to 'Generate Preview', dashboard_build_path, data: { confirm: 'Are you sure?' },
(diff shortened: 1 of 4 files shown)


patrickvinograd:
Interestingly this did not cause any spec failures, even though the dashboards_controller spec does a `get :build` and a `get :push`. But maybe we'd need a request spec to exercise this distinction?

patrickvinograd:
Yes, looks like only request specs exercise routing. Will add one after next block of meetings.

saneshark:
@patrickvinograd I'm +1 for either request specs or routing specs. Both should help surface these type of issues in the future.

saneshark:
Since we already have the controller specs, maybe routing specs would suffice, but replacing controller specs with request specs (which encapsulate controller and routing better), would probably be the better long term solution (though a little annoying to do when dealing with file upload).

saneshark (approved):
Feel free to open a separate issue to track improving specs if you wish. This LGTM though.

PR #242, opened 2017-04-10
Pv add inst indexes

patrickvinograd:
Avoids full table scan of ever-growing institutions table, primarily via addition of version index.
Reduces time for "generate preview" from 5 minutes down to 30-40 seconds. 

Pulled this branch into staging and applied the db migration there for verification.


saneshark (approved):
LGTM

PR #243, opened 2017-04-10
Pv add search indexes

patrickvinograd:
Layers on top of #242. This is the low-hanging fruit of https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1306 - leaves aside full-text search of institution name/city for now. 

* Upcases the facility_code query parameter which is already upcased in the database. 
* Uses a functional prefix index of the appropriate case for autocomplete. Increases autocomplete
latency quite a bit. Will do a more formal load test sometime in the future, but casual testing shows autocomplete query latency drops from anywhere in the range of 20-100 ms to < 2 ms. 

One downside of the functional index is that it is not reflected in schema.rb because it is postgres-specific. So there will be a bit of work in devops to make sure that index is created for the one-time database re-setup.

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -206,12 +206,12 @@ class Institution < ActiveRecord::Base
   scope :search, lambda { |search_term|
     return if search_term.blank?
     clause = [
-      'lower(facility_code) = (?)',
+      'facility_code = (?)',
       'lower(institution) LIKE (?)',
       'lower(city) LIKE (?)'
     ].join(' OR ')
     terms = [
-      search_term,
+      search_term.upcase,
       "%#{search_term}%",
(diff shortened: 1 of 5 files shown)


saneshark:
@patrickvinograd I tried to fix the merge conflict on this but I think I set the date wrong. Should have just let you merge it. LGTM

I think you will need to pull the changes I made (resolving conflicts), and fix the date in the schema.rb so that the build will pass, I probably merged in the wrong order.

patrickvinograd:
It's okay, it was kind of screwy how I had layered these PRs. Fix inbound.

patrickvinograd:
Can you give it a formal 👍 so github is happy?

saneshark (approved):

PR #244, opened 2017-04-10
adding uuid to versions and helpful rake task

saneshark:
- Added uuid to versions table
- Added pre-populate method as before_create action on Version.rb
- Migration uses pre-populate method to set the default uuid before setting not null constraint.
- Added index on uuid, with not null and uniqueness constraints
- Added a helpful rake task for dumping the database and restoring from backup (handy for not having to reseed all the damn time for testing purposes)

### DIFF
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'securerandom'
+
 class Version < ActiveRecord::Base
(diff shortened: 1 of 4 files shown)


patrickvinograd (approved):
LGTM. My pending PRs #242 and #243 have migrations that numerically come before this so let's sync up on what order to merge, assuming all of these are good to go.

PR #245, opened 2017-04-11
Error handling - IN PROGRESS

mphprogrammer:
Handles missing/extra headers when uploading and provides blanket coverage for everything else.


mphprogrammer:
@patrickvinograd Checked the build log and everything seemed to be OK - what am I missing?

patrickvinograd:
Hard to pick out but right at the bottom there's a file with low coverage:
```
File (/home/jenkins/workspace/bct-data-service_export-bad-PF3C46IFPI3KX4VLWLD4PIGER5SCRBIEAGDBGPHGAS24YCRO6NSQ/app/controllers/concerns/flashable.rb) is only (76.92%) covered. This is below the expected minimum coverage per file of (90.00%).
```

patrickvinograd:
You're now falling through to the rescue block before calling `Flashable#errors_for_alert` - so maybe no longer showing the excerpt of errors from the uploaded file?

app/controllers/uploads_controller.rb
```diff
@@ -9,48 +9,61 @@ def index
...
-    else
-      render :new, alert: errors_for_alert([@upload]), notice: "Failed to upload #{original_filename}."
+      render :new
     end
   end
 
   def show
     @upload = Upload.find_by(id: params[:id])
-    redirect_to uploads_path, alert: ["Upload with id: '#{params[:id]}' not found"] unless @upload
+    return if @upload.present?
```
saneshark: you want to return without rendering?

app/controllers/uploads_controller.rb
```diff
@@ -9,48 +9,61 @@ def index
   end
 
   def new
-    csv_type = params[:csv_type]
-
-    @upload = Upload.new(csv_type: csv_type)
-    @upload.skip_lines = defaults(csv_type)['skip_lines']
-
+    @upload = new_upload(params[:csv_type])
     return if @upload.csv_type.blank? || @upload.csv_type_check?
```
saneshark: I'm confused on this, if you return without rendering how will a user know what happened? You should at least render a flash message or something.  Are you sure you don't mean 'render new and return' ?

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170410221932) do
+ActiveRecord::Schema.define(version: 20170410215512) do
```
saneshark: not sure why this is necessary if you didn't have any migrations

saneshark (commented):
@mphprogrammer can you address these comments and i'll approve?

app/controllers/uploads_controller.rb
mphprogrammer: If the upload fails it redirects to the index action, if it succeeds the show view is presented by default.

db/schema.rb
mphprogrammer: There were, although I don't know why - I was presented with a migrations pending error...

mphprogrammer: I had to fast_forward on the same branch (no changes though!) and then migrate...

app/controllers/uploads_controller.rb
mphprogrammer: In light of your changes (removing the generic upload - I should have taken this out.

saneshark (approved):
I'm satisfied with the input you provided to my questions. If all of @patrickvinograd comments have been addressed too then I think it looks good to merge.

LGTM

mphprogrammer:
@patrickvinograd could you take a peek as well?

patrickvinograd (approved):
LGTM.

PR #248, opened 2017-04-11
Version meta flag for preview had wrong logic

saneshark:
- .preview? has some incorrect logic and needs to be looked at more closely. 
- for now I'm using !production? instead.


patrickvinograd (approved):
LGTM.

PR #250, opened 2017-04-12
Pv add sentry

patrickvinograd:
Sentry DSN endpoint is already defined/configured in devops repo - this should be sufficient to start injecting exceptions into Sentry.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,7 +54,7 @@ gem "puma", "~> 3.6"
 gem 'figaro'
 
 # Sentry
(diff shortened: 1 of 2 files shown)


patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2053

saneshark:
@patrickvinograd sounds good -- we should talk through having a nice way to delineate API request errors from admin interface errors.

saneshark (approved):

PR #259, opened 2017-04-12
Remove non-type-specific upload form button

### LINKED ISSUE #253: Clicking on green "Upload a file" button produces error button
Starting from the "uploads" page that lists the most current uploads to the DS, I clicked on the green "Upload a File" button 

image

and I was taken to the dashboard page (also another button) and was given the following error message.

image

So two things... 
1) I did not try to upload a file so this is an odd error message and 
2) there really is no need for this button now that both just go back to the dashboard page. 

@mphprogrammer @saneshark

patrickvinograd:
Fixes #253 
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2112

This logic got dropped during a recent error handling improvement.

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -10,7 +10,7 @@ class UploadsController < ApplicationController
 
   def new
     @upload = new_upload(params[:csv_type])
-    return if @upload.csv_type_check?
+    return if @upload.csv_type.blank? || @upload.csv_type_check?
 
     log_and_display_error(errors_for_alert([@upload]), 'Warning')
     redirect_to dashboards_path
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -10,7 +10,7 @@ class UploadsController < ApplicationController
 
   def new
     @upload = new_upload(params[:csv_type])
-    return if @upload.csv_type.blank? || @upload.csv_type_check?
+    return if @upload.csv_type_check?
 
     log_and_display_error(errors_for_alert([@upload]), 'Warning')
     redirect_to dashboards_path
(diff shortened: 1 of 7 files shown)


mphprogrammer:
The "upload" button should be removed ... since this function has been removed from the dashboard, I had changed the logic to disallow blank? csv_type

saneshark:
@mphprogrammer I think this check is still wise. But we should still remove the button from upload history page and default the type automatically from the dashboard using one of the 21 options there.

app/controllers/uploads_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
 
   def new
     @upload = new_upload(params[:csv_type])
-    return if @upload.csv_type_check?
+    return if @upload.csv_type.blank? || @upload.csv_type_check?
```
saneshark: shouldn't we have a flash message that describes the problem?

saneshark (changes requested):

app/controllers/uploads_controller.rb
mphprogrammer: The original logic should be left in place if we agree to remove the "Upload File" button from this view, since the generic upload function was removed from the dashboard".

patrickvinograd:
@saneshark Ready for re-review.

saneshark:
What if someone modifies the form input for CSV-type and passes nil? I think the check you had in there was OK @patrickvinograd 

Even if we remove the generic form, we should be doing an upfront check to make sure the csv type passed is valid. It's not the end of the world, but it was a nice check to have. I'm going to go ahead and approve this anyway since it's unlikely that someone will manually modify the form params, but just mentioning it here.

saneshark (approved):
LGTM, but the inclusion of the csv_type check you had was still useful IMO (just needed a handy flash error to notify what happened).

patrickvinograd:
There is a flash alert if you can somehow navigate to the form with an invalid csv_type - that's what the obsolete button was doing. I think we're good here. If one of our 5 users is modifying form inputs, we tell them to knock it off.

Issue #251, opened 2017-04-12
Exported files have no file type


saintsoup52:
I tried exporting a number of the CSV files and the big institution file and they were an "unknown file type".

![image](https://cloud.githubusercontent.com/assets/4397508/24955897/4ba8f768-1f53-11e7-8b9b-19efb7ce2c79.png)

![image](https://cloud.githubusercontent.com/assets/4397508/24955883/3c68a32a-1f53-11e7-9829-f5f469747d20.png)

I was able to eventually open them in notepad but that is a cumbersome process to get them back into an excel format.

@mphprogrammer @saneshark

patrickvinograd:
Should add `text/csv` Content-Type header during download.

mphprogrammer:
Yup, missed that. A quick workaround would be to rename the file with a .csv extension.

saneshark:
I think Rails will be smart enough to negotiate the content-type based on extension, but doesn't hurt to do both.

patrickvinograd:
When I download this using Chrome on a mac, it is serving a content-type of `text/csv`, and Chrome puts it in my Downloads folder with a reasonable filename ("Institution.csv", "Hcm.csv"). And if I open it from the Finder it opens in Excel. The latter is dependent on the desktop setup of the user, I don't have a windows machine to test on. 

The only server-side thing I see missing is that we are serving a Content-Disposition header but it doesn't have the `filename` parameter in that header, which may account for the missing extension on Windows. Maybe chrome derives it from the Content-Type but IE or whatever browser was used above does not?

PR #261, opened 2017-04-12
Adds filename parameter to content-disposition header

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/gibct-data-service/issues/251

Will need confirmation that this better addresses the issue on Windows where it was observed, but it's certainly not any worse.


mphprogrammer (approved):
LGTM

PR #264, opened 2017-04-12
version refactor phase 1

saneshark:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/1990

- removed a bunch of unnecessary methods.
- added a bunch of helpful methods
- ensured scopes always returning relation object

I'm calling this a phase 1, because I would still like to improve this more in the future, but seeing as we don't currently support rollback and other things that would increase complexity on this, i'm comfortable with just these changes.

### DIFF
--- a/spec/controllers/dashboards_controller_spec.rb
+++ b/spec/controllers/dashboards_controller_spec.rb
@@ -109,7 +109,7 @@ RSpec.describe DashboardsController, type: :controller do
         get :push
 
         expect(flash.alert).to eq('No preview version available')
-        expect(Version.production_version).to be_blank
+        expect(Version.current_production).to be_blank
       end
     end
 
@@ -125,7 +125,7 @@ RSpec.describe DashboardsController, type: :controller do
 
         it 'sets the new production version number to the preview number' do
           get :push
-          expect(Version.production_version.number).to eq(Version.preview_version.number)
+          expect(Version.current_production.number).to eq(Version.current_preview.number)
         end
       end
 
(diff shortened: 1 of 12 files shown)


mphprogrammer (approved):
LGTM - but we may want to change the partial that displays the preview record actions to use publish? rather than passing in a local. Thats in the other PR so lets push this one out, and then change the other pr (preview == production) to disable the publish button.

saneshark:
Good point, i might as well change the query for fetching previews and production to use the updated scopes now that `newest` returns a relation object.

saneshark:
@patrickvinograd are you good with me merging this one?

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -109,7 +109,7 @@ def load_table(klass, options)
         get :push
 
         expect(flash.alert).to eq('No preview version available')
-        expect(Version.production_version).to be_blank
+        expect(Version.current_production).to be_blank
```
patrickvinograd: Should this be `current_preview`?

patrickvinograd (approved):
LGTM. One question on the dashboard spec, but I didn't read through the spec thoroughly, just noticed something potentially out of place.

patrickvinograd:
I think we should leave it here in terms of refactoring, unless there's an actual bug fix needed.

spec/controllers/dashboards_controller_spec.rb
saneshark: the spec is testing to make sure that push returns a flash error and that nothing is pushed to production. The last expectation in the example is kind of dumb, but I guess it doesn't hurt to have it.

saneshark:
Yes any future refactor will take place if and when we support rollback or showing more than one production / preview version in the dashboard.

PR #266, opened 2017-04-13
Include active search terms in facets

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2096 
by ensuring that any active search terms are included in the facets result so that they will be present in the drop down menus and can be un-selected. 

Not wild about having to disable those rubocops, but it's an awkward method by its nature because we had to tweak the facet structure for countries. And I don't know why it wanted me to use a guard instead of an if clause for only the country clause but not the first two...

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -70,7 +70,7 @@ module V0
 
     def facets
       institution_types = search_results.filter_count(:institution_type_name)
-      {
+      result = {
         category: {
           school: institution_types.except(Institution::EMPLOYER).inject(0) { |count, (_t, n)| count + n },
           employer: institution_types[Institution::EMPLOYER].to_i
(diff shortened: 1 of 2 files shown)


patrickvinograd:
I manually pushed this to dev and did a quick test and it avoids the getting-stuck issue that is reported in the issue, no frontend changes required.

app/controllers/v0/institutions_controller.rb
```diff
@@ -84,8 +84,28 @@ def facets
...
+        raw_facets[:type][key] = 0 unless raw_facets[:type].key? key
+      end
+      if @query[:country].present?
+        key = @query[:country].upcase
+        raw_facets[:country] << { name: key, count: 0 } unless
+          raw_facets[:country].any? { |c| c[:name] == key }
+      end
+      raw_facets
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity, Style/GuardClause
```
saneshark: @patrickvinograd this looks fine to me -- don't have time to worry about stupid rubocop right now, can revisit later if anything. Plenty readable. But I am concerned about one thing. Aren't the facets used by the front-end? By defaulting null values rather than having the front-end handle them, I think we might be misleading.

saneshark (approved):
Other than the one comment I'm going to say LGTM. @mphprogrammer can you take a look at this one too.

app/controllers/v0/institutions_controller.rb
patrickvinograd: Yeah, although I think there's an argument to be made that a query with a search parameter should include that parameter in the facets - not just because it happens to make our frontend work but because it's making the output metadata reflect the input parameters of the request.

PR #268, opened 2017-04-13
Dont generate preview if no uploads

### LINKED ISSUE #254: What does "generate new preview version" do?
In the old GIDS uploaded files were not merged into the bigger institution until a user told the DS to merge all the files together. The purpose, as I understood it, was so that if a file was uploaded that had errors there would be a chance to do it again before merging it all together.

However, now the "generate new preview version" is now in a different place and has a user warning that says, "Are you sure?"

image

image

At what point are CSVs merged into the larger institution file? And if it is when you select this "generate new preview version" button shouldn't it be up with the other three action buttons?

image

@mphprogrammer @saneshark

saneshark:
- optimized query for latest_uploads
- added helper method on version to determine if buildable?
- updated views to Fix #254
- added a check to build action to ensure buildable or to raise flash error.

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 class DashboardsController < ApplicationController
   include Flashable
+  before_action :verify_buildable, only: :build
 
   def index
     @uploads = Upload.last_uploads
@@ -53,6 +54,12 @@ class DashboardsController < ApplicationController
   end
 
   private
+
+  def verify_buildable
+    return true if Version.buildable?
+    flash.alert = 'Cannot build a new version since no new uploads have been made.'
+    redirect_to dashboards_path and return
+  end
 
   def csv_model(csv_type)
     return Institution if csv_type == 'Institution'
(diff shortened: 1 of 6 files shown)


saneshark:
This probably needs specs!

app/models/version.rb
```diff
@@ -26,6 +26,20 @@ def self.current_preview
...
+  def self.buildable?
+    Version.buildable_state.to_s.match('can_create')
+  end
+
+  def self.buildable_state
+    upload_dates = Upload.last_uploads.to_a.map(&:updated_at)
+    preview = Version.current_preview
+    return :not_enough_uploads if upload_dates.length < InstitutionBuilder::TABLES.length
+    return :too_many_uploads if upload_dates.length > InstitutionBuilder::TABLES.length
+    return :can_create_first_preview if preview.nil?
```
patrickvinograd: Good catch!

app/controllers/dashboards_controller.rb
```diff
@@ -54,6 +55,12 @@ def push
 
   private
 
+  def verify_buildable
+    return true if Version.buildable?
+    flash.alert = 'Cannot build a new version since no new uploads have been made.'
```
patrickvinograd: I would change this to "Cannot build a new version since no new CSVs have been uploaded."

patrickvinograd (approved):
LGTM, with one optional alert phrasing suggestion.

mphprogrammer (approved):
LGTM - like the query

PR #270, opened 2017-04-14
Headers warnings alerts fixes #256

### LINKED ISSUE #256: Error uploading files missing non-critical columns or too many columns
I tried uploading the directory file from IPEDS, AKA "Ipeds Hd" and I got the following error message.

"Failed to upload HD2015 20170412.csv.

IpedsHd has an issue with headers. Missing headers: ccipug, ccipgrad, ccugprof, ccenrprf, ccsizset. Extra headers: c15basic, c15ipug, c15ipgrd, c15ugprf, c15enprf, c15szset."

None of the missing columns are used in the DS and therefore I thought they should not be causing a critical error when they are missing.

Additionally, we had talked about not causing errors when files might have additional columns as well.

Many times non VA files like IPEDS, Accreditation, 8-keys, HCM, etc. will get rid of data points have new data points that we do not use. Requiring updates to the data service each time this happens will cause a significant workload.

image

@mphprogrammer @saneshark

mphprogrammer:


mphprogrammer:
Ignore previous branch from which this was derived ...

PR #271, opened 2017-04-14
Headers warnings alerts fixes #256

### LINKED ISSUE #256: Error uploading files missing non-critical columns or too many columns

mphprogrammer:
* Eliminates flashable
* Moves diffed_headers to Upload model


patrickvinograd:
Looking at this now.

app/controllers/uploads_controller.rb
```diff
@@ -12,22 +10,29 @@ def new
...
     redirect_to dashboards_path
   end
 
   def create
     @upload = Upload.create(merged_params)
 
     begin
-      data = load_csv
-      display_failed_instances(data.failed_instances)
+      failed = load_csv.failed_instances
```
patrickvinograd: This block feels a little confusing to me. You get half of your warnings from the load_csv result object, half from the upload object. I'm not suggesting a major refactor at this point but the object encapsulation feels a little off.

app/controllers/uploads_controller.rb
```diff
@@ -12,22 +10,29 @@ def new
...
+
+      @upload.check_for_headers
+      headers = @upload.all_warnings
```
patrickvinograd: Variable name "header_warnings" would be more descriptive for future maintainers.

app/controllers/uploads_controller.rb
```diff
@@ -84,11 +82,16 @@ def upload_params
 
   def call_load
     file = @upload.upload_file.tempfile
-    data = @upload.csv_type.constantize.load(file, skip_lines: @upload.skip_lines.try(:to_i))
+    skip_lines = @upload.skip_lines.try(:to_i)
+    data = klass.load(file, skip_lines: skip_lines)
 
     @upload.update(ok: data.present? && data.ids.present?)
-    raise StandardError, errors_for_alert([@upload]) unless @upload.ok?
+    raise(StandardError, "There was no saved #{klass} data!") unless @upload.ok?
```
patrickvinograd: I don't think vets.gov has an error message style guide, but I tend to err on the side of business-like. I think we should remove the `!`.

app/models/upload.rb
```diff
@@ -1,37 +1,80 @@
...
+
     false
   end
 
-  def derive_dependent_columns
-    self.csv = upload_file.try(:original_filename)
+  def check_for_headers
+    return unless upload_file && csv_type && skip_lines
+
+    missing_headers.clear
```
patrickvinograd: This feels off to me - why do missing_headers and extra_headers need to be stateful in the Upload object? Why not have check_for_headers just return `missing, extra` and assign that to variables in uploads_controller, instead of setting state on the Upload object and then calling `upload.all_warnings`?

patrickvinograd (changes requested):
Some questions around object encapsulation.

app/controllers/uploads_controller.rb
mphprogrammer: Different types of warnings from different objects can occur at this stage. There are two types of issues this controller deals with - Record validation warnings with the CSV being uploaded and structural errors with the CSV file - which can now be independent of the data within the CSV.

app/models/upload.rb
mphprogrammer: Embedding the Error object in upload was a Kam suggestion - but for these purposes, either way works fine.

app/controllers/uploads_controller.rb
saneshark: So i might be responsible for this one. Perhaps we could make an alias on the 'all_warnings' 

The idea was to create a new "warnings" or "notices" handler that inherits from ActiveModel::Errors to cleanup some of the mess that was "Flashable."

I proposed calling it "warnings" because its equally generic to "errors," perhaps for this specific model we could assign an alias "header_warnings" but I do see value in keeping this concept "warnings" or "notices" (errors that are not actual errors preventing persistence of object) from becoming too specific to the context they are being user for.

app/controllers/uploads_controller.rb
mphprogrammer: Point taken...

saneshark: I definitely agree.

app/models/upload.rb
saneshark: @patrickvinograd I think we share differing opinions on this one. I am the one advocating for this change vs Marc's initial PR in which does precisely what you describe. 

The notices or warnings are specific to the upload instance, and I think it's best to keep that logic at a level where it can be easily unit tested vs at the controller level. I'm also a strong believer in the "thin" controller paradigm. I prefer cruft to be in the model or service level objects. 

That Flashable Mixin was a mess, and for me it's a welcome improvement to get rid of it.

saneshark: Most of the work that Flashable was doing is something that ActiveModel::Errors does much more effectively. By inheriting from that class and building "Notices" or "Warnings" it removes the need from having something like this. What Flashable was doing to me is almost the equivalent of 

object.errors.add(:base, 'some error message')  <- for a rails developer this is more of a common practice than some random mixin. Not to mention that the mixin was being added to dashboards and uploads controllers, even though only a subset of its methods were ever invoked in each.

mphprogrammer:
@saneshark Please check the dashboards_controller and upload model BEFORE committing. There was a conflict with your recent PR, not to mention a few rubocop errors to institutions_controller, and version spec, none of which this PR touched.

```
app/controllers/v0/institutions_controller.rb:90:53: W: [Corrected] Unnecessary disabling of Style/GuardClause.
    # rubocop:disable Metrics/CyclomaticComplexity, Style/GuardClause
                                                    ^^^^^^^^^^^^^^^^^
spec/controllers/v0/institutions_controller_spec.rb:142:1: C: [Corrected] Extra empty line detected at block body end.
spec/models/version_spec.rb:9:67: C: [Corrected] Prefer 1.day.
    let(:after_preview_date) { Faker::Time.between(preview_date + 1.days, preview_date + 2.days).utc }
                                                                  ^^^^^^
spec/models/version_spec.rb:10:91: C: [Corrected] Prefer 1.day.
    let(:before_preview_date) { Faker::Time.between(preview_date - 2.days, preview_date - 1.days).utc }
                                                                                          ^^^^^^
```

app/controllers/uploads_controller.rb
```diff
@@ -12,22 +10,29 @@ def new
...
-      display_failed_instances(data.failed_instances)
+      failed = load_csv.failed_instances
+
+      @upload.check_for_headers
+      headers = @upload.all_warnings
+
+      flash.alert = {}
+      flash.alert['The upload succeeded: '] = @upload.csv_type
+      flash.alert['But following rows should be checked: '] = failed.map(&:display_errors_with_row) unless failed.empty?
+      flash.alert['The following headers should be checked: '] = headers unless headers.empty?
```
saneshark: @mphprogrammer can we tidy this up a bit and use the same warnings mechanism? I'm  not following that's not just used.

app/controllers/uploads_controller.rb
```diff
@@ -12,22 +10,29 @@ def new
     @upload = new_upload(params[:csv_type])
     return if @upload.csv_type_check?
 
-    log_and_display_error(errors_for_alert([@upload]), 'Warning')
+    alert_and_log(@upload.errors.full_messages.join(', '))
```
saneshark: why do we need to log this?

app/controllers/uploads_controller.rb
```diff
@@ -12,22 +10,29 @@ def new
...
-      display_failed_instances(data.failed_instances)
+      failed = load_csv.failed_instances
+
+      @upload.check_for_headers
+      header_warnings = @upload.all_warnings
+
+      flash.alert = {}
+      flash.alert['The upload succeeded: '] = @upload.csv_type
+      flash.alert['But following rows should be checked: '] = failed.map(&:display_errors_with_row) unless failed.empty?
+      flash.alert['The following headers should be checked: '] = header_warnings unless header_warnings.empty?
```
saneshark: @mphprogrammer this is looking a lot better. But I wonder if we can clean this up here a bit. This logic seems a bit confusing to me.

app/controllers/uploads_controller.rb
mphprogrammer: Actually, the whole check is superfluous if we say that the only legitimate way to reach this action is via a link. Then there is no point in checking whether the argument is valid. However, if we do not say this (e.g., someone manually enters the url in a browser) then it needs to be checked.

app/models/upload.rb
patrickvinograd: Yeah, I'm fine with this being a method on the Upload object, and getting rid of the mixin stuff. I just don't quite understand why you need to call check_for_headers to set some state on the object, then immediately call all_warnings to get the warning messages (which is the only way this is used outside of specs) instead of just calling a method (like "check_for_header_warnings") that just returns the warnings messages directly. If for some reason this gets refactored it opens the possibility of somebody calling all_warnings but forgetting to call check_all_headers first.

mphprogrammer:
Any shot at pushing this?

patrickvinograd (approved):

patrickvinograd:
Switched to approved in the interest of getting this merged before the weekend. I'm not entirely convinced on the stylistic arguments but it's functionally correct so let's move forward.

app/models/upload.rb
saneshark: Yea I think it could be one call. We can revisit that.

app/controllers/uploads_controller.rb
saneshark: Checked sure. But logged? I thought that method is doing Rails.logger but you're not rescuing from any error

saneshark:
Also @mphprogrammer I think your rubocop is doing something weird. Specs would have never passed in my earlier PR. Please make sure you are running. Bundle exec rubocop -a, might have a newer or older version maybe.

PR #273, opened 2017-04-14
Removes falses from Institution export fixes #257

### LINKED ISSUE #257: "FALSE" in institution file should be blank
In the following fields a value of "False" should show up as a blank in the institution data download file:
* correspondence
* flight
* yr
* poe
* accredited

Here is what the current production data dashboard looks like:
image

Here is what the staging version is producing:
image

@mphprogrammer @saneshark

mphprogrammer:


patrickvinograd (approved):
LGTM. Glad this is well-contained to just the institution CSV export. I downloaded the current production export.csv and confirmed that there are no `FALSE`s anywhere in it, not just in the columns P.C. mentioned in the issue.

PR #274, opened 2017-04-17
Pv zero facet counts

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2196
Set boolean facets to a default value that still declares the possible dimensions (true/false) without bothering to calculate counts. Sets value to `nil` instead of `0` to better indicate intentional lack of information. Still return filter counts for other facets like country/state/etc because returning the counts is no more expensive than returning the list of countries/states/etc which are needed for populating dropdown menus. 

You'll see two commits, one with this behind a feature flag, then one without. I think the feature flag is unnecessary as it will still be total 1 PR to enable this whether we set a feature flag in devops or just do a code change here. Left a TODO as a breadcrumb.

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -77,14 +77,30 @@ module V0
         },
         type: institution_types,
         state: search_results.filter_count(:state),
-        country: embed(search_results.filter_count(:country)),
-        caution_flag: search_results.filter_count(:caution_flag),
-        student_vet_group: search_results.filter_count(:student_veteran),
-        yellow_ribbon_scholarship: search_results.filter_count(:yr),
-        principles_of_excellence: search_results.filter_count(:poe),
-        eight_keys_to_veteran_success: search_results.filter_count(:eight_keys)
+        country: embed(search_results.filter_count(:country))
       }
+      result.merge!(boolean_facets)
       add_active_search_facets(result)
+    end
+
+    DEFAULT_BOOLEAN_FACET = { true: nil, false: nil }.freeze
+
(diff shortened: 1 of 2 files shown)


app/controllers/v0/institutions_controller.rb
```diff
@@ -68,6 +68,12 @@ def search_results
                  .filter(:eight_keys, @query[:eight_keys_to_veteran_success]) # boolean
     end
 
+    # rubocop:disable Style/MutableConstant
```
patrickvinograd: Rubocop wants a `.freeze` on this, but then that causes an error somewhere in the JSON serialization of the response.

app/controllers/v0/institutions_controller.rb
```diff
@@ -78,11 +84,10 @@ def facets
         type: institution_types,
         state: search_results.filter_count(:state),
         country: embed(search_results.filter_count(:country)),
-        caution_flag: search_results.filter_count(:caution_flag),
```
patrickvinograd: Caution_flag is no longer a search dimension so it was removed intentionally.

app/controllers/v0/institutions_controller.rb
saneshark: That's really strange. What kind of error? I don't think I've encountered this type of issue before with ruby constants and serialization.

app/controllers/v0/institutions_controller.rb
saneshark: good catch.

saneshark (approved):

PR #275, opened 2017-04-18
Added field name to error message

mphprogrammer:


saneshark (commented):
Does this need specs?
Can key be null? if it is you would end up with a bunch of:  "" : ""

mphprogrammer:
The key/fieldname is in a map section so it shouldn’t be blank and would for the most part have been generated by ActiveRecord. Tested locally, and it seems good.
Doesn’t need specs, same error messages just added the column name to the message

saneshark (approved):
LGTM, would prefer to have specs for this, but not critical to merge.

PR #260, opened 2017-04-12
Initial revision of training document

mphprogrammer:

PR #278, opened 2017-04-20
Revision of training document

patrickvinograd:
Thorough editing pass based on Marc's v2 revisions.

Supplants #260


saneshark:
@patrickvinograd thanks for looking over this. I agree with all your changes. I believe there was a typo in one of your revisions "afte" should be "after," but I'm pretty sure spell-check would catch that minor revision.

We should probably get Natalie to review this as well, but I'm comfortable with merging before that happens. @mphprogrammer Do you agree with these proposed changes? I think this will clean things up a lot.

ayaleloehr:
GitHub with word docs is hard!

- There is an "Error! Bookmark not defined." in the table of contents
- I believe technical access and forgotten passwords should go to Patrick instead of me. Feel free to give yourself any reasonable title. 
- I trust you all know the workflow details far better than I do, so not going to have any edits there.

koconnor719:
yeah sorry guys, i'm happy to review but can't figure out where the doc even is--and i don't want to screw this up on you all...  @saneshark if you or leanna can give me some instructions or send me the doc i'll turnaround fast!  don't want to hold you up!

koconnor719:
@saneshark kam i think my name is just mentioned as the PM so that's ok w/ me!  thanks:)  i don't think there was anything else for me to comment on?  great job:)

patrickvinograd:
Incorporated feedback. @saneshark just need a formal :+1: so I can click merge.

saneshark (approved):

PR #280, opened 2017-04-26
Avoid double-http on institution and vet-policy URLs

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/2447

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -160,11 +160,13 @@ class Institution < ActiveRecord::Base
 
   def website_link
     return nil if insturl.blank?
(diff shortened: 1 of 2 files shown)


saneshark (approved):
So this whole fix is really crazy to me... I think this would be a good candidate for raising warnings to  the admins as well so that they could look to normalize this.

Also there is a ruby URI class that can be used to verify urls, but it is probably not necessary. For now this PR looks like path of least resistance. Thanks for also adding specs. I wonder if these specs cover the full gamut though. Should we maybe pull all instances in production today to see what they have?

PR #283, opened 2017-05-31
Providing temp storage for working on cvs. Fixes #282

### LINKED ISSUE #282: Data Service stripping "None's" from crosswalk file
The data service is converting and changing files that are uploaded into the data service.

The most problematic area this happens is that when a crosswalk file is uploaded most of the schools and employers had a "None" in the OPE and IPEDS column. The "NONE" is an important placeholder for us to let us know that that we already researched this school and their is not a ED equivalent. There were nearly 40,000 "None"'s that took over a year to research.

If someone then downloads the file they just uploaded the "NONE"'s are all stripped from the file. This is a BIG problem for us because we can't tell which schools we already researched.

I noticed this seems to be happening with other files that they are converted when uploaded and then what we download is different then what we uploaded.

Please help... we are going to need to reconstruct the crosswalk from an older file because of this.

@mphprogrammer @leannammiller @melwoodard

mphprogrammer:
This was (inadvertently) supported in old version, although not explicitly designed. Created a new resource so that the admins can keep editing Csvs, with the <b>understanding</b> that this does not constitution a version control. I am keeping it in a separate resource to avoid the potential for a work-in-progress being accidentally uploaded.


saintsoup52:
Thank you for doing this... this is going to save us countless man hours. Once this change is live do you mind giving us a quick primer on how it works?

mphprogrammer:
@saintsoup52 Yup (that is - I don't mind!) - it will be put in a PR today and pushed following a review.

saneshark:
@mphprogrammer lets hold off on review / merging of this one until we can determine how to better optimize generating institutions to prevent db locking issues. I have a feeling this might just compound that issue.

db/migrate/20170526180712_create_storages.rb
```diff
@@ -0,0 +1,20 @@
...
+
+
+      t.timestamps null: false
+      t.index :user_id
+      t.index :csv_type, unique: true
+      t.index :updated_at
+
+      t.timestamps null: false
+    end
+  end
```
saneshark: ✅ - This migration is OK.

saneshark (approved):
LGTM

mphprogrammer:
@saintsoup52 can you checkout the file db/seeds/02_constants.rb to make sure the changes we made to the GIBCT constants some time ago are correct in this older branch?

https://github.com/department-of-veterans-affairs/gibct-data-service/blob/raw/db/seeds/02_constants.rb

saintsoup52:
@mphprogrammer The constants look good in the link you sent me. :-)

PR #285, opened 2017-08-04
Redid Patch-1 Fork as a branch

mphprogrammer:
Changed seeds file per original saintsoup fork, which cannot be merged (for security reasons).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -173,13 +173,13 @@ GEM
     mime-types (3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2016.0521)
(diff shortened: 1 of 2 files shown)


saneshark (approved):
LGTM provided you get someone to verify the constants are correct, and can explain to me why mini_portile2 also got bumped in version number, probably OK, but I'm not familiar with this gem at all.

mphprogrammer:
@saneshark The mini_portile2 version number (now 2.2.0) was bumped by `bundle update nokogiri` to version 1.8 as a dependency.

saintsoup52:
These constants look good to me. Thank you for the quick turnaround @mphprogrammer !

PR #287, opened 2017-08-04
added rake task to update constants

mphprogrammer:
Broke out 02_data.rb from seeds into 02_constants and 03_data.
Added rake task db:update_constants to update constants table from console.

### DIFF
--- a/db/seeds/02_constants.rb
+++ b/db/seeds/02_constants.rb
@@ -1,0 +1,30 @@
+
+puts 'Creating sample constants'
+constants = {
+  'TFCAP' => 22805.34,
+  'AVGBAH' => 1681,
+  'BSCAP' => 1000,
+  'BSOJTMONTH' => 83,
+  'FLTTFCAP' => 13031.61,
+  'CORRESPONDTFCAP' => 11076.86,
+  'MGIB3YRRATE' => 1857,
+  'MGIB2YRRATE' => 1509,
+  'SRRATE' => 369,
+  'DEARATE' => 1024,
+  'DEARATEOJT' => 747,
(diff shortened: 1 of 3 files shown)


saintsoup52:
What does this do exactly? Aside from make the magic happen?

saneshark:
@saintsoup52 it allows us to update a table that stores the constants from a database seed file. The database seed file is typically used for building the table when the database is empty. What Marc has done here is made it such that, you can modify part of the database seed, and call a rake task to update only the constants.

mphprogrammer:
@saintsoup52 The latest version of GIBCT was written to draw the calculator constants from GIDS and those values were loaded in the database from a "seeds" file that runs once. Because the constants live in a table, simply changing the "seeds" won't change the data, because it really only runs once in production to initially populate the tables. This rake task can be run from the command line on the server, and it will populate the constants table with your changes.

saneshark (approved):
@mphprogrammer can we test this on dev/staging to make sure it works as expected before trying it on production? Should it be run during off-peak?

saintsoup52:
Thanks for the explanation!

Probably would be easier in the future to make the constants just another file that we can upload and update without having to do any coding/raking.

mphprogrammer:
I, having the brain of a dehydrated tardigrade, forgot that it was the seeds file we changed.

@saneshark yes, I can ask James to try it on the staging server first to make sure all is ok. And then have him do it in production if it works. There are only 23 rows that will be changed.

mphprogrammer:
Actually, the CSV metaphor for Constants would be a simple thing to do.

PR #291, opened 2017-08-08
Log exceptions to Sentry

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4158

Sentry is configured and gem is added but rescue block in api_controller is not re-raising so nothing ever makes it to Sentry unless we log it explicitly. If we decide later we need to filter/exclude anything, we can, but I'm guessing the range of GIDS errors will be much smaller than vets-api.


saneshark (approved):
LGTM 👍

PR #292, opened 2017-08-08
Avoid table-wide lock during institution build

patrickvinograd:
Fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/4195

See above for rationale of removing ALTER TABLE statements from transaction. 

Manually pushed to staging and confirmed that query requests can proceed while the generate preview transaction is ongoing.


mphprogrammer (approved):
Much better. Thanks for getting this.

saneshark (approved):
LGTM: 
- Bulk INSERT vs ALTER TABLE to avoid locking 👍 
- improvement on updating timestamps, in the same INSERT vs separate query.

PR #294, opened 2017-08-16
Constants

mphprogrammer:
Quick and dirty - can upload calculator constants using same metaphor as uploading CSV. If there is a later need, can add a show view for constants table.


saintsoup52:
YAY!!

spec/controllers/v0/calculator_constants_controller_spec.rb
```diff
@@ -4,10 +4,8 @@
 RSpec.describe V0::CalculatorConstantsController, type: :controller do
   describe 'GET #index' do
     it 'returns calculator constants' do
-      create(:calculator_constant, :float)
-      create(:calculator_constant, :float)
-      create(:calculator_constant, :string)
-      create(:calculator_constant, :string)
+      create(:calculator_constant)
+      create(:calculator_constant)
```
saneshark: NITPICK: 
- Putting this in a loop would help make it seem more deliberate than looking like a duplicate.

`2.times { create(:calculator_constant) }`

db/migrate/20170816193809_remove_string_value_from_calculator_constant.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveStringValueFromCalculatorConstant < ActiveRecord::Migration
+  def change
+    remove_column :calculator_constants, :string_value
```
saneshark: @patrickvinograd @aub this migration looks good to me, but do you want to double check? I think for migrations we cannot never be to safe. My understanding is that postgres just makes the column invisible and wouldn't actually remove it unless you do another operation. 

https://stackoverflow.com/questions/15699989/dropping-column-in-postgres-on-a-large-dataset

saneshark (commented):
@mphprogrammer This LGTM but I'd like to get one more set of eyes on it since it involves a database migration.

db/migrate/20170816193809_remove_string_value_from_calculator_constant.rb
mphprogrammer: I think there are 22 calculator constants when this migration is run. I looked at the `pg_attribute` table entries for `calculator_constants, and the `string_value` column is not listed. I  assume that were it present (even disabled) it would be in that list.

screen shot 2017-08-23 at 11 13 42 am

spec/controllers/v0/calculator_constants_controller_spec.rb
mphprogrammer: OK will change

db/migrate/20170816193809_remove_string_value_from_calculator_constant.rb
aub: Since there are so few rows I think we can do whatever we need to this table and it would be fine.

aub (approved):
LGTM

PR #295, opened 2017-08-24
Site

mphprogrammer:
Creates a site map for the GIDS/GIBCT.


saneshark:
@mphprogrammer this LGTM, but can we get a sample of the generate sitemap? We might have to do some url munging I would think seeing as F/E invokes GIDS via vets-api and additionally has its own concept of routes. It's those routes I would think that matter to google, not the API routes. 

And I might be totally wrong on this, but aren't sitemaps only really relevant to static pages? I would think we need to generate static pages for this to have any value, but I might be completely wrong on that.

mphprogrammer:
@saneshark I'm no pro-at this either (this is my first attempt), but it doesn't have to be static pages. For one thing, the institution show page has the facility_code embedded in the url, not in a query string, making it easier for search engines to work with. Also, the purpose of the sitemap is to point the crawler directly to the dynamic pages - negating the need for duplicate static pages, which would require notation that there is duplication in the robots.txt file. The sitemap contains links in the form:

```
  
  <url>
      <loc>https://www.vets.gov/gi-bill-comparison-tool</loc>
      <lastmod>2017-08-25T14:13:38-04:00</lastmod>
      <changefreq>always</changefreq>
      <priority>1.0</priority>
  </url>
  <url>
    <loc>https://www.vets.gov/gi-bill-comparison-tool/search</loc>
    <lastmod>2017-08-25T14:13:38-04:00</lastmod>
    <changefreq>monthly</changefreq>
    <priority>0.9</priority>
  </url>
  <url>
    <loc>https://www.vets.gov/gi-bill-comparison-tool/profile/00000002</loc>
    <lastmod>2017-08-25T14:13:38-04:00</lastmod>
    <changefreq>weekly</changefreq>
    <priority>0.8</priority>
  </url>
```
Through the search console (I think, and will know more tomorrow) I can register `www.vets.gov/gids/sitemap.xml` as the location of GIBCT's sitemap. Once that is done, everytime I ping google,

```
http://www.google.com/ping?sitemap=www.vets.gov/gids/sitemap.xml
```

it will reindex GIBCT using the GIDS /public url.

As far as I can tell (will verify) there is no requirement that the urls in the site map correspond to the base URL where the sitemap is located. So in effect, even though google crawls gids for the link, it returns gibct links in search.

robertfairhead:
For the historical record, I asked Marc via slack to remove the Google verification HTML file. We already verified with google (https://github.com/department-of-veterans-affairs/vets-website/pull/6183) and adding another would basically hijack control of the entire TLD.

I gave him access to the search console UI so he can submit the URL of the sitemap for google to start scraping. Google generally pulls them down once a day but the `ping` URL for google can be used to nudge it to reindex if there's an issue. So he should be set. If not, let me know and I'm happy to help.

No problem with any overlap in this sitemap and the main sitemap or google's indexing. Google wants to index all it can, so it will gladly visit the same URL multiple times.

mphprogrammer:
The idea behind leaving a sitemap on GIDS that contains GIBCT school profile pages (`www.vets.gov/gi-bill-comparison-tool/profile/999999`) is that GIBCT has no awareness of school additions and deletions, but GIDS does and can inform the crawlers as changes occur. Having multiple sitemaps for a single domain poses no additional issues or adverse effects. Remembering that there are 36,000+ profile pages makes publishing additional static pages that redirect to dynamic pages an extra step that isn't required.

public/robots.txt
```diff
@@ -3,3 +3,5 @@
 # To ban all spiders from the entire site uncomment the next two lines:
 # User-agent: *
 # Disallow: /
+
+Sitemap: https://www.vets.gov/gids/sitemap.xml.gz
```
patrickvinograd: AFAIK robots.txt can only be placed at `/` in a domain. Since GIDS as a whole is deployed to `www.vets.gov/gids`, will this even work?

app/controllers/dashboards_controller.rb
```diff
@@ -45,6 +45,9 @@ def push
 
       if pv.persisted?
         flash.notice = 'Production data updated'
+
+        # Build Sitemap and notify search engines
+        GibctSiteMapper.new(ping: true)
```
patrickvinograd: Approximately how long does this process take? Production publishing is fast right now, which is great; we don't want to lose responsiveness if we can help it. If it's a lengthy process maybe we ought to make it asynchronous.

app/models/gibct_site_mapper.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# Creates sitemips for gi-bill-comparison-tool
+class GibctSiteMapper
+  DEFAULT_HOST = 'https://www.vets.gov/gi-bill-comparison-tool'
+
+  def initialize(ping: true, default_host: DEFAULT_HOST, sitemaps_path: nil, verbose: true)
```
patrickvinograd: Since we never override `DEFAULT_HOST`, won't we generate a sitemap in dev/staging that points to www.vets.gov, potentially introducing wrong links (e.g. if an institution is added in staging but not yet in prod). If we need this to be configurable we can add an env var that distinguishes the host prefix.

patrickvinograd: Oh I see ping is false so google ought to never crawl sitemaps for dev/staging, so maybe this is okay.

patrickvinograd (commented):
Couple questions before we merge this.

public/robots.txt
robertfairhead: The existing robots.txt is at: https://github.com/department-of-veterans-affairs/vets-website/blob/master/assets/robots.txt and a second `sitemap:` line can be added there to point to this.

app/controllers/dashboards_controller.rb
mphprogrammer: Its rather quick, not even noticeable.

app/models/gibct_site_mapper.rb
mphprogrammer: @patrickvinograd I think the thing to do is only ping to true in production by examining the `original_url` in the request to see if it contains 'www.vets.gov'
 Do you see any issue?

saneshark (approved):
LGTM, @patrickvinograd are you cool with it too or did you still have comments you wanted addressed?

patrickvinograd (approved):
LGTM

PR #308, opened 2018-03-22
gem updates

wyattwalter:
Added an exception for rubocop in brakeman for now since it's a build-time vulnerability not a run-time issue. The upgrade is not huge for this codebase, but not something I can pause for at the moment. Created https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9391 to circle back.


patrickvinograd (approved):
LGTM

PR #310, opened 2018-03-23
update rails-html-sanitizer

wyattwalter:
Builds breaking due to one more gem that was released just after the last PR.

ref: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9387


patrickvinograd (approved):
LGTM

Issue #304, opened 2017-10-25
GI Bill CT Data Service is stripping quotes around OPE IDs


saintsoup52:
When a user clicks on the "download export csv" file, the GI Bill CT file is missing the quotes around the OPE ID numbers. For example American Public University (FC #21913148) should have an OPE ID "03819300" but the export file says [3819300].

This is problematic because most OPE ID's start with zeros and therefore the number  of characters for this field in csv changes.  Downstream this messes up many of the calculations for the GI Bill CT Data Dashboards because OPE IDs are use to group calculations for "All campuses"

This was an issue before #84 that was resolved but it appears to have come back again.

@leannammiller @mphprogrammer @patrickvinograd

mphprogrammer:
Sorry about that, in the meantime,  in the XLS you can format the column using custom and entering in "00000000", and change the column type to text - I think that works.

saintsoup52:
Unfortunately changing the formatting in excel doesn't change the calculation issues.

Thank you for looking at this.

saintsoup52:
Sorry to bug, but has there been any update on this issue?

jmcneal65:
Morning P,

First, are you back in the world?

Secondly, let me check with leanna on this…

With best personal regards,

       John

John T. McNeal
U.S. Department of Veterans Affairs
Veterans Benefits Administration, Office of Education Service
Business Process Development – Legacy Systems (224B)
Office: 202.461.9613 | iPhone: 202-957-2281





Confidentiality Notice: If you received this email by mistake, you should notify the sender of the mistake and delete the email and any attachments. An inadvertent disclosure is not intended to waive any privileges.

mphprogrammer:
@saintsoup52 This is only ope we are talking about, or is it cross and fac codes too?

saintsoup52:
Just OPE... the other numbers don't have IDs that start with zeroes so this isn't a problem anywhere else.

saneshark:
@saintsoup52 If I understand correctly, the issue is with Excel, not with the CSV export. Currently vets.gov is in effect returning everything as a String. Excel happens to automatically make assumptions about field type. I characterize the fact that everything is currently returned in a consistent manner as a feature, not a bug. If we were quoting certain fields/columns elsewhere, then I agree that this would constitute a bug.

saneshark:
[...]

PR #312, opened 2018-04-02
Fixed the OPE_ID quotes getting stripped issue

maneshkommidi:


nagaborra (approved):

saintsoup52:
Yay!!!

I am soo... soo... soo... excited about this!

patrickvinograd (commented):
(Mis-submitted review, see below)

patrickvinograd (changes requested):
This needs test cases to ensure that:
- The exported CSV has quotes preserved in the OPE_ID
- The data presented by the API is unaffected with quoted OPE_IDs.

patrickvinograd:
For example: https://api.vets.gov/v0/gi/institutions/11919137 has 
`"ope":"00321600"`
string key, string value. Make sure with this change that value looks the same, is not double-quoted or anything.

maneshkommidi:
@patrickvinograd added test cases and fixed the api response OPE field to have the exact string.

app/controllers/v0/institutions_controller.rb
```diff
@@ -29,7 +29,7 @@ def index
     # GET /v0/institutions/20005123
     def show
       resource = Institution.version(@version[:number]).find_by(facility_code: params[:id])
-
+      resource[:ope] = resource[:ope].tr('\"', '') if (resource.present? && resource[:ope].present?)
```
jkassemi: Is this a candidate for inclusion in the model or possibly something that can be worked into a converter process?

maneshkommidi: @jkassemi We cannot write this piece of code in to converter process,  as in CSV we need quotes for OPE_ID and coming to model I think it does not make big difference as we are querying the data here.

jkassemi: What about the serializer?

Would like, where possible, to maintain focus in the controller on authentication, input sanitization, data lookup, and orchestrating rendering. This seems like presentation logic within the controller.

maneshkommidi: Thanks, Updated the code

jkassemi (commented):
Thanks, @maneshkommidi, looks great! One more thing, can you add some specs for the `InstitutionProfileSerializer`?

maneshkommidi:
@jkassemi added spec's for  InstitutionProfileSerializer, when you get a chance please check and let me know your thoughts.

spec/serializers/institution_profile_serializer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    expect(data['id'].to_i).to eq(institution.id)
+  end
+
+  it 'links to vets api' do
+    expect(links['vet_website_link']).to_not be_empty
+  end
+
+  it 'should include OPE value' do
+    expect(data['attributes']['ope']).to eq(institution.ope)
+  end
```
jkassemi: So from my reading on this the ope value (`object.ope` from `InstitutionProfileSerializer `) can be one of three things before it hits the serializer:

1) a non-nil value containing a "
2) a non-nil value that doesn't contain a "
3) a nil value

So to make sure this feature is working as intended, we'd need to ensure that this spec hits each of those variants.

jkassemi:
@patrickvinograd mind taking another look?

jkassemi:
@patrickvinograd bump

patrickvinograd (changes requested):
I pulled this down locally and it does not appear to have the desired effect. I think it would work if OPEs were quoted in the input file (crosswalk.csv), but they are not. To be clear GIDS is handling the values correctly as strings all the way through upload, processing, cross-referencing, exposing via the API, it's just that when we *export* a CSV, we need to add quotes. 

While we could add quotes at import time, I believe this would be error prone in some of the processing/cross-correlation that is done among tables. Instead, I think the smallest change would be to just quote the ope value at export time. That could be done by introducing a "export_converter" field to the `CSV_CONVERTER_INFO` dict in the Institution model class, and using it in `lib/csv_helper/exporter.rb`. The vast majority of columns will have no export_converter so should pass through as-is, but the ope column of the Institution model will get a converter that quotes the values.

patrickvinograd:
Above suggested change should result in no impact to the API data or to how the institution table is built and stored, and only changes how it is exported as a CSV which is the desired effect.

saintsoup52:
The "input file" should have quotes around the OPE IDs... and they will in the future.

The problem is that the GIDS has been stripping them for so long that the files we are inputting now don't have the quotes. I can provide a new version with the quotes for testing.

jkassemi:
I reread the original issue https://github.com/department-of-veterans-affairs/gibct-data-service/issues/304 - I'd thought this was a fix for a different problem. There's no issue linked here, so assuming that's the one, and I'm closing. This is not the right fix for the problem.

PR #317, opened 2018-04-13
Docker-based CI

edpaget:
Everything else is using docker for CI, so this should too.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+tmp/
+data/
+log/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM centos:6
+
+# Match the jenkins uid/gid on the host (504)
+RUN groupadd -r gibct && \
+      useradd -r -g gibct gibct
+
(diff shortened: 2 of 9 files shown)


wyattwalter:
`make clean` results in some changes picked up by git:

```
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    log/.keep
```

wyattwalter (approved):
I adjusted the `make clean` task slightly to only modify the contents of `log/` rather than deleting it entirely to avoid issues with git and that task. Looks good to me!

wyattwalter:
Noticed on the change after I approved it that we were not doing a `make clean` at the end and leaving Postgres containers around. Added that post task so we wouldn't do that.

edpaget:
@maneshkommidi sorry this took so long to resolve, but once this is merged you'll need to merge master into your PR's branch and then it should work fine on Jenkins

maneshkommidi:
@edpaget Sure will do, Thanks

PR #318, opened 2018-04-18
SSOe integration

adi-pen:
Working on debugging the traits received from AccessVA and using them to validate with the records in the GIDS User table.


app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      user = User.from_saml_callback(response.attributes)
+      if user
+        sign_in_and_redirect user
+      else
+        flash.alert = 'User does not have a valid GIDS system account. ' \
+                      'Contact the system admin for further assistance.'
+        redirect_to root_url + '?auth=fail'
+      end
+    else
+      flash.alert = "Invalid SAML response. Errors: #{response.errors.join("\n")}"
```
patrickvinograd: I don't think this is the alert we want to display to users. Should be a nicer user-facing message.

patrickvinograd:
Just to make sure I'm understanding - you're still using Devise as the mechanism for managing user sessions, but not using its for authentication? Instead you take the SAML assertion and use that to initiate a Devise session?

app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      user = User.from_saml_callback(response.attributes)
+      if user
+        sign_in_and_redirect user
+      else
+        flash.alert = 'User does not have a valid GIDS system account. ' \
+                      'Contact the system admin for further assistance.'
+        redirect_to root_url + '?auth=fail'
+      end
+    else
+      Rails.logger.info("Invalid SAML response. Errors: #{response.errors.join("\n")}")
```
jkassemi: is there the potential for any PII to be included in the response errors?

app/models/user.rb
```diff
@@ -5,8 +5,15 @@ class User < ActiveRecord::Base
...
-  # :confirmable, :lockable, :timeoutable, :omniauthable and
+  # :confirmable, :lockable, :omniauthable and
   # :registerable
-  devise :database_authenticatable,
+  devise :database_authenticatable, :timeoutable,
          :recoverable, :rememberable, :trackable, :validatable
+
+  def self.from_saml_callback(response_attrs)
+    email = response_attrs[:va_eauth_emailaddress]
+    Rails.logger.info("email: #{email}")
```
jkassemi: Is there more context that can be provided on how this info call should be used?

spec/models/user_spec.rb
```diff
@@ -14,13 +14,10 @@
...
       end
 
-      it 'are properly formatted' do
-        expect { create :user, :bad_email }.to raise_error(ActiveRecord::RecordInvalid)
-      end
-
-      it 'come from .gov' do
-        expect { create :user, :bad_email_domain }.to raise_error(ActiveRecord::RecordInvalid)
-      end
+      # TODO: Add a complex regex or use email verification
```
jkassemi: Guessing there's no longer a requirement for a `.gov` email address?

Hopping into a meeting but didn't find the validation for that, either way.

app/controllers/auth_controller.rb
adi-pen: Not as per the [source code](https://github.com/onelogin/ruby-saml/blob/51ebceee5ef6e529159f88e18bcbd3f3ae6c53b5/lib/onelogin/ruby-saml/response.rb#L355).
I will go ahead and remove this line to be completely sure.

app/models/user.rb
adi-pen: Nope. It's a leftover from debugging.
Will remove

spec/models/user_spec.rb
adi-pen: @jkassemi That's right. There is one external user who would not be using a `.gov` email and will use `ID.me` to login.
Modified the regex in devise.

wunderhund (dismissed):
Wrong branch!

jkassemi:
@adi-pen - can you provide an update on this functionality and where testing is at?

adi-pen:
@jkassemi Dev testing is done. The SQA testing is in progress and is almost complete.
The next step is to configure GIDS with the pre-prod metadata.
It should be pushed to production after the successful completion of testing in pre-prod.

briangryth:
@jkassemi or @wunderhund - has the pre-prod metadata been loaded into Staging? I just heard from the IAM team that the metadata is getting a failure page in Pre-Prod, which would indicate that the metadata is not loaded.

wunderhund:
@briangryth I think I'm the wrong person to ask, I haven't had anything to do with this other than deploying it a few times. Is loading that metadata something that falls into the platform/devops sphere of responsibility?

wyattwalter:
We changed it and deployed new metadata files 3 days ago -- https://github.com/department-of-veterans-affairs/devops/pull/2890/files

jkassemi (approved):

PR #320, opened 2018-04-19
address issue with builds on branches with slashes

wyattwalter:
Same fix as https://github.com/department-of-veterans-affairs/vets-api/pull/1864

Addresses issues where Jenkins can't do builds if branch name has a slash in it due to invalid characters in Docker image name.


jkassemi (commented):
👍 thanks, Wyatt!

wunderhund (approved):
LGTM

Issue #314, opened 2018-04-12
Display non-approved schools in GI Bill CT Data Dashboards


saintsoup52:
In order to ensure the data integrity of the data dashboards, we should add the following data points ONLY to the GI Bill CT dashboards:
1. Include information for non-approved school
2. Add a column for approval status

In FY 2016 almost 6% of Post-9/11 GI Bill Students attended a facility code that is no longer approved in WEAMS, going back to FY 2013 that number increases to over 15%

Here is an example of why this matters:
I was asked a random question from a VSO earlier this week about why the enrollment data on GI Bill CT for the XXXX University system seemed way off.   The Comparison Tool was showing GI Bill enrollment at XXXX Universities at only 4,100, but the other data they were seeing said it should be more like 10,000 GI Bill students.

I think I found the issue… XXX (FC XXXX) had 5,198 GI Bill Students in 2016 but now its approval has been withdrawn in WEAMS which means it does not show up on the CT. What happened to XXX? Is it like XXX University and they just moved somewhere and their facility code changed? Either way this is definitely an argument for including non-approved schools in the data dashboards.

briangryth:
@nagaborra or @maneshkommidi - do you have specific questions that would help clarify this issue. 

The overall goal here is to allow non-approved schools to appear in the GIBCT Dashboard. Currently, only approved schools appear in the dashboard. 

Specific questions will help me get you the answers you need..

saintsoup52:
Not that you asked, but I am pretty sure this is where the database is filtering out the non-approved schools.

https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/spec/factories/weams.rb

As for the new column indicating approval status, it should be labelled something like:

**Column Header:** 
"approval_status"

**Possible Responses:** 
- "Approved"
- "Approved for chapter 31 only" [not currently in the screen above]
- "Not approved"

briangryth:
@nagaborra and @maneshkommidi does the above address your questions? Or do you need additional information? If so, please address let me know what you questions are? 

@saintsoup52 - are those the only possible responses? Or is that a question that Stephanie or Shay would need to verify?

maneshkommidi:
@saintsoup52 while going through the WEAM model I observed the current logic for approved school's is **poo-status** field value must be **aprvd** or **applicable_law_code** value must be '**educational institution is not approved'** or '**educational institution is approved for chapter 31 only**'. 

[...]

PR #321, opened 2018-04-24
Added approval status migrations

maneshkommidi:
https://github.com/department-of-veterans-affairs/gibct-data-service/issues/314


patrickvinograd (changes requested):
Overall this looks okay, but I think you need to think through the roll-out of this. As soon as you deploy, the institutions controller will start trying to filter on `approval_status`, but that won't exist in the database right away, because migrations aren't run strictly in sync with code getting deployed. So API requests would probably start breaking. 

Similarly even if the column is in the database, it won't be populated until somebody goes into the tool and creates a new version from the data. So this may need to be broken out into a few phases to get the data into place before using it.

wyattwalter:
@maneshkommidi that's exactly how we roll out features that need migrations. Db schema stuff goes into a PR, goes out and migrations run, then you can add the functionality that depends upon it.

maneshkommidi:
@wyattwalter Thanks, updated the PR, will add the filter logic once after the new version of data is ready.

patrickvinograd (approved):

patrickvinograd:
Looks good, sorry for the delay.

patrickvinograd:
I think there will actually need to be *three* phases:
1) this migration
2) code change to populate the approved status in both these tables
(followed by a publishing a new version of the data in each environment)
3) code change to start filtering based on the approved status.

maneshkommidi:
Yes @patrickvinograd we will address this in the same way.

PR #327, opened 2018-05-14
Adding migrations for yellow ribbon

maneshkommidi:


jkassemi:
@maneshkommidi - can you provide a link to the corresponding issue and some more context on how this table will be used?

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+  create_table "yellow_ribbons", force: :cascade do |t|
+    t.string   "facility_code"
+    t.string   "school_name_in_yr_database"
+    t.string   "school_name_in_weams"
+    t.string   "campus"
+    t.string   "street_address"
+    t.string   "city"
+    t.string   "state"
+    t.string   "zip"
+    t.string   "public_private"
```
jkassemi: Would like some more information on this column, seeing the following values:

* `- Please make a selection -`
* `NULL`
* `Private`
* `Private NP`
* `Private-NP`
* `Public`
* ``

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "degree_level"
+    t.string   "division_professional_school"
+    t.string   "contribution_amount"
```
jkassemi: Is this a nullable numeric?

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "updated_for_2011_2012"
```
jkassemi: All of the values in the source file are null or blank - is this supposed to be a boolean value?

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "missed_deadline"
```
jkassemi: would like to see this as a boolean, but I'm seeing a number of "- Please make a selection -" values. Can we get some clarity on this?

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "ineligible"
```
jkassemi: Same desire for a boolean here, but more of the "- Please make a selection -" values.

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "date_agreement_received"
```
jkassemi: Can this be a date column?

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "dat_yr_signed_by_yr_official"
```
jkassemi: Possibly a simple numeric, or ideally a date with a zeroed/default day/month.

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "amendment_date"
```
jkassemi: No non-null values in the source file yet, but I imagine this should be a date value

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "flight_school"
```
jkassemi: Requesting clarification here, too. Would like a boolean value, but we're seeing NO/YES/- Please make a selection - values.

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "date_confirmation_sent"
```
jkassemi: Date type preferred for this

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "consolidated_agreement"
```
jkassemi: Prefer boolean

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "new_school"
```
jkassemi: Prefer boolean

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "open_ended_agreement"
```
jkassemi: Prefer boolean

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "modified"
```
jkassemi: Prefer boolean

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "withdrawn"
```
jkassemi: Prefer boolean

db/schema.rb
```diff
@@ -1202,6 +1202,48 @@
...
+    t.string   "sco_name"
+    t.string   "sco_telephone_number"
+    t.string   "sco_email_address"
+    t.string   "sfr_name"
+    t.string   "sfr_telephone_number"
+    t.string   "sfr_email_address"
+    t.string   "initials_yr_processor"
+    t.string   "year_of_yr_participation"
```
jkassemi: Would prefer a date column here, but there are values that specify multiple years, and more of the "- Please make a selection - " values, too. I'm guessing we should just keep this a string value for now, but more clarity required.

jkassemi (commented):
Needs some feedback from the team that built out the source file, since there are some inconsistencies that I'd like to clarify before we set up the schema for import. 

It may be preferable to simply set up a table with columns for the data we need to use vs importing it all unless we can get some more consistency from the source.

maneshkommidi:
@jkassemi Sure will review the CSV as per comments and address it Thanks.

jkassemi:
Any updates on the field documentation?

Issue #325, opened 2018-05-10
STEM CIP code list and STEM Indicator


briangryth:
According to the Colmery Act, the GIBCT needs to include an indicator whether an educational institution offer STEM programs. Currently, the GIDS does not include a list of the STEM CIP codes. This ticket is to add that information to the GIDS. Once added, this list will need to be cross referenced with one to the IPEDS files to verify that an institution offers a STEM program. 

Here is additional information from Patrick Campbell regarding this requirement:
-  I pulled this list of STEM fields from an ED Report (see attached file below)
- National Center for Education Statistics, Table A-1. List of STEM fields and associated 2000 CIP codes
- The Classification of Instructional Programs (CIP) codes corresponding to science, technology, engineering, and mathematics (STEM) fields were selected based on a U.S. Immigration and Customs Enforcement list of designated STEM degree programs.
- SOURCE: U.S. Department of Homeland Security, Immigration and Customs Enforcement, STEM Designated Degree Programs, retrieved April 30, 2010, from http://www.ice.gov/sevis/stemlist.htm. 

**New Data File**
[STEM CIP Code List 20180504.xlsx](https://github.com/department-of-veterans-affairs/gibct-data-service/files/1990384/STEM.CIP.Code.List.20180504.xlsx)

*Associated Vets.gov-team Ticket* [Issue #9371](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9371)

maneshkommidi:
@briangryth Could you please let me know is this the correct approach
Do we need to import/export the stem_cip_code.csv in to GIDS and generate the boolean value in any of the existing table for stem check, if so what is the business logic to generate the stem check.

briangryth:
@saintsoup52 - Do you have a recommendation for @maneshkommidi questions above? 

I believe we would need to import the stem_cip_code.csv into GIDS. 
That file would be used to check against an IPEDS file and if any of the stem codes match then the STEM program indication in GIBCT should be marked as TRUE.

saintsoup52:
@briangryth @maneshkommidi  Here is how I would do it... 

I would allow EDU to upload the two files we need into the GIDS
1) STEM CIP List (file above)
2) IPEDS CIP File (see recent email)

[...]

PR #329, opened 2018-05-21
STEM program availability

maneshkommidi:
Provides a `stem_offered` boolean value on an `Institution`. Exposes this value in the institution index as a new facet (filterable with the `stem_offered` GET parameter), and to the  institution profile endpoint.

This value is determined based on two new data files, IpedsCipCode and StemCipCode. The former contains a reference to programs by CIP code that are supported by an institution, and the latter contains a set of CIP codes that are applied to STEM programs. The Institution Builder will correlate the two new data sources to determine if an institution offers a STEM program.

The `institutions.csv` export contains a `stem_offered` boolean value.

Local Reference: https://github.com/department-of-veterans-affairs/gibct-data-service/issues/325
Consumer: https://github.com/department-of-veterans-affairs/vets-website/pull/7641
Internal: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9371


db/migrate/20180521180832_create_ipeds_ciplists.rb
```diff
@@ -0,0 +1,14 @@
+class CreateIpedsCiplists < ActiveRecord::Migration
+  def change
+    create_table :ipeds_ciplists do |t|
+      t.string :cross, null: false
```
jkassemi: minor - can use `index: true` option here instead of separately adding an index

db/migrate/20180521181959_add_stem_offered_to_crosswalks.rb
```diff
@@ -0,0 +1,5 @@
+class AddStemOfferedToCrosswalks < ActiveRecord::Migration
+  def change
+    add_column :crosswalks, :stem_offered, :boolean
```
jkassemi: Can you describe the process you're thinking through? 

My understanding is that crosswalks is an EDU provided upload, and that we're using the CSV from that file to populate this table. Are you thinking of setting stem_offered based on the process from https://github.com/department-of-veterans-affairs/gibct-data-service/issues/325#issuecomment-390193350? 

Forgive my ignorance here - I'm not 100% fluent on the details of GIDS, since I've only made relatively minor changes to it so far. Is `stem_offered` not supposed to be a new attribute on the institution?

maneshkommidi: Yes @jkassemi here I'm planning to generate the stem_offered in crosswalks based on the logic in comment.

jkassemi: So, `stem_offered` will be a new column on the `institutions` after they're built? 

If we're currently just mapping the source file to the crosswalks table, then this approach would mean we'd need to regenerate that any time either crosswalks or IPEDS was updated. Is it just possible to set the value on the institution record on rebuild so we can keep these tables representative of input data and not modifications our system is making?

maneshkommidi: Yes @jkassemi stem_offered will be new column on the institutions will add migration as a part of this PR

db/migrate/20180525141749_create_stem_cipcodes.rb
```diff
@@ -0,0 +1,11 @@
+class CreateStemCipcodes < ActiveRecord::Migration
+  def change
+    create_table :stem_cipcodes do |t|
+    	t.integer :two_digit_series
+      t.string :twentyten_cip_code
```
jkassemi: minor nit - consistency in spacing

db/migrate/20180525130650_add_stem_offered_to_institutions.rb
```diff
@@ -0,0 +1,5 @@
+class AddStemOfferedToInstitutions < ActiveRecord::Migration
+  def change
+    add_column :institutions, :stem_offered, :boolean
```
jkassemi: @maneshkommidi what's the null state of this column indicate?

saneshark (approved):
LGTM

Issue #324, opened 2018-05-10
Housing Allowance by Zip Code 


briangryth:
As part of the Comery Act changes the GIBCT is required to allow a Veteran to be paid the housing allowance according to where a credit is obtained (e.g. paid at the rate for the city where an internship is  taken). This housing allowance by ZIP code is not currently in the GIDS. Thus, we need to add that file to the GIDS and make the data accessible to the GIBCT front end

**New Data Files**

[BAH Rates 2017_18.csv.zip](https://github.com/department-of-veterans-affairs/gibct-data-service/files/1990371/BAH.Rates.2017_18.csv.zip)
[LTS_EDU_DOD_BAH_RATES_2018-2019_05022018.txt](https://github.com/department-of-veterans-affairs/gibct-data-service/files/1990373/LTS_EDU_DOD_BAH_RATES_2018-2019_05022018.txt)
[LTS_EDU_ZIPMHA_2018-2019_05022018.txt](https://github.com/department-of-veterans-affairs/gibct-data-service/files/1990374/LTS_EDU_ZIPMHA_2018-2019_05022018.txt)

**Note from Patrick C and Education Services staff**
Housing Allowance rates by zip code (with grandfathered rates)
- I put this together myself… EDU might want to organize it in another way, but this it will be something like this
- You will see that I had to put quotes around the zips so the leading zeroes didn’t drop off… sorry
- We should be sure to have error codes if the zip codes aren’t found in our list or the rates return as zero

*Vets.gov-team related ticket* - [Issue #9368](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9368)

maneshkommidi:
@briangryth @saintsoup52 could you please let me know what is the use of .txt files(LTS_EDU_DOD_BAH_RATES_2018-2019_05022018.txt, LTS_EDU_ZIPMHA_2018-2019_05022018.txt)

jkassemi:
@maneshkommidi, will also need an implementation plan for this functionality to assist with assessing delays.

maneshkommidi:
@jkassemi as already related migrations are merged in, know we need to implement search functionality based on zip code.

saneshark:
@saintsoup52 for now, I am going to manually merge these two files into 1 and populate the table using a rake task. I will not introduce the functionality to do this via upload until we have a chance to discuss this further. I think that should be OK since it appears these do not get updated very regularly.

[...]

PR #330, opened 2018-05-22
Adding zip code migrations and CSV file 

maneshkommidi:
https://github.com/department-of-veterans-affairs/gibct-data-service/issues/324


db/migrate/20180522142853_create_zipcode_rates.rb
```diff
@@ -0,0 +1,14 @@
+class CreateZipcodeRates < ActiveRecord::Migration
+  def change
+    create_table :zipcode_rates do |t|
+      t.string :zip_code
+      t.string :mha_code
+      t.string :string
```
jkassemi: What's this column for?

db/migrate/20180522142853_create_zipcode_rates.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.string :mha_name
+      t.string :mha_rate
```
jkassemi: Should `mha_rate` and `mha_rate_grandfathered` be strings?

db/migrate/20180522142853_create_zipcode_rates.rb
```diff
@@ -0,0 +1,14 @@
+class CreateZipcodeRates < ActiveRecord::Migration
+  def change
+    create_table :zipcode_rates do |t|
+      t.string :zip_code
```
jkassemi: Does this need a `version` in order to track entries based on the source file that was uploaded?

db/migrate/20180522142853_create_zipcode_rates.rb
maneshkommidi: Sorry @jkassemi will correct it

db/migrate/20180522142853_create_zipcode_rates.rb
maneshkommidi: I do not think for this model we need version.

jkassemi: FYI I may be misguided on ^.

maneshkommidi: I don't think we will need version for this model

jkassemi (approved):
Nothing major pops out, so 👍 - we can iterate as necessary prior to this getting hooked up for production use.

Issue #323, opened 2018-05-10
Yellow Ribbon File 


briangryth:
As Part of the Colmery Act changes the GIBCT is required to expand the amount of information about Yellow Ribbon programs is displayed. The data required to be displayed is not currently in the GIDS. 

This ticket is intended to add a new Yellow Ribbon data file to GIDS and make that data accessible to the GIBCT front-end. 

**New Data File**

[Master Copy of Yellow Ribbon.xlsb.zip](https://github.com/department-of-veterans-affairs/gibct-data-service/files/1992088/Master.Copy.of.Yellow.Ribbon.xlsb.zip) *Update at 5/10/2018 at 11:00 AM EDT*

*Associated Vets.gov team ticket:* [Issue #9365](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9365)

briangryth:
@jkassemi @maneshkommidi - please note that data file has been replaced as on 5/10 at 11:00 am EDT. The file in the description is correct.

maneshkommidi:
@jkassemi Sure will do

maneshkommidi:
@briangryth Could you please let me know, what are all the columns that needs to be send to GIBCT from GIDS.

saintsoup52:
We definitely need to use the following four columns from the YR file in the GIDS:
- Degree Level	
- Division Professional School	
- Number of Students	
- Contribution Amount

I need to figure out from EDU if there any rules to using this spreadsheet... e.g., are there some YR agreements that should NOT be appearing.

jkassemi:
@maneshkommidi would appreciate the details on the implementation plan (GIDS API changes, and an overview of how they'll be connected to the new migrations) - would like to piece together any expected delay on deliverable.

maneshkommidi:
@jkassemi once the migrations PR merged, will need to work on sending the YR related data to website through institutions hash.

jkassemi:
What changes to the API are required to expose this new data, and how will it be exposed? Will there be a new attribute on the existing institutions API?

Also, how will the changes to the institution builder be made? I'm anticipating (but could be wrong, so I'm hoping for clarification here) that we'll either need a new table to store and version this information on publishing a new version, or we'll need to store this data in a new column on the institutions table.

saneshark:
Code has been merged to staging and production! 🍻 🍾 🎉

PR #332, opened 2018-06-04
Yellow Ribbon Program support

jkassemi:
Provide the ability to upload Yellow Ribbon Program data provided by EDU
team. A new file upload option: "Yellow Ribbon Program Source" is
provided. General validation and conversion is supported, though some
fields that are only used on re-export of data source will need to be
worked through with the EDU team in more detail to ensure we have
consistent and correct input data.

Yellow Ribbon Program Source records are created for each row in the
input file. The Institution Builder is modified to create a new Yellow
Ribbon Program record for each program that can be associated with an
institution. Noticing that we're not seeing a full match to all programs
in the sample data file, so we may need some sort of cross-reference
check with institutions, though the naive approach to this would require
that source files be uploaded in a specific order, so we're going to
want to avoid that.

The Yellow Ribbon Program records corresponding with an Institution are
exposed as part of the institution profile endpoint. Swagger
documentation in Vets-API will be provided shortly to reference this new
functionality.

```
{
   ...
   "data" : {
      "id" : "1565",
      "attributes" : {
         "facility_code" : "21905928",
         ...
         "yellow_ribbon_programs" : [
            {
               "division_professional_school" : "All",
               "number_of_students" : 1000,
               "degree_level" : "All",
               "contribution_amount" : "10000.0"
            }
         ],
	...
}
```

Supersedes: #327 
Local Reference: #323 
Consumer: https://github.com/department-of-veterans-affairs/vets-website/pull/7654
Tracker Reference: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9365

cc @briangryth


db/migrate/20180603183002_create_yellow_ribbon_programs.rb
```diff
@@ -0,0 +1,14 @@
+class CreateYellowRibbonPrograms < ActiveRecord::Migration
+  def change
+    create_table :yellow_ribbon_programs do |t|
+			t.integer :version, null: false, index: true
```
jkassemi: @jkassemi - need to fix spacing here

app/models/institution.rb
```diff
@@ -153,6 +153,8 @@ class Institution < ActiveRecord::Base
   validates :version, :institution, :country, presence: true
   validates :institution_type_name, inclusion: { in: TYPES }
 
+  has_many :yellow_ribbon_programs
```
saneshark: adding dependent destroy here makes sense I believe, unless this should really be a `has_many :through` relationship with a join table?

app/models/yellow_ribbon_program.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+class YellowRibbonProgram < ActiveRecord::Base
+end
```
saneshark: should we define the `belongs_to` association?

spec/models/converters/date_converter_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  it 'converts strings to dates' do
+    expect(subject.convert('4/17/2017')).to eq(Date.parse('April 17, 2017'))
+  end
+
+  it 'converts blank value to nil' do
+    expect(subject.convert('')).to be_nil
+  end
+
+  # TODO: we should find a way to pass warnings off for data that doesn't parse
+  it 'converts bad date to nil' do
```
saneshark: is this TODO comment something that needs to be addressed prior to merge?

saneshark (commented):
@jkassemi this looks good to me, but I had a few comments.

saneshark (approved):
This LGTM as is, but please see the outdated comments and changes I made. And of course lets wait on merging until tomorrow (post migration).

app/models/institution.rb
jkassemi: 👍 should be good for the dependent destroy.

spec/models/converters/date_converter_spec.rb
jkassemi: I think we'll need to rethink the approach to the converters in order to surface these errors, so something that needs to be done later.

PR #333, opened 2018-06-08
Merge database migrations first

saneshark:
- These are the migrations from each of the three PRs in the appropriate order.
- Will need to remove the migrations from each of the PRs since some of the migrations had to be modified. But holding off on doing that until this PR is merged so that I can upmerge the other PRs for schema changes.


saneshark:
@jkassemi I noticed that this repo doesn't have zero-downtime gem for testing database migrations. Do you think that is worth adding? Is this something we would want @wyattwalter to run off peak hours or do think it will be ok to run at any time?

lihanli (approved):

jkassemi:
@saneshark let's not throw that gem in place just yet - we can do that along with squashing existing migrations in the future so we don't have to exclude migrations manually.

db/migrate/20180525130651_add_stem_offered_to_institutions.rb
```diff
@@ -0,0 +1,5 @@
+class AddStemOfferedToInstitutions < ActiveRecord::Migration
+  def change
+    add_column :institutions, :stem_offered, :boolean, default: false, index: true
```
jkassemi: @saneshark can you break this migration out so we don't lock the table? Use a change default in a separate migration? Not concerned with the index, since that won't lock reads.

jkassemi (commented):
One update requested for the default value. 

I'd opt for the after-hours run vs coordinating with anybody who may be using the dashboard.

jkassemi (approved):
🚀

PR #338, opened 2018-06-12
only build-and-deploy on ssoe branch

wyattwalter:
Super duper hacky, but we want this branch to be the auto-deploy one for dev and staging for now.

This is causing issues with folks testing SSOe in that branch because merges to master are auto-deploying to staging.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ pipeline {
     }
 
     stage('Deploy dev and staging') {
-      when { branch 'master' }
+      when { branch 'pv-ssoe-integration' }
 
       steps {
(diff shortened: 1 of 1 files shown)


wyattwalter:
This (long-lived feature branches) reeeeaaallly doesn't play well with our current deployment model. It requires us to do an additional build (manually) in order to do a deploy in prod. We can certainly write up a solution, but I'd really like to figure out how to make this go away soon instead.

jkassemi (approved):
👍 to get this out and tested. I anticipate we may have some issues testing and deploying changes @saneshark is working into GIDS with this configuration, so cc'ing here. This branch will just need to be updated frequently over the next week.  

@saneshark, @wyattwalter let me know if there are bandwidth problems / complications from this and I can assist.

PR #339, opened 2018-06-13
Housing Allowance Rates (upload / seeds/ and controller)

### LINKED ISSUE #324: Housing Allowance by Zip Code

saneshark:
fixes: https://github.com/department-of-veterans-affairs/gibct-data-service/issues/324


db/seeds/02_zipcode_rates.rb
```diff
@@ -0,0 +1,59 @@
...
+    }
+    SmarterCSV.process(ZIP_CITY_PATH, options).group_by { |record| record[:zip_code] }
+  end
+
+  def insert_data
+    records = []
+    mha_rates = mha_rate
+    zip_cities = zip_city
+    zip_mha.each do |zip_mha_record|
+      mha_rate_data = mha_rates[zip_mha_record[:mha_code]].first.slice(*%i[mha_rate mha_rate_grandfathered])
```
jkassemi: I didn't catch any in this `mhv_rate.csv`, but maybe a TODO comment here about throwing an exception if we have multiple rates associated with a code?

db/seeds/02_zipcode_rates.rb
```diff
@@ -0,0 +1,59 @@
...
+      mha_name_data = { mha_name: zip_cities[zip_mha_record[:zip_code]]&.first&.slice(*%i[city state])&.values&.join(', ') }
+      records << ZipcodeRate.new(zip_mha_record.slice(*%i[zip_code mha_code])
```
jkassemi: My take on this is the name is just for display under "What is the ZIP Code where you'll be located?" in https://github.com/department-of-veterans-affairs/vets.gov-team/issues/9368#issuecomment-382547068.

If that's the case, I'd suggest we drop the name since your observations about "primary" name would likely be confusing. Maybe address in a separate refinement of the tool?

jkassemi (commented):
@saneshark, thank you!

I have one question about the MHA name that I think you're also needing some clarity on.

db/seeds/02_zipcode_rates.rb
jkassemi: @briangryth pulling you in to see if you have any clarifying comments on this one. Essentially, if ^ is correct, what do we do when a zip code covers more than one city/state?

db/seeds/02_zipcode_rates.rb
lihanli: how about activerecord validation for that?

saneshark: I think this is a temporary solution until we can get a single consolidated file (sometime in the future I suppose based on comments made in the issue).

It isn't really possible to have more than one MHA for a given zip code. The only reason I'm calling `.first` on it is because the group_by function always returns an array, even when there is only one member in the grouping.

db/seeds/02_zipcode_rates.rb
saneshark: according to @saintsoup52 comments there is a way to get the name using crosswalks, but I'm not entirely sure how to do that just yet. I'm also thinking that perhaps the consolidated file that they provide could also include the name so that there is no need to build the dynamically, and it will just be a direct import instead.

jkassemi: 👍 interested in getting additional details ASAP, but I don't see problems moving forward with this and getting the API schema in place.

lihanli (approved):

jkassemi (approved):

app/controllers/v0/zipcode_rates_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module V0
+  class ZipcodeRatesController < ApiController
+    skip_before_action :resolve_version
```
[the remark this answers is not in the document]
jkassemi: Yes, but I agree that it's safe to add that functionality when the data is built from upload sources.

We'll base this on the existing pattern - uploads into new tables that are cleared with each upload (I started a `_sources` suffix with the yellow ribbon program sources to differentiate tables that are simply storing the latest upload data), and a `version` on the `zip_code_rates` table which will be populated when the InstitutionBuilder process is started. We can flesh that out and maybe do some table renaming for clarity with that round of development.

spec/request/zipcode_rates_request_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'support/controller_macros'
+require 'support/devise'
+
+RSpec.describe 'institutions', type: :request do
```
jkassemi: s/institutions/zipcode_rates

spec/request/zipcode_rates_request_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:valid_zip_code) { '12345' }
```
jkassemi: can we use these or strip them?

jkassemi (changes requested):
Minor changes requested. 

Feel free to ping me via slack or text when they're addressed for another review. We're also probably going to need to make sure the swagger docs in vets-api get updated. Not sure if that'll block FE or if they can get started as soon as the endpoint is shipped by playing with the service in dev (cc @rianfowler).

jkassemi (approved):

saneshark:
Added the ability to upload the zipcode_rate.csv in the same format that @saintsoup52 provided the xls file. It must however be a csv file. No validations on the model at present.

screen shot 2018-06-19 at 3 30 36 am

screen shot 2018-06-19 at 3 30 50 am

IMPORTANT NOTE: unlike the other uploads, this upload will affect the GIDS immediately. Other uploads require you to preview / generate, but this particular upload is in no way to tied to institution builder and as such, uploading it will have an immediate affect on the data without PREViEW.

db/seeds/03_data.rb
```diff
@@ -5,6 +5,8 @@ def seed_table(klass, user, options = {})
 
   puts "Loading #{klass.name} from #{csv_path}/#{csv_name} ... "
 
+  binding.pry if csv_name == 'zipcode_rate.csv'
```
jkassemi: ❌

jkassemi (changes requested):
Great that we can make changes to this without an overly manual process. Some minor concerns that this will get used when it shouldn't prior to supporting versioning. 

Is this (other than the binding.pry) ready to go?

saneshark:
@jkassemi 

- i reverted back to requiring the rake task to populate `ZipcodeRate` with `sample_csvs/zipcode_rate.csv` because I was concerned with the interface being non intuitive, changes making their way into production upon upload without a preview option.

- I kept some of the logic for merging those two text files in case it proves useful when we might want to revisit building this table like we do for institutions but I think we're going to need design and development work on how that would be done. I do not know how to get the `mha_name` from crosswalks for example. By changing the extention to `.ignore` the seed task will skip over the file, but it's there in case we need to revisit in the future.

- I updated the seed task and moved some of the code into a module mostly because I wanted to leverage some of the code for the `db:update_zipcode_rate` rake task. I also felt it was safer to namespace the methods.

- The additional files that are modified are just rubocop style changes for files that were merged prior to the changes that updated rubocop. Or so I surmise. Not sure why the other commits didn't have a problem with them though.

See also: https://github.com/department-of-veterans-affairs/vets-api/pull/2065
These vets-api changes are required for testing locally, should be ok to merge vets-api independently as long as the FE can do their stuff behind a feature toggle.

jkassemi (approved):

Issue #336, opened 2018-06-12
Create migration to support adding column to WEAMS file and the corresponding migration for changes to institutions table.


briangryth:
The migration for the WEAMS file needs to be update to include following new columns:
·         Priority Enrollment
·         Online Only
·         Independent Study
·         Distance Learning

Corresponding changes need to be made to the Institution table.

Sample of the new WEAMS file. 

[WEAMS Extract expanded clean.zip](https://github.com/department-of-veterans-affairs/gibct-data-service/files/2095808/WEAMS.Extract.expanded.clean.zip)

saneshark:
Questions:
1. I have downloaded the extract, it appears as though all of the values for the newly created columns are Null. There is no real data available other than the column headers in the above extract. Is this by design or should there be T / F values for some of these?

2. Should Null be interpreted as false?

3. Do we have to continue supporting the possibility that an old style WEAMS csv would get uploaded? Or moving forward after this code is merged can we expect that only CSVs with appropriate columns available will be uploaded (NOTE it will be much lower effort if we dont need to be backward compatible with legacy CSV uploads).

briangryth:
Okay here is the information I know:

Question 2. Should Null be interpreted as false?

Answer: According to the WEAMS additions requirements (see below), the value can be "True or False or Null". However, on the front-end we will treat Null as false and not display the Independent Study and Priority Enrollment indicators. Thus, we need to support a file have a NULL value, but I don't see a problem interpreting that as false. 

[Requirments PIR 18-051 - Task 613055 -  Add Checkboxes to COMP tool Interface.pdf](https://github.com/department-of-veterans-affairs/gibct-data-service/files/2103413/Requirments.PIR.18-051.-.Task.613055.-.Add.Checkboxes.to.COMP.tool.Interface.pdf)

Question 3. Do we have to continue supporting the possibility that an old style WEAMS csv would get uploaded? Or moving forward after this code is merged can we expect that only CSVs with appropriate columns available will be uploaded (NOTE it will be much lower effort if we dont need to be backward compatible with legacy CSV uploads).

Answer: My understanding is that we would have to support the old style WEAMS csv, which is not ideal. The WEAMS system that generates the file is being updated later this year (maybe as early in the summer). Until then we have to be backward compatible. Also on this item @saintsoup52 has stated in emails that

[...]

PR #340, opened 2018-06-14
Migrations for WEAMS and Institutions

saneshark:
Still have questions posed in the issue below.
Will likely continue to work out of this branch and then pull migrations out at the end.
Fixes: https://github.com/department-of-veterans-affairs/gibct-data-service/issues/336


saneshark:
I think we probably want to merge this in and make sure these get run on staging and production before any actual code is introduced.

jkassemi (approved):
👍

PR #343, opened 2018-06-21
update the deploy to load seeded values from a sql dump

saneshark:
This was just an experiment to see if I could speed things up a bit.


saneshark:
meh, closing this, it's really not that big of a problem.

PR #346, opened 2018-07-02
Independent study and priority enrollment exposed

jkassemi:
* Expose the independent_study and priority_enrollment attributes from
the Institution model to the collection and detail endpoints.
* Provide filters and facets for both of these attributes for search

Follows existing patterns. I've added an exclusion for class length on rubocop for this change - I don't think a refactor is warranted here yet, but opinions and suggestions would be great.
 
cc @rianfowler


U-DON (approved):
LGTM. One potential correction.

PR #348, opened 2018-07-17
allow searching for address in School Autocomplete widget api

lihanli:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11469
modifies autocomplete api to allow an optional `include_address` param which will use the search term to query address fields

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,8 +57,6 @@ gem 'figaro'
 # Sentry
 gem 'sentry-raven', '~> 2.3.0'
 
-gem 'zero_downtime_migrations'
-
 group :development, :test do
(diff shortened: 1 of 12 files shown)


saintsoup52:
This is exciting... How will this new search pattern work exactly?

Any way we could have an advance search function? Instead of searching by name, search by state or zip code.

lihanli:
this api will be used by the gi bill complaint tool to look up schools by name or address
we could modify it to search by state or zip code, @rtluu is my product manager, you can talk to him about feature suggestions

saintsoup52:
Thank you for the explanation... I just don't know why anyone would look up a school by address (vs state or zip).

jholton (approved):
LGTM

PR #349, opened 2018-07-30
WEAMS - add appropriate columns when building institutions table

saneshark:


saneshark:
We're not ready for this yet.

PR #352, opened 2018-08-10
quoted OPE in exports

jholton:
As written, this will add quotes around OPE in _all_ exported reports that contain OPE. 
Please advise if this should only add quotes in the two exports specifically mentioned in the ticket.

In the exported CSV, these will render triple-quoted. e.g. `"""00884300"""`
This does match the way these are quoted in our sample input files.

I can generate a sample output file for testing.  I am unable to test myself as I do not own Microsoft Office.


jkassemi (changes requested):
I don't _think_ this quite does the trick. I'm seeing quoted values, but I think we can coerce Excel into treating the column as a string type so that the quotes don't actually end up part of the data.

I'll see if I can find a formatting trick that does the job locally. Looks like we'll also have to catch a few more key names for fields specified in https://github.com/department-of-veterans-affairs/gibct-data-service/issues/304#issuecomment-412531958

jkassemi:
@jholton your change displays the value within quotes in Excel (`"03458300"`). I was also able to successfully import the download and the associated `weams` record still has the correct value (`03458300`). 

The change I just committed displays the value within Excel as `03458300`, which is great. But trying to import that file directly into GIDS results in a `weams` record with `=03458300`. So I'm unable to import the file as it's exported. That said, opening it in Excel, and then exporting it as a CSV results in a file that can be uploaded into GIDS without issue. I'm going to revert this change now.

Can you work with @saintsoup52 to establish the workflows around these files? What you had originally may work (though we still need to get those additional columns in place).

jholton:
Will do.  However I think [#304 comment](https://github.com/department-of-veterans-affairs/gibct-data-service/issues/304#issuecomment-412531958) was just listing other reports that contain the OPE field and not other fields that need formatting.

> That said, opening it in Excel, and then exporting it as a CSV results in a file that can be uploaded into GIDS without issue.

What did the data look like in the CSV in a text editor after exporting it?

saintsoup52:
We have tried mightily to have an excel work around for this... even if we figure it out in excel, it still messed up when other people use the data in their excel files.

Can we please just have the quotes? Or any other non-number character?

We only use quotes because that is what ED does on their reports for OPE numbers.

jkassemi:
I'm 👍 quotes in the excel field if that works for everyone involved.

@saintsoup52 it would be useful for us to understand more details on how this file is generated, how/what modifications are generally made, how the export itself is used, who's using it, and how they're viewing/exporting it (Excel, CSV, etc). I'd like to know what options we have for future efforts, and better understand whether or not things like moving to an Excel native format would be beneficial.

jkassemi:
@jholton raw output looks like `"""03458300"""`

saintsoup52:
@jkassemi The crosswalk and final output file (dashboard) are the most important.

The crosswalk file is usually exported from the data service, updated and reimported back into the data service. The crosswalk file does not live anywhere else (e.g., share drive) because it is so hard to keep up to date and I need to have access to it. The actual process is of updating it quite laborious, but that is not a problem you need to work about.

The final output file (or data dashboard) is cut and paste into the file users can download from the CT with information about all schools. This is downloaded thousands of times per month and we have received a number of complaints from key stakeholders lamenting what happens when the leading zeros disappear.

Hope that helps.

jkassemi:
Thanks, @saintsoup52, that's super helpful. The "data dashboard" is the "Comparison_Tool_Data_Full" tab of what you download with the 
"Download Data on All Schools (Excel)" link on https://www.vets.gov/gi-bill-comparison-tool, right?

I think @jholton's PR looks good - this is the quoting style I see in the crosswalk file after this change:

screen shot 2018-08-13 at 11 12 28 am

and the following from the final export:

screen shot 2018-08-13 at 3 38 21 pm

saintsoup52:
@jkassemi - Yes, what I call the data dashboards is what users can download from the CT by clicking on "Download Data on All Schools (Excel)"

jkassemi (approved):

PR #366, opened 2019-02-11
Accreditation probation status update

jholton:
Per Patrick Campbell, removing "Loss of Accreditation or Preaccreditation: Voluntary Withdrawal" from list of probationary statuses.

Also removes a couple temporary tables from `db/schema` that were accidentally added.


lihanli (approved):
LGTM, were these tables added to production? you might need to make a migration to remove them

jholton:
> LGTM, were these tables added to production? you might need to make a migration to remove them

No, the code with that schema was never deployed to production.

PR #370, opened 2019-04-26
Change `from` address for Devise emails

jholton:
* Change `from` address for Devise emails
* Use latest GovDelivery config
* Update nokogiri, Devise, Rails to fix security bugs.


hpjaj (approved):
LGTM

PR #372, opened 2019-04-30
ENV variables won't be boolean because they are set via Python.

jholton:
re: [vets.gov-team#8123](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18123)


hpjaj (approved):
LGTM

PR #374, opened 2019-06-07
CT: 109 - Add "Approved" column to the Institution CSV generated by GIDS#18880

zurbergram:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18880
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18920


ksantiagoBAH (approved):
lgtm

devin-mccurdy (approved):
LGTM!

zurbergram:

db/migrate/20190607014201_add_column_approved_to_institutions.rb
```diff
@@ -0,0 +1,6 @@
+class AddColumnApprovedToInstitutions < ActiveRecord::Migration
+  def change
+    # Note this is going to result in null values, three-state-boolean problem
```
hpjaj: Regarding the [three-state-boolean](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem) problem, perhaps add in `null: false, default: false` to the `add_column` call?

zurbergram: This would set all previously approved institution versions in the institution table to false which is incorrect as currently any row in the institutions table are approved due to logic that is currently in https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L86 

The plan is that after these changes and the changes in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/375 are merged and deployed to production, in production a new version of institutions is created (Generate New Preview Version button below) and then published to production (Publish to production button below). 

Screen Shot 2019-06-10 at 3 20 48 PM

db/migrate/20190607014201_add_column_approved_to_institutions.rb
```diff
@@ -0,0 +1,6 @@
...
+    add_column :institutions, :approved, :boolean
```
jholton: Following the comments above, you can set the existing records to `true` when you run the migration:
```suggestion
    add_column :institutions, :approved, :boolean, null: false, default: false
    reversible do |dir|
      dir.up do
        Institution.update_all(approved: true)
      end
    end
```

zurbergram: Will do, thanks

db/migrate/20190607014201_add_column_approved_to_institutions.rb
```diff
@@ -0,0 +1,10 @@
+class AddColumnApprovedToInstitutions < ActiveRecord::Migration
+  def change
+    add_column :institutions, :approved, :boolean, null: false, default: false
+    reversible do |dir|
+      dir.up do
+        Institution.update_all(approved: true)
```
kfrz: This could end up in [broken migrations](https://github.com/rubocop-hq/rails-style-guide#define-model-class-migrations) if the model name changes. Something like:

```rb
# Define `table_name` in a custom named class to make sure that you run on the
# same table you had during the creation of the migration.
# In future if you override the `Product` class and change the `table_name`,
# it won't break the migration or cause serious data corruption.

class InstitutionModel < ActiveRecord::Base
  self.table_name = :institutions
end

# ...

  def change 
# ...
      dir.up do
        InstitutionModel.update_all(approved: true)
      end
# ...
```

cc: @hpjaj @jholton

db/migrate/20190607014201_add_column_approved_to_institutions.rb
```diff
@@ -0,0 +1,18 @@
+# Define `table_name` in a custom named class to make sure that you run on the
+# same table you had during the creation of the migration.
+# In future if you override the `Product` class and change the `table_name`,
```
jholton: Nitpick, so the comment is specific to this migration:
```suggestion
# In future if you override the `Institution` class and change the `table_name`,
```

db/migrate/20190607014201_add_column_approved_to_institutions.rb
hpjaj: @zurbergram - Is updating the data in a separate `rake` task an option?  I personally favor updating data in `rake` tasks over migrations, if possible.  That would mitigate against all concerns I believe.  You could run the migration as:

```ruby
def change
  add_column :institutions, :approved, :boolean, null: false, default: false
end
```

And then have a separate `rake` task that updates the records to `approved: true`, that you could scope to `Institution` records where the `created_at` is `<=` to the time of the migration (potentially as a passed value through `args`).

jholton: That would keep the migration clean.
This is essentially just a SQL UPDATE that is only pertinent to the existing database.  That is, it is inconsequential for anyone running all migrations on a new setup.
@hpjaj how would the rake task be executed?  If you have to ssh to the servers to do so anyway, and since this is just a one-time SQL UPDATE, you could consider it even cleaner to not write the rake task at all and just run `InstitutionModel.update_all(approved: true)` in a Rails console in each environment after deployment.
Thoughts?

hpjaj: Agreed!  I like both of those options (rake or live update).

jholton: @zurbergram, sorry for all the back-and-forth on this.  If you can remove the the code related to the data update from the migration, I can help you with getting the data updates done right after the deployments happen.

zurbergram: rake task added then removed once i saw this 😄

hpjaj (approved):
LGTM!

jholton:
@zurbergram 
I want to make sure I understand what is happening, and then raise a concern:
Before this PR, the only records in `institutions` were ones that were `approved` via WEAMS data.
Now institutions will be added to the `institutions` table regardless of the WEAMS status, but that WEAMS status will be reflected in the new `institutions.approved` column.
Is that correct?

If so, how does this affect the results from the API?
Should the `institutions_controller` be updated to only return results `where(approved: true)`?

jholton (changes requested):
@zurbergram 
I want to make sure I understand what is happening, and then raise a concern:
Before this PR, the only records in `institutions` were ones that were approved via WEAMS data.
Now institutions will be added to the `institutions` table regardless of the WEAMS status, but that WEAMS status will be reflected in the new `institutions.approved` column.
Is that correct?

If so, how does this affect the results from the API?
Should the `institutions_controller` be updated to only return results `where(approved: true)`?

zurbergram:
> @zurbergram
> I want to make sure I understand what is happening, and then raise a concern:
> Before this PR, the only records in `institutions` were ones that were `approved` via WEAMS data.
> Now institutions will be added to the `institutions` table regardless of the WEAMS status, but that WEAMS status will be reflected in the new `institutions.approved` column.
> Is that correct?
> 
> If so, how does this affect the results from the API?
> Should the `institutions_controller` be updated to only return results `where(approved: true)`?

https://github.com/department-of-veterans-affairs/gibct-data-service/pull/375 will add filtering to the queries in https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/controllers/v0/institutions_controller.rb to only find institutions rows where approved is true.

The institutions generated csv and database table will contain rows whose approved value is false (in addition to approved is true rows) but these rows will not be returned by the REST calls made to gibct-data-service

jholton:
> #375 will add filtering to the queries 

Got it.  Let's just make sure we get both merged before deploying to prod

jholton (approved):
LGTM

zurbergram:
These changes are in staging for edu approval before going into production

PR #375, opened 2019-06-11
Bah ksantiag 109

ksantiagoBAH:
## Description
As a Comparison Tool user, I only need to see institutions that are considered approved so that I can compare approved institutions at which I am able to use my benefits.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18882

## Testing done
I changed the values in the approved column in the database and ran the comparison tool. If the value in the approved column is true, auto-complete and information on facility appears. If false, auto-complete does not display school name and facilities information is not found. 

## Acceptance criteria

- [x] Institutions where "Approved" = false in the "Institution" CSV are not searchable in the Comparison Tool.
- [x] Institutions where "Approved" = false in the "Institution" CSV are not included in the Comparison Tool search results.

## Definition of done


shawkey-daniel (commented):
Add approvedInstitutions. Once PR #374 is merged, pull changes, then this will be approved.

zurbergram (approved):
LGTM

zurbergram:
These changes are in staging for edu approval before going into production

ksantiagoBAH:

app/controllers/v0/institutions_controller.rb
```diff
@@ -4,11 +4,16 @@ module V0
   # rubocop:disable Metrics/ClassLength
   class InstitutionsController < ApiController
     # GET /v0/institutions/autocomplete?term=harv
+
+    def approved_institutions
```
jholton: This should be a `private` method

spec/request/institutions_request_spec.rb
```diff
@@ -27,7 +27,8 @@
     end
 
     it 'allow searching by address fields' do
```
jholton: Can we get a spec showing that it does not return institutions where `approved: false`?

jholton (changes requested):

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -40,15 +40,15 @@
...
+      7.times { create(:institution, :contains_harv, approved: true) }
       get :autocomplete, term: 'harv', version: 'production'
       expect(response.content_type).to eq('application/json')
       expect(response).to match_response_schema('autocomplete')
     end
 
     it 'returns empty collection on missing term parameter' do
       create(:version, :production)
-      7.times { create(:institution, :contains_harv) }
+      7.times { create(:institution, :contains_harv, approved: true) }
```
kfrz: Does this need to be specifically seven times, and if so, please comment as to why so we can avoid [mystery guest smell](https://thoughtbot.com/blog/mystery-guest).

spec/request/institutions_request_spec.rb
kfrz: :heavy_plus_sign:

spec/controllers/v0/institutions_controller_spec.rb
kfrz: Also to reduce time it takes - we're hitting the DB 7 times to create 7 records, this is a lot of overhead for one example.

lihanli (approved):

app/controllers/v0/institutions_controller.rb
jholton: Anything that appears below `private` is a private method, so adding this here makes all of our public actions private, which we don't want.  You need to remove the `private` declaration above and just move the whole `approved_institutions` method below the `private` declaration lower in the class.

ksantiagoBAH:
Added the private method to the bottom.

spec/controllers/v0/institutions_controller_spec.rb
jholton: `approved` is a new column (via the most recent previous PR) that defaults to `false` at the database level.
these controller methods only return results where `approved: true`, so it needs to be specified here, or he can change the factory to specify `approved: true`.  I am unsure if that would affect other existing specs.

ksantiagoBAH:
I have remove the private access modifier as the linter does not like the redundant access modifier. I am receiving this message. 

Lint/UselessAccessModifier: Useless private access modifier.

spec/controllers/v0/institutions_controller_spec.rb
ksantiagomorgan6: Not sure why the 7.times is used there, I just added the approved: true portion to that line.

app/controllers/v0/institutions_controller.rb
```diff
@@ -147,6 +141,12 @@ def embed(group_counts)
         array << { name: k, count: v }
       end
     end
+    #private method added to account for added approved column
```
kfrz: Nitpicks:

Probably don't need to comment that this is a private method, it's implicit.

Also ensure there's a line of whitespace between `private` and the next line of code.

kfrz:
To be clear - you don't need to add an *additional* `private` access modifier - you simply need to move the `#approved_institutions` method below line 45 where the existing `private` keyword lives

app/controllers/v0/institutions_controller.rb
ksantiagomorgan6: Comment and extra private statement was removed.

ksantiagoBAH:
All requested changes have been made

spec/request/institutions_request_spec.rb
ksantiagomorgan6: Added the approved: false check to existing specs.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -6,7 +6,9 @@
   context 'version determination' do
     it 'uses a production version as a default' do
       create(:version, :production)
-      create(:institution, :contains_harv)
+      create(:institution, :contains_harv, approved: true)
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 10 and 11.  This spec is just testing that `it 'uses a production version as a default'` and having a non-approved row does not affect the test either way.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -15,8 +17,9 @@
 
     it 'accepts invalid version parameter and returns production data' do
       create(:version, :production)
-      create(:institution, :contains_harv)
-
+      create(:institution, :contains_harv, approved: true)
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 21 and 22 for the same reason as above.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -27,8 +30,9 @@
     it 'accepts version number as a version parameter and returns preview data' do
       create(:version, :production)
       v = create(:version, :preview)
-      create(:institution, :contains_harv, version: Version.current_preview.number)
-
+      create(:institution, :contains_harv, approved: true, version: Version.current_preview.number)
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 34 and 35 fore the same reason as above.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -40,15 +44,19 @@
   context 'autocomplete results' do
     it 'returns collection of matches' do
       create(:version, :production)
-      7.times { create(:institution, :contains_harv) }
+      7.times { create(:institution, :contains_harv, approved: true) }
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 48 and 49.  We are testing autocomplete of `approved` records and having this non-approved record does not affect it.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -40,15 +44,19 @@
...
       expect(response.content_type).to eq('application/json')
       expect(response).to match_response_schema('autocomplete')
     end
 
     it 'returns empty collection on missing term parameter' do
       create(:version, :production)
-      7.times { create(:institution, :contains_harv) }
+      7.times { create(:institution, :contains_harv, approved: true) }
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 58 and 59 for the same reason as above.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -40,15 +44,19 @@
   context 'autocomplete results' do
```
jholton: Within this `context 'autocomplete results'` it would be good to show that autocomplete does not return results of non-approved institutions, like so:
```suggestion
  context 'autocomplete results' do
    it 'does not return results for non-approved institutions' do
      create(:version, :production)
      create(:institution, :contains_harv, approved: false)
      get :autocomplete, term: 'harv', version: 'production'
      expect(JSON.parse(response.body)['data'].count).to eq(0)
      expect(response.content_type).to eq('application/json')
      expect(response).to match_response_schema('autocomplete')
  end
```

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -59,9 +67,11 @@
...
     before(:each) do
       create(:version, :production)
-      2.times { create(:institution, :in_nyc) }
-      create(:institution, :in_chicago, online_only: true)
-      create(:institution, :in_new_rochelle, distance_learning: true)
+      2.times { create(:institution, :in_nyc, approved: true) }
+      create(:institution, :in_chicago, online_only: true, approved: true)
+      create(:institution, :in_new_rochelle, distance_learning: true, approved: true)
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: This is good here because in the next spec (on line 79) it is shown that adding this did not affect the number of returned rows.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -185,8 +195,10 @@
   context 'category and type search results' do
     before(:each) do
       create(:version, :production)
-      create(:institution, :in_nyc)
-      create(:institution, :ca_employer)
+      create(:institution, :in_nyc, approved: true)
+      create(:institution, :ca_employer, approved: true)
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 200 and 201 because adding this record does not change testing for `category and type search results` either way.

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -224,8 +236,9 @@
     end
 
     it 'returns profile details' do
-      school = create(:institution, :in_chicago)
-
+      school = create(:institution, :in_chicago, approved: true)
+      # adding a non approved institutions row
+      create(:institution, :contains_harv, approved: false)
```
jholton: You can remove lines 240 and 241 because they do not affect the retrieving of the details of the school created on line 239.

jholton (changes requested):

spec/controllers/v0/institutions_controller_spec.rb
ksantiagomorgan6: Sure, both lines have been removed.

jholton (approved):
LGTM

PR #376, opened 2019-06-11
Generate ZipcodeRates from Weams data

devin-mccurdy:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18857


zurbergram (approved):
LGTM

devin-mccurdy:

app/models/institution_builder.rb
```diff
@@ -438,4 +439,36 @@ def self.add_school_closure(version_number)
 
     Institution.connection.update(str)
   end
+
+  def self.build_zip_code_rates_from_weams(version_number)
```
jholton: Can we get a spec for this in `institution_builder_spec.rb`?

sonntag-adam (approved):

app/models/institution_builder.rb
dmccurdybah: @jholton - spec'd!

app/controllers/v0/zipcode_rates_controller.rb
```diff
@@ -2,10 +2,11 @@
 
 module V0
   class ZipcodeRatesController < ApiController
-    skip_before_action :resolve_version
-    # GET /v0/housing_rates/20001
+    # GET /v0/zipcode_rates/20001
     def show
-      resource = ZipcodeRate.find_by(zip_code: params[:id])
+      resource = ZipcodeRate.version(@version[:number])
```
annaswims: It would be more performant if you wrote it as:  then you could use a composite index. 
`resource = ZipcodeRate.where(version: @version[:number], zip_code: params[:id]).order(:mha_rate).first`

db/migrate/20190626110700_add_columns_to_zipcode_rates.rb
```diff
@@ -0,0 +1,8 @@
+class AddColumnsToZipcodeRates < ActiveRecord::Migration
+  def change
+    add_column :zipcode_rates, :version, :integer
+
+    add_index :zipcode_rates, :zip_code, name: "index_zipcode_rates_on_zip_code", using: :btree
```
annaswims: I'd suggest making a composite index with 
`add_index :zipcode_rates, :[zip_code, :version] , using: :btree`

app/controllers/v0/zipcode_rates_controller.rb
annaswims: I take that back.  They both end up with the same sql, though I do think the suggested version is clearer to read.

dmccurdybah: Can do!

app/models/zipcode_rate.rb
```diff
@@ -4,10 +4,14 @@ class ZipcodeRate < ActiveRecord::Base
...
+    'zip_code' => { column: :zip_code, converter: ZipConverter },
     'mha_code' => { column: :mha_code, converter: BaseConverter },
     'mha_name' => { column: :mha_name, converter: BaseConverter },
     'mha_rate' => { column: :mha_rate, converter: NumberConverter },
     'mha_rate_grandfathered' => { column: :mha_rate_grandfathered, converter: NumberConverter }
   }.freeze
+
+  validates :version, presence: true
+
+  scope :version, ->(n) { where(version: n) }
```
annaswims: this scope is no longer used anywhere

db/migrate/20190626110700_add_columns_to_zipcode_rates.rb
```diff
@@ -0,0 +1,7 @@
+class AddColumnsToZipcodeRates < ActiveRecord::Migration
+  def change
+    add_column :zipcode_rates, :version, :integer
+
+    add_index :zipcode_rates, :[zip_code, :version] , using: :btree
```
annaswims: The colon should be inside the brackets.

annaswims (dismissed):

db/migrate/20190626110700_add_columns_to_zipcode_rates.rb
jholton: Make sure this runs locally.  Pretty sure there's a typo there.  Should be `[:zip_code, :version]`

dmccurdybah: Oof, thanks!

db/migrate/20190626110700_add_columns_to_zipcode_rates.rb
```diff
@@ -0,0 +1,7 @@
+class AddColumnsToZipcodeRates < ActiveRecord::Migration
+  def change
+    add_column :zipcode_rates, :version, :integer
+
+    add_index :zipcode_rates, [:zip_code, :version] , using: :btree
```
jholton: Can we swap the order of the columns in this index?  
```suggestion
    add_index :zipcode_rates, [:zip_code, :version] , using: :btree
    add_index :zipcode_rates, [:version, :zip_code] , using: :btree

```
I'm making an assumption that we'll only care about a `zip_code` for a given version.  That is, we're probably never going to look for records for a specific `zip_code` regardless of it's `version`, right?

Changing the order will improve efficiency when just filtering records on `version` too.

lib/csv_helper/exporter.rb
```diff
@@ -42,9 +44,20 @@ def set_class_for_export
...
-    def format(key, value)
-      return "\"#{value}\"" if key == :ope && value.present?
+    def format_value(record, field)
+      value = field['non_hash'] ? record.send(field['column']) : record[field['column']]
+      return "\"#{value}\"" if field['column'] == :ope && record[field['column']].present?
+      value
+    end
+
+    def format_institution_value(record, field)
+      value = false_to_nil(field['non_hash'] ? record.send(field['column']) : record[field['column']])
```
jholton: This `false_to_nil` conversion is unnecessary.
Both `false` and `nil` respond to `.present?` with `false`

db/migrate/20190626110700_add_columns_to_zipcode_rates.rb
dmccurdybah: That's correct.

Can do!

app/models/zipcode_rate.rb
```diff
@@ -4,10 +4,12 @@ class ZipcodeRate < ActiveRecord::Base
   include CsvHelper
 
   CSV_CONVERTER_INFO = {
-    'zip_code' => { column: :zip_code, converter: BaseConverter },
+    'zip_code' => { column: :zip_code, converter: ZipConverter },
```
jholton: I was looking back at the [history](https://github.com/department-of-veterans-affairs/gibct-data-service/commit/fe82564eef46f7320cfedcf3b85c75376820fb46) of this file.
It looks like this class has never been updated by a CSV uploaded via the dashboard (and thus has never been "exported" either).
As far as I can tell, the table has only been populated by a seed job.
That seed job utilized this `CSV_CONVERTER_INFO` hash.

If from now on this table will be populated by data in the WEAM spreadsheet, this can be an empty class:
```
class ZipcodeRate < ActiveRecord::Base
end
```
* we don't need to include `CsvHelper` since we're not loading or exporting.
* we don't need this `CSV_CONVERTER_INFO` hash at all for the same reason.
* the `validates :version, presence: true` is never used either since we're doing a direct SQL insert.

app/models/institution_builder.rb
```diff
@@ -448,4 +449,36 @@ def self.add_school_closure(version_number)
...
+    timestamp = Time.now.utc.to_s(:db)
+    conn = ActiveRecord::Base.connection
+
+    str = <<-SQL
+      INSERT INTO zipcode_rates (
+        zip_code,
+        mha_rate_grandfathered,
+        mha_rate,
```
jholton: What about the other `ZipcodeRate` attributes?
`mha_name`
`mha_code`
I was searching around the `vets-website` codebase, and I'm not sure that `mha_code` is used anywhere, but `mha_name` definitely _is_ used.

jholton (changes requested):

app/models/institution_builder.rb
dmccurdybah: They're not currently in the WEAMS data, and we have approval from the VA to remove `mha_name` from vets-website.

jholton: Are they going to be added to the WEAMS data?  If not, should we drop those columns from the table, and remove references to them from the codebase (e.g. `ZipcodeRateSerializer`)?

dmccurdybah: Unknown.  We're going to hold off making any changes until later because getting this in as-is is very high priority for them.

lib/csv_helper/exporter.rb
```diff
@@ -42,8 +44,15 @@ def set_class_for_export
       version.present? ? Institution.version(version.number) : Institution
     end
 
-    def format(key, value)
-      return "\"#{value}\"" if key == :ope && value.present?
+    def format_value(record, field)
+      value = field['non_hash'] ? record.send(field['column']) : record[field['column']]
```
jholton: It looks like all of these changes in this file revolve around getting the value of a method call into the `Institutions` export.
I propose a simpler solution:
Revert this entire file, and just change each usage of `record[k]` to `record.public_send(k)`.  I count 3 places.
And you can remove `non_hash: true` up in the `Institution` model.

It should be save to use `public_send` since we are only `send`ing methods hard coded in the models (that is, not `send`ing user input)

Let me know if I'm misunderstanding something.

jholton (changes requested):

lib/csv_helper/exporter.rb
dmccurdybah: No, that makes sense and works.  Thanks!

jholton (approved):
LGTM, with a caveat:
We will need to run a SQL update on `zipcode_rates` to populate the new `version` column for the existing data.  Otherwise `zipcode_rates_controller#show` will start throwing Exceptions for every request.

devin-mccurdy:
We're expecting that and are planning on generating a new preview version in staging.  I believe that we've set the expectation that this is something we will have to coordinate with the VA when we want this to be moved to production.

PR #382, opened 2019-06-25
Add support for child facilities

devin-mccurdy:
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19022


shawkey-daniel (approved):
LGTM

devin-mccurdy:

jholton:
I don't see any code that populates these new fields.  Is that intentional?
What is a facility child?  Is that, for example, various campuses of a university?

app/controllers/v0/institutions_controller.rb
```diff
@@ -40,6 +40,22 @@ def show
              meta: { version: @version }, links: @links
     end
 
+    # GET /v0/institituons/20005123/children
+    def children
+      children = Institution.version(@version[:number]).where(parent_facility_code_id: params[:id])
```
jholton: You have the query divided up between this line and line 53.
You can move it all here:
```suggestion
      children = Institution.version(@version[:number]).where(parent_facility_code_id: params[:id]).order(:institution).page(params[:page])
```

I believe the `children.count` will still return the full count;  the `page` method does not affect the count.

devin-mccurdy:
> I don't see any code that populates these new fields. Is that intentional?
> What is a facility child? Is that, for example, various campuses of a university?

Yes, an example would be extensions or other campuses.  

These are new columns in the weams CSV, which are pulled into the Instutitions table because they've been added to `COLS_USED_IN_INSTITUTION` in weam.rb

app/controllers/v0/institutions_controller.rb
jholton: Also, following @annaswims's comment [here](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/376/files#r298224965), it would be more performant if you wrote it as
```suggestion
      children = Institution.where(version: @version[:number], parent_facility_code_id: params[:id]).order(:institution).page(params[:page])
```
and added a composite index on `version` and `parent_facility_code_id`

config/routes.rb
```diff
@@ -27,6 +27,8 @@
   namespace :v0, defaults: { format: 'json' } do
     get '/calculator/constants' => 'calculator_constants#index'
 
+    get '/institutions/:id/children/' => 'institutions#children'
```
jholton: Can we move this into the `resources :institutions` block below?
`get :children, on: :member`

jholton (changes requested):

app/controllers/v0/institutions_controller.rb
```diff
@@ -40,6 +40,22 @@ def show
...
+
+      @meta = {
```
jholton: This `@meta` variable isn't used.  Should this be the value of line 55?

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -245,4 +245,30 @@
...
+      get :children, id: school.facility_code, version: school.version
+
+      expect(response.content_type).to eq('application/json')
+      expect(JSON.parse(response.body)['data'].count).to eq(1)
+      expect(JSON.parse(response.body)['data'][0]['attributes']['facility_code']).to eq(child_school.facility_code)
+      expect(JSON.parse(response.body)['data'][0]['attributes']['parent_facility_code_id']).to eq(school.facility_code)
+      expect(response).to match_response_schema('institutions')
+    end
+
+    it 'returns common exception response if child institutions not found' do
```
jholton: Is this wording inaccurate?  Does it return an exception response?  Or just an empty record set?

dmccurdybah: Thanks!, Yes this is inaccurate -- I forgot to update the description

app/controllers/v0/institutions_controller.rb
jholton: This one still needs addressing.

jholton (changes requested):

app/controllers/v0/institutions_controller.rb
dmccurdybah: Thanks!

jholton (changes requested):

db/migrate/20190701161400_add_parent_facility_code_index_to_intititutions.rb
```diff
@@ -0,0 +1,6 @@
+class AddParentFacilityCodeIndexToIntititutions < ActiveRecord::Migration
+  disable_ddl_transaction!
+  def change
+    add_index :institutions, [:version, :parent_facility_code_id] , using: :btree
```
jholton: This needs `algorithm: :concurrently` too:
```suggestion
    add_index :institutions, [:version, :parent_facility_code_id], using: :btree, algorithm: :concurrently
```

jholton (approved):
LGTM

delli-gatti-michael (approved):
Passes QA

devin-mccurdy:
DB migration has been split out into https://github.com/department-of-veterans-affairs/gibct-data-service/pull/407

kreek (approved):

jholton (dismissed):
Let's make sure we get the migration PR (#407) merged first.

PR #383, opened 2019-06-26
Update GIDS Database to archive historical versions of the software #19015

zurbergram:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19015

- requires `rake db:migration` in environments when deployed to them
- adding `gem 'config'`
    - `active_record.batch_size`
        defaults to 5000,  For use with https://api.rubyonrails.org/classes/ActiveRecord/Batches.html methods
    - `institutions.archive`
         Defaults to enabled,  Disables/enables the moving of rows from `institutions` to `institutions_archives`, defaults to disabled so that initial moving of rows can be done manually in production to reduce impact on production database, also so table can be added and then process turned on


shawkey-daniel (approved):
LGTM

zurbergram:

sonntag-adam (approved):

annaswims (approved):

app/controllers/dashboards_controller.rb
```diff
@@ -50,6 +50,8 @@ def push
         # Build Sitemap and notify search engines in production only
         ping = request.original_url.include?(GibctSiteMapper::PRODUCTION_HOST)
         GibctSiteMapper.new(ping: ping)
+        # Archive old institution rows
+        InstitutionsArchive.archive(pv)
```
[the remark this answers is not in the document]
zurbergram: yes

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def self.create_archives(version_number, conn)
+    str = <<-SQL
+      INSERT INTO institutions_archives
+        SELECT *
+          FROM institutions
+          WHERE version < ? ;
+    SQL
+
+    sql = InstitutionsArchive.send(:sanitize_sql, [str, version_number])
```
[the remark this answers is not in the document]
zurbergram: Newer version is https://apidock.com/rails/ActiveRecord/Sanitization/ClassMethods/sanitize_sql which is an alias for https://apidock.com/rails/ActiveRecord/Sanitization/ClassMethods/sanitize_sql_for_conditions

app/models/gibct_site_mapper.rb
```diff
@@ -28,11 +28,11 @@ def configure_sitemap(verbose)
     SitemapGenerator.verbose = verbose
   end
 
-  def generate_sitemap(v)
+  def generate_sitemap(version)
```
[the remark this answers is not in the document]
zurbergram: Changes came from new version of rubocop

Gemfile
```diff
@@ -65,7 +65,7 @@ group :development, :test do
   gem 'pry-nav'
 
   # Linters
-  gem 'rubocop', '~> 0.52.1', require: false
```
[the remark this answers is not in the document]
zurbergram: Upgrading removed deprecation warnings when running linter

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require_dependency 'institution'
+
+class InstitutionsArchive < Institution
+  self.table_name = 'institutions_archives'
+
+  # class methods
+  def self.archive(version)
```
annaswims: how long will this method take to run the first time it's run and what will be the results to the app while it's running.  It seems like inserting and deleting 5million rows will be problematic.

annaswims (dismissed):
I'd like to know more about the impact on the db of deleting and inserting 5m rows before I approve this pr.

db/migrate/20190708012000_create_institutions_archive.rb
```diff
@@ -0,0 +1,9 @@
+class CreateInstitutionsArchive < ActiveRecord::Migration
+  def up
+    execute "CREATE TABLE institutions_archives AS TABLE institutions WITH NO DATA;"
```
annaswims: What kind of indexes should this table have?

app/models/institutions_archive.rb
zurbergram: Screen Shot 2019-07-09 at 7 19 55 AM

From https://dsva.slack.com/archives/CBU0KDSB1/p1562596229148600

db/migrate/20190708012000_create_institutions_archive.rb
zurbergram: Changed sql to create table so that it inherits `institutions`'s defaults, constraints, and indexes

zurbergram:
> I'd like to know more about the impact on the db of deleting and inserting 5m rows before I approve this pr.

Requests were made at https://dsva.slack.com/archives/CBU0KDSB1/p1562596229148600 and https://dsva.slack.com/archives/CBU0KDSB1/p1562594109145700 to assess production database impact

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    conn.close
+  end
+
+  def self.create_archives(version_number, conn)
+    str = <<-SQL
+      INSERT INTO institutions_archives
+        SELECT *
+          FROM institutions
+          WHERE version < ? ;
```
jholton: I think we should make this `WHERE version = ?` to reduce the size of this operation.  Then we can control the execution of these statements through another mechanism (e.g. manually, or a loop)

app/models/institutions_archive.rb
annaswims: I'm not sure exactly when the explain was running, but around that time yesterday  the GIDS cpu was pegged at 100% Screen Shot 2019-07-09 at 10 55 21 AM

annaswims: and we were unable to fufill requests.  Not sure how much of that was due to a traffic spike or what.
image

zurbergram: I believe this issue was due to autocomplete causing issues, I think @jholton can speak to this

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,50 @@
...
+    SQL
+
+    sql = InstitutionsArchive.send(:sanitize_sql_for_conditions, [str, version_number])
```
jholton: `sanitize_sql_for_conditions` is `protected`, which means you can use it directly in an `ActiveRecord::Base` class.
So:
```suggestion
    sql = self.sanitize_sql_for_conditions([str, version_number])
```

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,50 @@
...
+    conn.execute(sql)
+  end
+
+  def self.cleanup_institutions(version_number, conn)
+    str = 'DELETE FROM institutions WHERE version < ? ;'
+
+    sql = Institution.send(:sanitize_sql_for_conditions, [str, version_number])
+    conn.execute(sql)
```
jholton: Let's use `ActiveRecord` where we can.
This whole method definition can be replace with
`Institution.where(version: version_number).delete_all`

Looking at this did prompt me to look at the `Institution` model and I noticed it has this association:
`has_many :yellow_ribbon_programs, dependent: :destroy`
Any thoughts on what we want to do with the associated `yellow_ribbon_programs` records?

jholton (changes requested):

app/controllers/dashboards_controller.rb
```diff
@@ -42,14 +42,19 @@ def push
...
+      if production_version.persisted?
         flash.notice = 'Production data updated'
 
         # Build Sitemap and notify search engines in production only
         ping = request.original_url.include?(GibctSiteMapper::PRODUCTION_HOST)
         GibctSiteMapper.new(ping: ping)
+
+        if ENV['ARCHIVE_INSTITUTIONS'] == 'True'
+          # Archive old institution rows
+          InstitutionsArchive.archive(production_version)
```
jholton: This doesn't have to be done immediately upon creating of a new production `version`.
What if we put it on a [scheduler](https://github.com/jmettraux/rufus-scheduler) to run in the middle of the night?
We could have a script that runs something like this at 4 AM (this assumes my other suggestion that `archive` only archives one version):
```
Version.production.where('number < ?', Version.current_production.number).each do |version|
  InstitutionsArchive.archive(version.number)
end
```
This could be added in a separate PR, after we get the archiving functionality in this PR merged and figure out the best approach to archiving the existing ~250 versions.

app/models/institutions_archive.rb
jholton: Yeah, my `EXPLAIN` queries should have put no load on the database.  It looks like our spike was due to `autocomplete` getting hammered.

db/schema.rb
```diff
@@ -303,8 +303,8 @@
     t.string   "physical_zip"
     t.string   "physical_country"
     t.integer  "dod_bah"
-    t.boolean  "approved",                                            default: false
-    t.boolean  "vet_tec_provider",                                    default: false
```
jholton: Hmm, why are these defaults removed?

app/helpers/application_helper.rb
```diff
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  # class methods
+  def self.batch_size
```
kfrz: This may be better suited to a `config/settings.yml` entry, and could benefit from a more specific name like `active_record_batch_size`, though I'm curious why we're redefining this at all

app/helpers/application_helper.rb
```diff
@@ -38,4 +46,6 @@ def pretty_error(errors, error_label = '')
       end
     end
   end
+
```
kfrz: Please remove excess whitespace

db/migrate/20190708012000_create_institutions_archive.rb
```diff
@@ -0,0 +1,17 @@
+class CreateInstitutionsArchive < ActiveRecord::Migration
+  def up
```
kfrz: Should we `disable_ddl_transaction!`?

spec/models/institutions_archive_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+          .with("There was an error occurring at the database level: #{error_message}")
+        InstitutionsArchive.archive(Version.current_production)
+      end
+    end
+  end
+
+  # private methods
+  private
+
+  def archive_test(initial_institution_count,
```
kfrz: This should be a [shared example](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) in my opinion.

spec/models/institutions_archive_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    expect(InstitutionsArchive.count).to eq(institutions_archive_count)
+    expect(InstitutionsArchive.where('version < ?', current_production_number).size).to eq(institutions_archive_count)
+  end
+
+  def create_production_version
+    create :version, :preview
+    create :version, :production, number: current_preview_number
+  end
+
+  def current_preview_number
```
kfrz: You can define these in a `let() {}` block, as is on lines 6-7. Like so:

```ruby
let(:current_preview_number) { Version.current_preview.number }
let(:current_production_number) { Version.current_production.number }
```

This allows for easy overriding if need be.

spec/models/institutions_archive_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    InstitutionsArchive.archive(Version.current_production)
+
+    expect(Institution.count).to eq(institution_count)
+    expect(Institution.where('version >= ?', current_production_number).size).to eq(institution_count)
+
+    expect(InstitutionsArchive.count).to eq(institutions_archive_count)
+    expect(InstitutionsArchive.where('version < ?', current_production_number).size).to eq(institutions_archive_count)
+  end
+
+  def create_production_version
```
kfrz: This can be done in a `before do ... end` block

spec/rails_helper.rb
```diff
@@ -54,7 +54,7 @@
   config.include FactoryGirl::Syntax::Methods
 
   # Allow skip_before_action in rspec controller tests
-  config.include Devise::TestHelpers, type: :controller
+  config.include Devise::Test::ControllerHelpers, type: :controller
```
kfrz: What is the impetus for this update?

config/application.yml.example
```diff
@@ -8,3 +8,5 @@ SAML_IDP_METADATA_FILE: /path/to/config/saml/metadata.xml
 SAML_CALLBACK_URL: http://localhost:3000/saml/auth/callback
 SAML_IDP_SSO_URL: https://example.com/idp/sso
 SAML_ISSUER: GIDS
+ARCHIVE_INSTITUTIONS: 'True'
+ACTIVE_RECORD_BATCH_SIZE: 1000
```
kfrz: This doesn't feel right to me, might we keep this batch size change in-line?

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+    rescue ActiveRecord::StatementInvalid => e
+      notice = 'There was an error occurring at the database level'
+      error_msg = e.original_exception.result.error_message
+      Rails.logger.error "#{notice}: #{error_msg}"
+    rescue StandardError => e
+      notice = 'There was an error of unexpected origin'
+      error_msg = e.message
+      Rails.logger.error "#{notice}: #{error_msg}"
+    end
```
kfrz: If this is something we will run often, we may consider adding further logging (i.e. to Sentry)

kfrz (changes requested):
Please extract rubocop upgrade and related variable changes to a separate PR in order to maintain git-log hygiene.

app/models/institutions_archive.rb
jholton: Just looking back over the past 7 months, it looks like new production versions are build 1 - 4 times/mo (which will be the impetus for this to run)

db/schema.rb
zurbergram: Fixed

spec/support/devise.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
-  config.include Devise::TestHelpers, type: :controller
+  config.include Devise::Test::ControllerHelpers, type: :controller
```
kfrz: Think this was updated in another PR?

zurbergram: in `spec/rails_helper.rb` it was,
I missed this change in this file https://github.com/department-of-veterans-affairs/gibct-data-service/pull/390/files

app/models/institutions_archive.rb
zurbergram: This would delete the current production version (i.e. the one being published) as `version_number` comes from https://github.com/department-of-veterans-affairs/gibct-data-service/blob/71a099c64eba46ecaea6682c51ae871f9364d0fc/app/controllers/dashboards_controller.rb#L45

app/models/institutions_archive.rb
zurbergram: For the `yellow_ribbon_programs` records, we can leave them in the table

app/models/institutions_archive.rb
zurbergram: Changing the https://github.com/department-of-veterans-affairs/gibct-data-service/blob/ed4e0bde150bd8fe8bbb94a9c7deec71c80c652f/app/models/institutions_archive.rb#L9 to no longer have a parameter and just do below for clarity

```
    production_version = Versions.current_production.number
```

app/helpers/application_helper.rb
zurbergram: there currently isn't a `config/settings.yml` so I would need to add gem 'config'

zurbergram: Should adding `gem 'config'` be a separate PR ?

config/application.yml.example
zurbergram: adding gem 'config'

app/models/institutions_archive.rb
zurbergram: Changed archiving process to now archive all versions that are less than newest production version and greater than or equal to previous production version

```
delete from institution 
    where version >= <previous_production_version> and version < <current_production_version>
```

app/helpers/application_helper.rb
zurbergram: Moved rubocop upgrade to https://github.com/department-of-veterans-affairs/gibct-data-service/pull/390
Adding `gem 'config'` to add `settings.yml`

app/models/institutions_archive.rb
zurbergram: Added sentry logging

spec/models/institutions_archive_spec.rb
zurbergram: At line 14 there is `before do ... end` block that calls this function twice, this is also used to create additional versions in individual tests for creating additional production versions

zurbergram:
@kfrz Rubocop upgrade has been moved to https://github.com/department-of-veterans-affairs/gibct-data-service/pull/390

@jholton @annaswims Changed archiving process to now archive all versions that are less than newest production version and greater than or equal to previous production version

```
delete from institution 
    where version >= <previous_production_version> and version < <current_production_version>
```

Can you review changes that have been made and see if reduction in number of rows archived along with other changes are sufficient?

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require_dependency 'institution'
+
+class InstitutionsArchive < Institution
+  self.table_name = 'institutions_archives'
+
+  # class methods
+  # rubocop:disable Metrics/MethodLength
+  def self.archive_previous_versions(user)
```
jholton: If I'm reading this right, this only archives and then deletes the most recent previous version.
How are you planning to archive all the existing previous versions in the table?

zurbergram: Correct, it will only archive previous production version and preview versions less than the newest production version.  This was done to reduce the database impact.  The plan would then be to manually archive all the older stuff so that you wouldn't have a 5 million insert/delete on the first run of the archive process.

shawkey-daniel: We have been working on this PR to get the table to hold only the latest version and move all the historical data to an archive table. We are open to suggestions for the best way to achieve this goal without causing any production performance issues. We can setup a meeting, if that would be beneficial.

jholton: No, I understand.  I suggested something similar previous.  I will suggest changing the name of the method to `self.archive_previous_version` (singular)

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,56 @@
...
+    production_version = Version.current_production.number
+    previous_version = Version.previous_production.number
+
+    begin
+      ActiveRecord::Base.transaction do
+        create_archives(previous_version, production_version)
+        Institution.where('version >= ? and version < ?', previous_version, production_version).delete_all
```
jholton: Since `production_version` == `Version.production.newest.first` and `previous_version` == `Version.production.newest.second`, then the above lines will only ever archive/delete `previous_version`.
Thus `create_archives` only needs to take `previous_version` as a parameter (and alter method definition as appropriate).  Same with the deletion:
```suggestion
        Institution.where('version = ?', previous_version).delete_all
```

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    rescue ActiveRecord::StatementInvalid => exception
+      notice = 'There was an error occurring at the database level'
+      error_msg = exception.original_exception.result.error_message
+      Raven.tags_context(
+        user: user,
+        date_time: Time.current,
```
jholton: I don't think we need these two tags above:
* Not sure what useful data `user` gets us that would help us troubleshoot an error.  Plus we should be specifying primitive values, not Ruby objects
* All the Sentry exceptions are already timestamped.

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    begin
+      ActiveRecord::Base.transaction do
+        create_archives(previous_version, production_version)
+        Institution.where('version >= ? and version < ?', previous_version, production_version).delete_all
+      end
+    rescue ActiveRecord::StatementInvalid => exception
+      notice = 'There was an error occurring at the database level'
+      error_msg = exception.original_exception.result.error_message
+      Raven.tags_context(
```
jholton: I think this should be `extra_context`.  `extra_context` is general details.  `tags_context` is for key/value pairs that Sentry will index/aggregate.  i.e. good for filtering.  I'm not sure we're ever going to need to (for example) see all errors where `production_version` is `n`.

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,56 @@
...
+    production_version = Version.current_production.number
+    previous_version = Version.previous_production.number
+
+    begin
+      ActiveRecord::Base.transaction do
+        create_archives(previous_version, production_version)
+        Institution.where('version >= ? and version < ?', previous_version, production_version).delete_all
+      end
+    rescue ActiveRecord::StatementInvalid => exception
+      notice = 'There was an error occurring at the database level'
```
jholton: Extract the body of this `rescue`and the one below into a method (since they are identical other than the `notice` value) and call it with the `notice`.
e.g. `process_exception('There was an error occurring at the database level')`

jholton (changes requested):

app/models/institutions_archive.rb
zurbergram: Made this change

app/models/institutions_archive.rb
zurbergram: removed

app/models/institutions_archive.rb
zurbergram: removed

app/controllers/dashboards_controller.rb
```diff
@@ -42,14 +42,19 @@ def push
...
+      if production_version.persisted?
         flash.notice = 'Production data updated'
 
         # Build Sitemap and notify search engines in production only
         ping = request.original_url.include?(GibctSiteMapper::PRODUCTION_HOST)
         GibctSiteMapper.new(ping: ping)
+
+        if Settings.institutions.archive
+          # Archive old institution rows
+          InstitutionsArchive.archive_previous_versions()
```
annaswims: nitpick, but no need for trailing `()`

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require_dependency 'institution'
+
+class InstitutionsArchive < Institution
+  self.table_name = 'institutions_archives'
```
annaswims: I'm guessing this line is unnecessary

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  # class methods
+  def self.archive_previous_versions
+    production_version = Version.current_production.number
+    previous_version = Version.previous_production.number
+
+    begin
+      ActiveRecord::Base.transaction do
+        create_archives(previous_version, production_version)
+        Institution.where('version >= ? and version < ?', previous_version, production_version).delete_all
```
annaswims: Is will this ever delete anything other than `previous_version`?
```suggestion
       `Institution.where(version: previous_version).delete_all`
```

app/models/institutions_archive.rb
```diff
@@ -0,0 +1,49 @@
...
+      process_exception(notice, error_msg, production_version, previous_version)
+    end
+  end
+
+  def self.create_archives(previous_version, production_version)
+    str = <<-SQL
+      INSERT INTO institutions_archives
+        SELECT *
+          FROM institutions
+          WHERE version >= ? and version < ? ;
```
annaswims: I'm confused about how this might archive anything other than `previous_version`

db/schema.rb
```diff
@@ -1307,6 +1443,8 @@
     t.integer  "dod_bah"
     t.boolean  "preferred_provider",                       default: false
     t.boolean  "stem_indicator",                           default: false
+    t.string   "campus_type"
```
annaswims: these changes don't seem relevant to this PR

spec/support/api/schemas/institutions.json
```diff
@@ -20,7 +20,9 @@
           }
         }
       },
-      "required": ["version"]
```
annaswims: do these changes have anything to do with archiving institutions?

app/models/institutions_archive.rb
zurbergram: Required as this class needs to override the value of `self.table_name` that it inherits from Institution

app/models/institutions_archive.rb
zurbergram: There can be multiple preview versions between production versions

app/models/institutions_archive.rb
annaswims: ... but now I understand

db/schema.rb
zurbergram: bad merge possibly, will cleanup

spec/support/api/schemas/institutions.json
zurbergram: bad merge possibly, will cleanup

spec/models/institutions_archive_spec.rb
zurbergram: This didn't work

app/controllers/dashboards_controller.rb
kfrz: yeah, rubocop should [catch this](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodCallWithoutArgsParentheses)

spec/support/api/schemas/institutions.json
kfrz: :heavy_plus_sign: to cleanup

kfrz (approved):

spec/support/api/schemas/institutions.json
zurbergram: looks like this is due to linter rules i have for json files in my IDE

app/controllers/dashboards_controller.rb
zurbergram: fixed

db/schema.rb
zurbergram: mixed database migrations on local, removed

annaswims (approved):
Thanks for all the changes.  

I recently learned that migrations are not run at deploy time for production. So, this code will fail until the migration is run.  Approving because this problem is already true in master and there will need to be a global solution.

jholton (approved):

PR #385, opened 2019-07-03
Remove new institutions columns defaults

devin-mccurdy:
Removed default values for new institution columns

### DIFF
--- a/db/migrate/20190607014201_add_column_approved_to_institutions.rb
+++ b/db/migrate/20190607014201_add_column_approved_to_institutions.rb
@@ -1,5 +1,5 @@
 class AddColumnApprovedToInstitutions < ActiveRecord::Migration
   def change
-    add_column :institutions, :approved, :boolean, null: false, default: false
+    add_column :institutions, :approved, :boolean, null: false
   end
 end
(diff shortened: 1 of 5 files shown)


zurbergram (approved):
LGTM

annaswims:
You're making changes to a rails migration that has been in master for almost 20 days.  How do we know which environments this change will apply to?

db/migrate/20190607014201_add_column_approved_to_institutions.rb
```diff
@@ -1,5 +1,5 @@
 class AddColumnApprovedToInstitutions < ActiveRecord::Migration
   def change
-    add_column :institutions, :approved, :boolean, null: false, default: false
```
hpjaj: @dmccurdybah -  Instead of making changes to an existing migration, perhaps create a new one and use the [`change_column`](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/change_column) method.  This will not impact existing data,  only data moving forward (post-migratation).  This should also help to mitigate against any migration edge cases.

Can you also pls explain the need for removing the defaults, or link to where this conversation is going on, for context?

cc: @jholton

dmccurdybah: This is because of a performance issue when running this migration in prod.  @jholton manually ran his own and will be opening a PR, so I'm closing this one.

PR #386, opened 2019-07-03
refactor migrations to add `approved` and `vet_tec_provider` to `institutions` table

jholton:
The previous version of this migration specified `null: false`, which meant that an update was run to populate all 6mm rows after creating the column.  Because of this, it locked up the table, and never successfully completed in production.

This way:
* creates both columns in bulk
* doesn't try to update all those roles.  

We can update the rows' values in batches post db-changes.

### DIFF
--- a/db/migrate/20190607014201_add_column_approved_to_institutions.rb
+++ b/db/migrate/20190607014201_add_column_approved_to_institutions.rb
@@ -1,5 +1,0 @@
-class AddColumnApprovedToInstitutions < ActiveRecord::Migration
-  def change
-    add_column :institutions, :approved, :boolean, null: false, default: false
-  end
-end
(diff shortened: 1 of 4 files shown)


devin-mccurdy (approved):
LGTM

db/migrate/20190607014201_add_column_approved_to_institutions.rb
```diff
@@ -1,5 +0,0 @@
-class AddColumnApprovedToInstitutions < ActiveRecord::Migration
```
hpjaj: @jholton - I'm sure you already considered this, but want to confirm.  Prior to this PR being merged, you will have ensured that the two deleted migrations are migrated `down` in all envs (i.e. dev/staging/prod).

jholton: Yup, done already.

PR #387, opened 2019-07-03
Add strong migrations

annaswims:
in order to protect gibct-data-service and vets-api we must ensure that we don't have migrations that lock the database. 

I verified by making a migration that would lock the db locally.
````
annacarey ~/code/gibct-data-service [add_strong_migrations] $ rails g migration foo
I, [2019-07-03T15:31:44.519714 #24376]  INFO -- sentry: ** [Raven] Raven 2.3.1 configured not to capture errors: DSN not set
      invoke  active_record
      create    db/migrate/20190703193144_foo.rb

annacarey ~/code/gibct-data-service [add_strong_migrations] $ be rake db:migrate
I, [2019-07-03T15:32:40.579700 #24456]  INFO -- sentry: ** [Raven] Raven 2.3.1 configured not to capture errors: DSN not set
== 20190703193144 Foo: migrating ==============================================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Passing a default value to change_column_null runs a single UPDATE query,
which can cause downtime. Instead, backfill the existing rows in the
Rails console or a separate migration with disable_ddl_transaction!.

class BackfillFoo < ActiveRecord::Migration
  disable_ddl_transaction!

  def change
    Instution.find_in_batches do |records|
      Instution.where(id: records.map(&:id)).update_all bad_idea: "default_value"
    end
  end
end

/Users/annacarey/code/gibct-data-service/db/migrate/20190703193144_foo.rb:3:in `change'
/Users/annacarey/.rbenv/versions/2.3.8/bin/bundle:22:in `load'
/Users/annacarey/.rbenv/versions/2.3.8/bin/bundle:22:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)

````


annaswims:
this was prompted by a migration in #374 that likely caused a lock.

jholton (approved):
Great idea.

PR #388, opened 2019-07-08
add institution index

annaswims:
we were seeing 100% cpu on the database today, mostly due to queries from autocomplete like 

`SELECT  id, facility_code as value, institution as label FROM "institutions" WHERE "institutions"."version" = $1 AND
"institutions"."approved" = $2 AND (lower(institution) LIKE ('university%')) LIMIT 6`

adding an index, on institution should help substantially. 

Unfortunately, schema.rb won't accurately reflect this index.  In the long term, the project could either move to [rails 5.0](https://blog.bigbinary.com/2016/07/20/rails-5-adds-support-for-expression-indexes-for-postgresql.html) or [structure.sql](https://guides.rubyonrails.org/v4.2/active_record_migrations.html#types-of-schema-dumps) to fix.

### DIFF
--- a/db/migrate/20190708211726_add_index_index_institutions_index_institution.rb
+++ b/db/migrate/20190708211726_add_index_index_institutions_index_institution.rb
@@ -3,6 +3,7 @@ class AddIndexIndexInstitutionsIndexInstitution < ActiveRecord::Migration
 
   def up
     safety_assured do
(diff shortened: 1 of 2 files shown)


annaswims:
Closing.  confirmed index_institutions_institution_lprefix is present in production and #383 will drastically reduce # of records in table

PR #392, opened 2019-07-15
Version Export and Generation Updates

devin-mccurdy:
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19235

Updated the version generation process in order to create the version outside of the transaction. This prevents multiple versions from being generated using the same number, which resulted in duplicate institutions.

Added export by version number to allow users to export the production and preview version.

Hid preview versions that are the same version number as the current production version.

Increased visibility of new production versions currently being generated.

image


zurbergram (approved):
LGTM

delli-gatti-michael (approved):
Testing passes

devin-mccurdy:

lib/csv_helper/exporter.rb
```diff
@@ -3,45 +3,54 @@
 module CsvHelper
   module Exporter
     def export
+      generate(csv_headers)
+    end
+
+    def export_version(number)
```
jholton: A bit of a nitpick, but can we rename this something like `export_institutions_by_version`, so that it's clear that `institutions` is the only model/csv that uses this?

jholton (changes requested):

app/models/weam.rb
```diff
@@ -165,12 +165,11 @@ def approved?
     return false unless poo_status =~ Regexp.new('aprvd', 'i')
 
     # VET TEC ONLY
-    return vet_tec_flags_for_approved? if applicable_law_code =~ Regexp.new("#{ALCVT}", 'i')
-    
+    return vet_tec_flags_for_approved? if applicable_law_code =~ Regexp.new(ALCVT.to_s, 'i')
```
jholton: Where'd this come from?  Can we remove this change if it's not related?
Also, `ALCVT` _is_ a `string`, right?  Since it already is a string, you shouldn't need to call `to_s` or use string interpolation on it.

app/models/institution_builder.rb
```diff
@@ -60,13 +60,15 @@ def self.run_insertions(version_number)
...
     error_msg = nil
+    version = Version.create!(production: false, user: user)
 
     begin
       Institution.transaction do
-        version = Version.create!(production: false, user: user)
         run_insertions(version.number)
       end
 
+      version.update(completed_at: Time.now.utc.to_s(:db))
```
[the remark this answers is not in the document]
dmccurdybah: image

It looks like the linter doesn't allow this

app/models/weam.rb
dmccurdybah: It looks like both of these are auto-corrected by Rubocop

image

app/models/institution_builder.rb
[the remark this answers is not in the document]
jholton: Interesting.  Today I learned....

app/models/weam.rb
jholton: I see.  We should have caught that when it was originally added.  
Anyway, I was saying since `ALCVT` is a string, you can just write:
```suggestion
    return vet_tec_flags_for_approved? if applicable_law_code =~ Regexp.new(ALCVT, 'i')
```
No `to_s` needed.  Rubocop will approve.

config/routes.rb
```diff
@@ -14,6 +14,7 @@
   get '/dashboards' => 'dashboards#index'
   post '/dashboards/build' => 'dashboards#build', as: :dashboard_build
   get '/dashboards/export/:csv_type' => 'dashboards#export', as: :dashboard_export, defaults: { format: 'csv' }
+  get '/dashboards/export/version/:number' => 'dashboards#export_version', as: :dashboard_export_version, defaults: { format: 'csv' }
```
jholton: What do you think of changing this to `get '/dashboards/export/institutions/:number'`?
Keeps it in line with the other `export` URL.

config/routes.rb
```diff
@@ -14,6 +14,7 @@
   get '/dashboards' => 'dashboards#index'
   post '/dashboards/build' => 'dashboards#build', as: :dashboard_build
   get '/dashboards/export/:csv_type' => 'dashboards#export', as: :dashboard_export, defaults: { format: 'csv' }
```
jholton: Given our new URL for `institutions`, I was wondering what would happen if you now visit `/dashboards/export/institutions`?  Will it generate an error?

I think you might need to alter `dashboards_controller#csv_model`.
Maybe that method should raise `ArgumentError` and specify in the message that you must provide a version number for `institutions`

config/routes.rb
dmccurdybah: Sounds good

jholton (changes requested):

config/routes.rb
dmccurdybah: There isn't any UI that would hit `/dashboards/export/institutions`, do you think we have to cover that case?

(but this is what you would get even if you tried to access it directly)

image

jholton: Ah, I see the old URL was `/dashboards/export/Institution`, not `/dashboards/export/institutions`.
What happens if you hit `/dashboards/export/Institution`?
It is a `GET` URL, so it's possible (if unlikely) that someone could type it in or have it in the browser history or have it bookmarked and visit it that way.

dmccurdybah: I removed the hardcoded rule that let Institution be exported without being in the allowed table list.

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -83,25 +83,48 @@ def load_table(klass, options)
 
     it 'causes a CSV to be exported' do
       expect(Weam).to receive(:export)
-      get :export, csv_type: 'Weam', format: :csv
+      get :export, csv_type: 'Weam', format: :csv, number: 1
```
jholton: These requests don't receive a `number` parameter, right?

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -83,25 +83,48 @@ def load_table(klass, options)
...
     end
 
     it 'includes filename parameter in content-disposition header' do
-      get :export, csv_type: 'Sva', format: :csv
+      get :export, csv_type: 'Sva', format: :csv, number: 1
```
jholton: Same as above: These requests don't receive a number parameter, right?

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -83,25 +83,48 @@ def load_table(klass, options)
...
+    before(:each) do
+      defaults = YAML.load_file(Rails.root.join('config', 'csv_file_defaults.yml'))
+
+      InstitutionBuilder::TABLES.each do |klass|
+        load_table(klass, skip_lines: defaults[klass.name]['skip_lines'])
+      end
+
+      get :build
+    end
+
```
jholton: Can we add a spec showing that an error is returned if you don't specify a `number` for an `institutions` export?

jholton (changes requested):

spec/controllers/dashboards_controller_spec.rb
dmccurdybah: `/dashboards/export/institutions/` would hit the route for the `export` method, which is already tested (it would be the same as someone exporting an invalid export type)

lib/csv_helper/exporter.rb
```diff
@@ -3,45 +3,54 @@
 module CsvHelper
   module Exporter
     def export
+      generate(csv_headers)
```
annaswims: I don't see relevant tests for this code, am I missing something?

annaswims (dismissed):
I don't see any specs for the code in lib/csv_helper/exporter.rb and I don't think we want to include css from cloudflare, due to our CSP policy. Also, from this point forward, migrations should be in a separate PR, merged and deployed before the code that depends on them.

lib/csv_helper/exporter.rb
dmccurdybah: I think all of the Exporter tests are through the dashboards_controller_spec

devin-mccurdy:
@jholton @annaswims I've split the DB and gemfile update out to separate PRs

https://github.com/department-of-veterans-affairs/gibct-data-service/pull/404
https://github.com/department-of-veterans-affairs/gibct-data-service/pull/406

devin-mccurdy:
@jholton @annaswims I've responded to all outstanding issues raised on here and the related PRs have been merged.  Can I get another look at this?

app/views/dashboards/_version.html.erb
```diff
@@ -4,19 +4,25 @@
...
-        <% end %>
+    <% if version.generating? %>
+        <i class="fa fa-gear fa-spin" style="font-size:24px"></i>
+        Reload page to refresh preview version status
+    <% else %>
+      <%= link_to 'Show GIBCT', version.gibct_link, target: "_blank",
+          class: "btn btn-success btn-xs", role: "button" %>
+        <%= link_to 'Download Export CSV', dashboard_export_version_path(version.number),
+            class: "btn btn-warning btn-xs", role: "button" %>
+      <% unless version.production? %>
```
jholton: Suggestion to make these branching `if` statements easier to follow:
Remove this `unless` and it's corresponding `end`.  Then change the `else` below to `elsif version.preview?` (see comment below)

app/views/dashboards/_version.html.erb
```diff
@@ -4,19 +4,25 @@
...
+        <% if version.publishable? %>
+          <%= link_to 'Publish to Production', dashboard_push_path, method: :post,
+              data: { confirm: "Are you sure you wish to publish preview version #{version.number} to production?" },
+              class: "btn btn-danger btn-xs",
+              role: "button",
+              id: "version_publish" %>
+          <% else %>
```
jholton:
```suggestion
          <% elsif version.preview? %>
```

app/views/dashboards/_version.html.erb
dmccurdybah: Thanks!

spec/controllers/dashboards_controller_spec.rb
dmccurdybah: I'm going to dismiss this because this is 10 days old and the institutions export route isn't accessible without a version number.

spec/models/version_spec.rb
```diff
@@ -150,7 +150,7 @@
         expect(subject.production?).to be_falsey
         expect(subject.preview?).to be_truthy
         expect(subject.latest_preview?).to be_truthy
-        expect(subject.publishable?).to be_truthy
+        expect(subject.publishable?).to be_falsey
```
jholton: Can we fix up these `version` specs?  
This is testing if a `Version.current_preview.publishable?` should be truthy.  It _should_ be truthy, right?
I think you need to fix this in the `versions` factory by setting a value for `completed_at`.

Along those lines, maybe you need another scope to make return records where `completed_at` is not `nil`  to be applied where appropriate.  First place I see is `Version.current_preview`.  Perhaps the definition of that method should be `Version.preview.completed.newest.first` (assuming you name the scope `completed`).  Also use it anywhere else you don't want a non-completed version to be returned.

dmccurdybah: We do want `current_preview` to return the most recent preview version regardless of whether or not is has been completed, so having `Version.current_preview.publishable?` be false is correct here.

I can add an additional test to cover the case where the current_preview version is completed.  Does that work?

jholton: I see.  Yes please to covering when `completed_at` is not `nil`.

jholton (approved):

PR #396, opened 2019-07-17
use safety_assured and change_table for CT 116 Preferred Provider

zurbergram:
Previous version of scripts produced in a staging environment

```
standardError: An error has occurred, this and all later migrations canceled:

Adding a column with a non-null default causes the entire table to be rewritten.
Instead, add the column without a default value, then change the default.

class AddPreferredProviderColumnToInstitutions < ActiveRecord::Migration
  def up
    add_column :institutions, :preferred_provider, :boolean
    change_column_default :institutions, :preferred_provider, false
  end

  def down
    remove_column :institutions, :preferred_provider
  end
end

Then backfill the existing rows in the Rails console or a separate migration with disable_ddl_transaction!.

class BackfillAddPreferredProviderColumnToInstitutions < ActiveRecord::Migration
  disable_ddl_transaction!

  def change
    Institution.find_in_batches do |records|
      Institution.where(id: records.map(&:id)).update_all preferred_provider: false
    end
  end
end
```

### DIFF
--- a/db/migrate/20190712019179_add_preferred_provider_column_to_institutions.rb
+++ b/db/migrate/20190712019179_add_preferred_provider_column_to_institutions.rb
@@ -1,5 +1,15 @@
 class AddPreferredProviderColumnToInstitutions < ActiveRecord::Migration
-  def change
-    add_column :institutions, :preferred_provider, :boolean, default: false
+  def up
+    safety_assured do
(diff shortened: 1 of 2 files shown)


zurbergram:

jholton (approved):
LGTM

PR #407, opened 2019-08-05
DB migration for: Add support for child facilities

devin-mccurdy:
Migration PR for https://github.com/department-of-veterans-affairs/gibct-data-service/pull/382


zurbergram (approved):
LGTM

jholton (approved):
LGTM

db/migrate/20190808112200_add_parent_facility_code_index_to_intititutions.rb
```diff
@@ -0,0 +1,6 @@
+class AddParentFacilityCodeIndexToIntititutions < ActiveRecord::Migration
+  disable_ddl_transaction!
+  def change
+    add_index :institutions, [:version, :parent_facility_code_id] , using: :btree, algorithm: :concurrently
```
jholton: If `institutions_archives` is supposed to be virtually identical to `institutions`, let's add the same index for it here.

jholton (changes requested):

db/migrate/20190808112200_add_parent_facility_code_index_to_intititutions.rb
dmccurdybah: There won't ever be queries ran against the institutions_archives table that would make use of this index. I know we copied indexes when creating the table, but this one won't be used.  I can definitely add it if you'd like us to keep that up.

jholton: Hmmm.  If it's `write` only, then that makes sense.  In fact, should we drop some/all of the other indexes so that writing to the table is faster?  Indexes slow down writes.

zurbergram: If we want to improve the performance of the archiving we can create a story to do that. The archive capability is now on staging so testing against it can happen.  The process will run when a user publishes a version to production.

jholton: Sounds good.  Something to keep in mind if performance needs improving.

jholton (approved):
LGTM

PR #411, opened 2019-08-12
Institution facility map

devin-mccurdy:
Add a "facility map" of related institutions to the institution profile.

There are 3 valid campus_type values:

- `Y` - Main campus
- `N` - Branch
- `E` - Extension

A main campus can have multiple branches or extensions, and branches can only have extensions.  The data provided from WEAMS will follow these rules.

The facility map has been added to the institution profile because it will be used on the vettec and non-vettec versions of institution profile page in the comparison tool.  The entire tree is returned for any institution (main campus, branch, or extension)

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/18827


shawkey-daniel (approved):
LGTM

zurbergram (approved):
LGTM

sonntag-adam (approved):

jholton:
Do we have a sample `weam.csv` document that I can use to populate my database locally that will work for this?

app/models/institution.rb
```diff
@@ -190,6 +190,10 @@ def complaints
     end
   end
 
+  def facility_map
+    InstitutionTree.build(facility_code, version)
```
jholton: Here you pass in the `facility_code` and `version` attributes of an institution, and then immediately retrieve the institution using those attributes in the `build` method.  How about just passing the institution instead?
```suggestion
    InstitutionTree.build(self)
```

app/models/institution_tree.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+module InstitutionTree
+  def self.build(facility_code, version)
```
jholton: following from my comment above, this would receive the institution:
```suggestion
  def self.build(institution)
    main = main_campus(institution)
```

app/models/institution_tree.rb
```diff
@@ -0,0 +1,86 @@
...
+    def build_extensions(institutions, branch_facility_code)
+      extensions = []
+      institutions.each do |institution|
+        next unless institution['parent_facility_code_id'] == branch_facility_code && institution['campus_type'] == 'E'
+        extensions << institution
+      end
+      extensions.sort_by { |extension| extension['institution'] }
+    end
+
+    def main_campus(facility_code, version)
```
jholton: And down here you no longer have to look up the institution:
```suggestion
    def main_campus(institution)
```

app/models/institution_tree.rb
```diff
@@ -0,0 +1,86 @@
...
+          UNION
+            SELECT i.facility_code
+            FROM institutions i
+            INNER JOIN related_down r ON i.parent_facility_code_id = r.facility_code
+            WHERE i.version = ?
+        ) SELECT facility_code FROM related_down WHERE facility_code != ?
+      SQL
+
+      sql = Institution.send(:sanitize_sql, [str, facility_code, version, version, facility_code])
+      facility_codes = Institution.connection.execute(sql).map do |institution|
```
jholton: You can clean this up like so:
```suggestion
      facility_codes = Institution.connection.execute(sql).field_values('facility_code')
```

app/models/institution_tree.rb
```diff
@@ -0,0 +1,84 @@
...
+        'branches' => build_branches(all),
+        'extensions' => build_extensions(all, main.facility_code)
+      }
+    }
+  end
+
+  class << self
+    private
+
+    def build_branches(institutions)
```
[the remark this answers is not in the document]
dmccurdybah: Cool, thanks!

app/models/institution_tree.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+module InstitutionTree
+  def self.build(institution)
+    main = main_campus(institution)
+    all = descendants(main)
+
+    {
+      'main' => {
```
jholton: I'm curious, why have this 'main' top level key?
Why not just return:
```
      {
        'institution' => main,
        'branches' => build_branches(all),
        'extensions' => build_extensions(all, main.facility_code)
      }
```

dmccurdybah: I agree, but the front-end work is already done and we are going to need to demo this ASAP.

jholton (approved):
LGTM.
I was curious about the speed of the "WITH RECURSIVE" Queries creating temporary tables.
After digging a bit, it looks like the "WITH RECURSIVE" queries are surprisingly blazing fast.  We should keep an eye on DB performance after deployment though.

I was also worried about the possibility of bad data resulting in an infinite recursion.
For the both queries, it shouldn't be possible since you used "UNION".  "UNION ALL" would have allowed it in the case where two rows erroneously referenced each other as parents (that should never happen, but you should not trust input data to be correct all the time).

Should `facility_code` be unique (per `version`)?  If so, we should look into adding an index that enforces that.  There isn't one right now.

We do have a [Rails validation](https://github.com/department-of-veterans-affairs/gibct-data-service/blob/3a3b8091d4510c2bd82833d7238d5f148651d916/app/models/institution.rb#L157):
` validates :facility_code, uniqueness: true, presence: true`, but this isn't ever executed because we do our inserting with raw SQL.  In fact, the validation as it is written doesn't work.  It doesn't scope for `version`.  It should be:
`validates :facility_code, uniqueness: { scope: :version }, presence: true`
Assuming you have more than one `version` in your `institutions` table, if you fire up a Rails console, retrieve a random `Institution` record and try to save it back, it fails because of this validation (i.e. you have more than one record in the database with the same `facility_code`).

PR #414, opened 2019-08-14
Remove extension institutions from search results

devin-mccurdy:
Remove extension institutions from search results.

```As a comparison tool user, I should not be shown extensions in the search results, because they are not considered separate institutions.```

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19462

### DIFF
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -106,9 +106,7 @@ module V0
       ].each do |filter_args|
         filter_args << filter_args[0] if filter_args.size == 1
         relation = relation.filter(filter_args[0], @query[filter_args[1]])
-
-        # relation = relation.select { |institution| institution.campus_type != 'E' }
-      end
+    end
 
       relation
     end
--- a/app/controllers/v0/institutions_controller.rb
+++ b/app/controllers/v0/institutions_controller.rb
@@ -106,6 +106,8 @@ module V0
       ].each do |filter_args|
         filter_args << filter_args[0] if filter_args.size == 1
         relation = relation.filter(filter_args[0], @query[filter_args[1]])
+
+        # relation = relation.select { |institution| institution.campus_type != 'E' }
       end
 
       relation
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by zurbergram — APPROVED** [review]
LGTM

**Review by shawkey-daniel — APPROVED** [review]
LGTM

**Review by delli-gatti-michael — APPROVED** [review]
Approved by QA

File: app/controllers/v0/institutions_controller.rb [hunk 1]
```diff
@@ -170,7 +170,7 @@ def embed(group_counts)
     end
 
     def approved_institutions
-      Institution.version(@version[:number]).where(approved: true)
+      Institution.version(@version[:number]).where(approved: true).where("campus_type != 'E' OR campus_type IS NULL")
```
kfrz: It might make sense to extract this to a named scope in the model definition, though that's a nitpick.

kfrz (approved):
LGTM overall!

jholton (changes requested):
I have some concerns.  Still working on my review.

app/controllers/v0/institutions_controller.rb
jholton: To be clear, you want to return Institution records that have any campus_type value other than 'E', including `null`.  Is that correct?

dmccurdybah: Correct

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -70,6 +70,15 @@
       expect(response.content_type).to eq('application/json')
       expect(response).to match_response_schema('autocomplete')
     end
+
+    it 'does not return results for extension institutions' do
+      create(:version, :production)
+      create(:institution, :contains_harv, campus_type: 'E')
```
jholton: `autocomplete` only matches text that _start_ with the search term.  This test is passing because the autocomplete search is failing to find a match, and not necessarily because `campus_type` is `E`.
You can change it like so:
```suggestion
      create(:institution, :start_like_harv, campus_type: 'E')
      create(:institution, :start_like_harv, campus_type: 'Y')
```
And in the expectation below:
`expect(JSON.parse(response.body)['data'].count).to eq(1)`
Since both institution records are identical other than their `campus_type` value, this will properly show that the reason only one result was returned due to `campus_type` value.

jholton (approved):
LGTM

PR #417, opened 2019-08-15
Have institution factory default `approved` attribute to true

jholton:
In order to make our specs less brittle, I'm giving the `approved` attribute a default value of true since it's a necessary value for results to be returned by search.
It's too easy to forget to set it every time when writing specs.


kfrz (approved):
Net negative diff. Rad! :ship:

zurbergram (approved):
LGTM

PR #420, opened 2019-08-19
Cleanup of vet tec institution 'approved' logic

devin-mccurdy:
Cleanup of vet tec weams institution approved logic to resolve an issue where vet tec institutions without valid `applicable_law_code` values were incorrectly being set as approved.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19519

### DIFF
--- a/app/models/weam.rb
+++ b/app/models/weam.rb
@@ -185,7 +185,7 @@ class Weam < ActiveRecord::Base
   end
 
   def vet_tec_approved?
-    applicable_law_code =~ Regexp.new(ALCVT, 'i')
+    applicable_law_code.downcase.include? ALCVT
   end
 end
 
--- a/app/models/weam.rb
+++ b/app/models/weam.rb
@@ -10,10 +10,6 @@
 class Weam < ActiveRecord::Base
   include CsvHelper
 
-  ALC1 = 'educational institution is not approved'
-  ALC2 = 'educational institution is approved for chapter 31 only'
-  ALCVT = 'vet tec only'
-
   COLS_USED_IN_INSTITUTION = %i[
     facility_code institution city state zip
     address_1 address_2 address_3
@@ -151,10 +147,6 @@ class Weam < ActiveRecord::Base
       flight_indicator || non_college_degree_indicator
   end
 
-  def vet_tec_flags_for_approved?
-    non_college_degree_indicator
-  end
-
   # To be approved, a school must be marked 'aprvd' in the poo-status, have
   # an approved applicable law code that is not restrictive of GI Bill
   # benefits, and be a higher learning institution, OJT, flight,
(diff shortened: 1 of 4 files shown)


zurbergram (approved):
LGTM

delli-gatti-michael (approved):
Passes QA Testing

app/models/weam.rb
```diff
@@ -163,16 +155,38 @@ def vet_tec_flags_for_approved?
...
-    # Other
-    return false if applicable_law_code =~ Regexp.new("#{ALC1}|#{ALC2}", 'i')
     flags_for_approved?
   end
+
+  private
+
+  ALC1 = 'educational institution is not approved'
+  ALC2 = 'educational institution is approved for chapter 31 only'
+  ALCVT = 'vet tec only'
```
jholton: Just a style nitpick, but can we move these back to the top of the file with the other constant(s)?
That's where you can generally expect to find constants, ala [this style guide](https://github.com/rubocop-hq/ruby-style-guide#consistent-classes).
Moving them down here does not make the constants private.
If you really do want to make the constants private, you can use [private_constant](http://www.chrisrolle.com/en/blog/private-constants-in-ruby)

app/models/weam.rb
```diff
@@ -163,16 +162,34 @@ def vet_tec_flags_for_approved?
...
   end
+
+  private
+
+  def poo_status_valid?
+    poo_status =~ Regexp.new('aprvd', 'i')
+  end
+
+  def invalid_law_code?
+    LAW_CODES_BLOCKING_APPROVED_STATUS.none? { |law_code| applicable_law_code.downcase.include? law_code }
```
jholton: It's possible I'm reading this wrong, but I think this should be `any?`, the opposite of `none?`:
```suggestion
    LAW_CODES_BLOCKING_APPROVED_STATUS.any? { |law_code| applicable_law_code.downcase.include? law_code }
```
And since specs are passing you should probably revisit the `approved?` specs to make sure you're covering all cases.

In the spec file, lines [219 and 220](https://github.com/department-of-veterans-affairs/gibct-data-service/blob/c55f1d63a8b769681404fcf244063dee306b9289/spec/models/weam_spec.rb#L220) are building instances with the exact same attributes.  You can remove one of those.

`invalid_law_code?` evaluates to true with this definition.
Your spec is still passing because `flags_for_approved?` evaluates to `false`.  
Make sure you have a spec where `flags_for_approved?` will evaluate to `true` to be sure that `invalid_law_code?` is working.
e.g. change line 219 from `let(:non_approved_law_code) { build :weam, :approved_poo_and_non_approved_law_code }` to `let(:non_approved_law_code) { build :weam, :approved_poo_and_non_approved_law_code, ojt_indicator: true }` and see what happens.

dmccurdybah: Good catch, thanks!

jholton (approved):
LGTM!

PR #421, opened 2019-08-20
CT 116: Make SCO & Program files not required to generate a new GIDS preview version #19526

zurbergram:
## Description
Adding a distinction between required and not required csvs along with visual changes to indicate which is which

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19526

To test will need to do below in psql or pgadmin if have already uploaded either file

```
delete from uploads where csv_type in ('Program','SchoolCertifyingOfficial')
```

## Acceptance Criteria
- [ ] should be able to "Generate a new preview" version without having an initial upload of `Program` or `SchoolCertifyingOfficial`

## Screenshots
Screen Shot 2019-08-20 at 2 32 19 PM


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):
Approved by QA

jholton:
Good refactor, but I think we have some code organization "best practices" issues.  I'll leave comments in-line.

app/helpers/csv_types.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
jholton: files in the `helpers` directory are intended to be helpers for views.  Everything here is static data that is used throughout the codebase.  These values should work fine as global constants.
You can create a file in the `initializers` directory and define them there.
Something like
```
CSV_TYPES_TALBES = [...] # as defined below
CSV_TYPES_REQUIRED_TABLE_NAMES = CSV_TYPES_TALBES.select { |table| table[:required?] }.map { |table| table[:klass].name }
CSV_TYPES_ALL_TABLES = CSV_TYPES_TALBES.map { |table| table[:klass] }
```
@annaswims any thoughts on the best place for this?

app/models/upload.rb
```diff
@@ -23,12 +23,24 @@ def ok?
     ok
   end
 
+  def required
```
jholton: These two methods are basically view data.  We should try to keep our views/presentation logic separate from our models.  These are prime candidates for helper methods.

You can create a `helpers/dashboards_helper.rb` and define these methods there (the methods would receive the `upload` object as a parameter)

app/helpers/csv_types.rb
zurbergram: I just went ahead and did this

app/models/upload.rb
zurbergram: Moved and rspeced

app/helpers/dashboards_helper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module DashboardsHelper
+  def latest_upload_class(upload)
+    return '' if upload.ok?
+    return 'danger' if CSV_TYPES_REQUIRED_TABLE_NAMES.include?(upload.csv_type)
+    return 'warning' unless CSV_TYPES_REQUIRED_TABLE_NAMES.include?(upload.csv_type)
```
jholton: The `unless` statement is not necessary.  If we make it to this line `CSV_TYPES_REQUIRED_TABLE_NAMES` does not include `upload.csv_type` since we checked for it's presence in the previous line.
```suggestion
    return 'warning'
```

app/helpers/dashboards_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def latest_upload_title(upload)
+    return '' if upload.ok?
+    return 'Missing required upload' if CSV_TYPES_REQUIRED_TABLE_NAMES.include?(upload.csv_type)
+    return 'Missing upload' unless CSV_TYPES_REQUIRED_TABLE_NAMES.include?(upload.csv_type)
```
jholton: Same as above:
```suggestion
    return 'Missing upload'
```

app/helpers/dashboards_helper.rb
zurbergram: changed

app/helpers/dashboards_helper.rb
zurbergram: changed

jholton (approved):
LGTM

PR #427, opened 2019-08-26
Put back sass

zurbergram:
http://sentry.vfs.va.gov/vets-gov/gi-bill-data-service-staging/issues/71569/?query=is%3Aunresolved


jholton (approved):
LGTM

PR #426, opened 2019-08-26
gibct-data-service change from FactoryGirl to FactoryBot #1310

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/1310

Adding back rainbow gem because of http://sentry.vfs.va.gov/share/issue/caaf644a525d45e5933c9d0fe521ff2c/

PR #429, opened 2019-08-27
Develop process to do a production dependency  build for testing #19575

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19575

As a developer and QA, I need a way to do a production dependency build so I can test as if on the staging or production environments.

Adding in `gem 'factory_bot_rails', '4.10.0'` because of error below after doing `bundle install`

```
~/.rbenv/versions/2.3.8/lib/ruby/gems/2.3.0/gems/factory_bot-5.0.2/lib/factory_bot/definition_proxy.rb:97:in `method_missing': undefined method 'dapip_id' in 'accreditation_action' factory (NoMethodError)
```

This error does not occur when doing `bundle install --without development test`


shawkey-daniel (approved):
LGTM

Gemfile
```diff
@@ -84,6 +89,8 @@ group :development, :test do
   gem 'guard-rspec', require: false
 
   gem 'factory_bot_rails'
+  # TODO: switch to a version number once that version is released
```
annaswims: I don't understand why any of the changes in the Gemfile were made.

zurbergram: Currently in master if you locally run `bundle exec rspec` you will get multiple

```
RuntimeError:
  can't modify frozen Array
# ./config/environment.rb:5:in `<top (required)>'
# ./spec/rails_helper.rb:5:in `<top (required)>'
# ./spec/serializers/institution_serializer_spec.rb:3:in `<top (required)>'
```

Additionally this is the same dependency pulled in the same way as in https://github.com/department-of-veterans-affairs/vets-api/blob/master/Gemfile#L141

In https://github.com/department-of-veterans-affairs/gibct-data-service/pull/426 it was recommended to remove this line, removal didn't affect build passing but it affects local rspec

zurbergram: `sass-rails` and `rainbow` are only needed as gems when the development and test groups are not installed

annaswims: I this case I don't think you want to copy-paste from vets-api.  The comment `switch to a version number once that version is released` is out of place, considering the version containing this commit was released 2 years ago.

Gemfile
```diff
@@ -54,14 +53,20 @@ gem 'jquery-ui-rails'
...
 
-gem 'rainbow'
 
 # Use ActiveModel has_secure_password
 gem 'ruby-saml'
 
 gem 'strong_migrations'
 gem 'will_paginate'
 
+group :production do
```
annaswims: These gems should likely be available in environments other than production

README.md
```diff
@@ -82,6 +82,15 @@ User.create(email: 'xxxxxx', password: 'xxxxxx')
 5. Load test users and sample data: `bundle exec rake db:seed`
 5. Start the application: `bundle exec rails s`
 
+### Local Production Dependency Build
+The steps below will setup your local without dependencies that are not loaded in dev, staging, and production environments.
+1. `bundle install --without development test`
+2. `bundle clean --force`
+3. `bundle exec rails s -p 4000`
```
annaswims: I think what you want is just 1 step - `bundle exec  rails server -e production` is there a reason you want it to run on port 4000?  The only change you'd need to make is to make sure that you're not actually connection to the production database in database.yml

Gemfile
zurbergram: They are pulled in via `rubocop` when it is installed

README.md
zurbergram: port 4000 is just how I run it, removing the option

zurbergram: So I tried running with `bundle exec rails server -e production` without the gems `rainbow` or `sass-rails` being installed or listed in Gemfile and the UI portion of GIDS was still functioning normally because they are pulled in by `rubocop`, I had not done `bundle install --without development test; bundle clean --force;` just `bundle install; bundle clean --force;`.

The original issue that this PR is trying to prevent again was that gems `rainbow` and `sass-rails` were removed which resulted in css not displaying on dev/staging.

In addition to the changes for `database.yml`, [config.assets.compile](https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/config/environments/production.rb#L32) has to be set to true otherwise you will receive an error message similar to what is below

```
Called from ~/.rbenv/versions/2.3.8/lib/ruby/gems/2.3.0/gems/activesupport-4.2.11.1/lib/active_support/dependencies.rb:240:in `load_dependency'
F, [2019-08-27T13:50:23.154067 #60897] FATAL -- : [5e7d682f-e8c3-4426-bb08-a65c8894adea] [localhost]
ActionController::RoutingError (No route matches [GET] "/assets/application-acec3ea9bc666e8034fe06c893b1ed371b8185f4d95a8130ccaf86247d3ce043.js"):
```

Because changes are required to 2 files that are included in source control I went with this process instead as it does not affect any files in source control and tests what I was trying to test.

I titled the section "Local Production Dependency Build" to indicate this is specifically for running with production dependencies i.e. doing `bundle install --without development test;`

Gemfile
zurbergram: So because of https://thoughtbot.com/blog/deprecating-static-attributes-in-factory_bot-4-11 the version will be set to `4.10.0` which is the version listed on the previously listed commit hash https://github.com/thoughtbot/factory_bot/commit/50eeb67241ea78a6b138eea694a2a25413052f49

An attempt to fix the issue referenced was not going to be simple, another story will be needed to update factory_bot

Gemfile
```diff
@@ -54,14 +53,20 @@ gem 'jquery-ui-rails'
...
+  # Use SCSS for stylesheets
+  gem 'sass-rails', '~> 5.0'
+
+  gem 'rainbow'
```
jholton: Since `rainbow` is used explicitly in the `lint` and `security` rake tasks (which are run as part of CI), let's move this to the global namespace/group.

I know it gets included by `rubocop` but we should still explicitly list it for all environments.
Removing `rubocop` should not break those rake tasks, which would happen without `rainbows` included  globally (not that we plan to ever remove `rubocop`)

zurbergram: Moved

README.md
```diff
@@ -82,6 +82,15 @@ User.create(email: 'xxxxxx', password: 'xxxxxx')
 5. Load test users and sample data: `bundle exec rake db:seed`
 5. Start the application: `bundle exec rails s`
 
+### Local Production Dependency Build
+The steps below will setup your local without dependencies that are not loaded in dev, staging, and production environments.
+1. `bundle install --without development test`
+2. `bundle clean --force`
+3. `bundle exec rails s`
```
jholton: This is just running the server locally in a `development` environment, which is standard.  This doesn't test/mimic a `production` environment.

If you want this to mimic a `production` environment, you'll need the flag: 
`bundle exec rails s -e production` so that it uses the `config/environments/production.rb` configuration.

And then you could see the css/js links are broken (e.g. if you've never precompiled assets locally), so you'll need to add
`rm -f public/assets/*` to remove old/stale compiled assets.
`bundle exec rake assets:clean assets:precompile` to compile assets.

And you'll have make sure you have a `RAILS_SERVE_STATIC_FILES` environment variable (used in `config/environments/production`) set locally so that Rails will server the assets in `public/assets`.  This is _not_ set in deployed environments because `nginx` serves the static assets there.

And you'll have to make sure `database.yml` config entry for `production` isn't trying to connect to the production database, as it will fail.

zurbergram: Correct, is a better title needed for this section as the intention to provide steps of how to test without development and test group dependencies?

jholton: I'd go with something like:
> The steps below will run `GIDS` locally with only gems installed that are used in Rails `production` environments.

But I'm still having trouble understanding what this is trying to accomplish.
What are you testing for with these commands that would be different than running the site locally after plain old `bundle install`?

zurbergram: Removed the section

jholton:
What steps can I take to reproduce the 
> undefined method 'dapip_id' in 'accreditation_action'. 

error?

zurbergram:
> What steps can I take to reproduce the
> 
> > undefined method 'dapip_id' in 'accreditation_action'.
> 
> error?

Remove `gem 'factory_bot', '4.10.0'`
Do `bundle update;bundle clean --force`
Then `bundle exec rspec`

Gemfile
```diff
@@ -84,6 +86,7 @@ group :development, :test do
   gem 'guard-rspec', require: false
 
   gem 'factory_bot_rails'
+  gem 'factory_bot', '4.10.0'
```
jholton: `factory_bot_rails` always includes the same specified version of `factory_bot`, so you can just remove the `factory_bot` listing and just specify the version on `factory_bot_rails`.  i.e.
> gem 'factory_bot_rails', '4.10.0'

zurbergram: changed

jholton (approved):
LGTM

PR #430, opened 2019-08-28
Update gibct-data-service Rails to version ~>5.2.3 #19616

zurbergram:
## Description

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19616

https://guides.rubyonrails.org/5_2_release_notes.html#removals

For convenience if anyone reading this is interested, here are [release notes](https://guides.rubyonrails.org/5_2_release_notes.html), [changelogs](https://github.com/rails/rails/releases/tag/v5.2.0) and [diffs](http://railsdiff.org/5.1.7/5.2.3) for this version upgrade.


devin-mccurdy (approved):
LGTM

Gemfile.lock
```diff
@@ -127,14 +131,14 @@ GEM
       concurrent-ruby (~> 1.0)
       dry-core (~> 0.2)
       dry-equalizer (~> 0.2)
-    dry-schema (1.3.4)
+    dry-schema (1.4.1)
```
johnpaulashenfelter: Was this necessary for the Rails 5.2.3 upgrade?

Upgrading dry-* isn't a problem, but it mixes Rails upgrade with other upgrades.

Gemfile.lock
```diff
@@ -127,14 +131,14 @@ GEM
...
       concurrent-ruby (~> 1.0)
       dry-configurable (~> 0.8, >= 0.8.3)
       dry-core (~> 0.4)
       dry-equalizer (~> 0.2)
       dry-initializer (~> 3.0)
       dry-logic (~> 1.0)
-      dry-types (~> 1.0)
+      dry-types (~> 1.2)
```
johnpaulashenfelter: Ditto above. Was this necessary for the Rails 5.2.3 upgrade?

Gemfile.lock
```diff
@@ -285,27 +293,27 @@ GEM
...
       actionpack (>= 5.0)
       railties (>= 5.0)
-    rspec (3.8.0)
-      rspec-core (~> 3.8.0)
-      rspec-expectations (~> 3.8.0)
-      rspec-mocks (~> 3.8.0)
-    rspec-core (3.8.2)
-      rspec-support (~> 3.8.0)
-    rspec-expectations (3.8.6)
+    rspec (3.9.0)
```
johnpaulashenfelter: Upgrading Rspec but would have preferred it as a separate PR.

spec/models/institution_spec.rb
```diff
@@ -207,7 +207,7 @@
 
       it 'should filter on field not existing' do
         expect(described_class.filter('institution', 'false').to_sql)
-          .to include("WHERE (\"institutions\".\"institution\" != 't')")
+          .to include("WHERE \"institutions\".\"institution\" != 't'")
```
johnpaulashenfelter: Did this break because of the upgrade? Its not a worrisome change -- just a surprise.

johnpaulashenfelter (commented):
There's a couple more changes mixed in here (Rspec, dry-*) that might confound figuring out the cause if this is rolled out and there are problems -- plus these changes are now tied to the Rails upgrade. Might make sense to separate

Also you may want to check some of the config files using http://railsdiff.org/5.1.7/5.2.3 to see what's changed in some of the support files -- there's not a lot of substance there but might be worth checking the different there to make sure you did everything that is relevant.

spec/models/institution_spec.rb
zurbergram: It did, it appears the sql generation changed

config/initializers/new_framework_defaults.rb
```diff
@@ -20,6 +20,3 @@
 
 # Require `belongs_to` associations by default. Previous versions had false.
 Rails.application.config.active_record.belongs_to_required_by_default = false
-
-# Do not halt callback chains when a callback returns false. Previous versions had true.
-ActiveSupport.halt_callback_chains_on_return_false = true
```
zurbergram: @johnpaulashenfelter This change was made because when doing `bundle exec rails s` the error below occurred 
This was removed in 5.2 https://guides.rubyonrails.org/5_2_release_notes.html#removals
```
/Users/matthewroth/dev/vets-gov/gibct-data-service/config/initializers/new_framework_defaults.rb:25:in `<top (required)>': undefined method `halt_callback_chains_on_return_false=' for ActiveSupport:Module (NoMethodError)
```

Gemfile.lock
zurbergram: After changing the rails version to 5.2.3 I ran a `bundle update` to bring in the newer gem

johnpaulashenfelter (approved):
@zurbergram I think you'd be safer separating the gem updates into single, smaller updates (eg RSpec wasn't necessary for _this_ PR) but since that's in the test group, it doesn't materially affect production so should be safe enough.

PR #435, opened 2019-09-04
Add programs to VET TEC Profile Page bah-19614

hughes-dustin:
## Description
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19614

## Acceptance Criteria
- [ ] If the provider has no programs, the following text is displayed: "Program data will be available for this provider soon. Learn more about VET TEC programs"
     a. "Learn more about VET TEC programs" is a link to the following page: https://staging.va.gov/education/about-gi-bill-benefits/how-to-use-benefits/vettec-high-tech-program/
- [ ]  If the provider has programs, the following is displayed:
     a. "Select a program below to view your estimated  benefits."
     b. A Table with the following columns: Program name, Length
     c. A row per program that displays the information loaded via the "Program" file in GIDS.


app/models/institution.rb
```diff
@@ -210,6 +210,10 @@ def school?
     institution_type_name != 'OJT'
   end
 
+  def programs
+    Program.where('facility_code' => self.facility_code)
```
shawkey-daniel: Code climate says self is redundant.

shawkey-daniel (changes requested):

app/models/institution.rb
hughes-dustin: Changes are in. Github just took it's time showing the update.

shawkey-daniel (approved):
LGTM

app/models/institution.rb
```diff
@@ -210,6 +210,10 @@ def school?
     institution_type_name != 'OJT'
   end
 
+  def programs
+    Program.where('facility_code' => facility_code)
```
shawkey-daniel: Matt asked Joe to use https://guides.rubyonrails.org/association_basics.html#has-many-association-reference. We should be consistent and use that here also.

shawkey-daniel (changes requested):

app/models/institution.rb
hughes-dustin: Added in the changes.

shawkey-daniel (changes requested):
The has_many association should have a dependent: destroy for cascade delete and you may want a belongs_to in program.

app/models/program.rb
```diff
@@ -15,4 +15,6 @@ class Program < ActiveRecord::Base
   }.freeze
 
   validates :facility_code, presence: true
+  belongs_to :insitutions
```
shawkey-daniel: Code climate detected trailing whitespace! Unacceptable!

shawkey-daniel (dismissed):

zurbergram (approved):
LGTM

delli-gatti-michael (approved):
Approved by QA

app/models/program.rb
```diff
@@ -15,4 +15,5 @@ class Program < ActiveRecord::Base
...
```
kfrz: This should be `belongs_to :institution` I think

kfrz (dismissed):
Overall looks good. I think with the long serializer it may be useful to extract out commonly used attributes, like address, out to a concern, which would reduce the length of the class as well as make the code more DRY overall.

app/models/program.rb
hughes-dustin: Nice catch. That definitely was a typo.

app/models/program.rb
```diff
@@ -15,4 +15,5 @@ class Program < ActiveRecord::Base
   }.freeze
 
   validates :facility_code, presence: true
+  belongs_to :institution
```
jholton: This line should probably just be deleted.
`programs` doesn't have a `institution_id` field.
It does have a `facility_code`, but it doesn't have a `version` attribute.  That is, there will be multiple `institutions` records (one per version) for each `program`

app/serializers/program_serializer.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class ProgramSerializer < ActiveModel::Serializer
+  attribute :facility_code
+  attribute :institution_name
```
jholton: The two above attributes should probably be removed.  It's redundant data that is available on the `institutions` object.

app/models/program.rb
hughes-dustin: Got it. Changes have been submitted.

app/serializers/program_serializer.rb
hughes-dustin: That's a 10-4. Changes are in.

jholton (approved):
lgtm

PR #441, opened 2019-09-17
Add institution programs

devin-mccurdy:
Programs need to be versioned and archived. Added `InstitutionProgram` which is a new model that is used when generating a new version.

The existing `programs` table will be used for the data from the uploaded CSV only.

Database migration PRs: 

- https://github.com/department-of-veterans-affairs/gibct-data-service/pull/442
- https://github.com/department-of-veterans-affairs/gibct-data-service/pull/447

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19680


zurbergram (approved):
LGTM

delli-gatti-michael (approved):
Approved by QA

app/models/institution.rb
```diff
@@ -161,7 +161,9 @@ class Institution < ActiveRecord::Base
   has_many :yellow_ribbon_programs, dependent: :destroy
   has_many :school_certifying_officials, -> { order 'priority, last_name' },
            primary_key: :facility_code, foreign_key: 'facility_code'
-  has_many :programs, -> { order(description: :asc) }, primary_key: :facility_code, foreign_key: 'facility_code'
+  has_many :institution_programs, -> { order(description: :asc) },
```
jholton: You also need to account for `version` in this `has_many`.  i.e. the relationship is built on a compound key of `facility_code` and `version`

spec/models/institution_builder_spec.rb
```diff
@@ -891,6 +891,17 @@
       end
     end
 
+    describe 'when generating institution programs' do
+      let(:institution) { institutions.find_by(zip: zipcode_rate.zip_code) }
```
jholton: I don't see this `institution` used anywhere.

spec/models/institution_builder_spec.rb
```diff
@@ -891,6 +891,17 @@
...
+      let(:institution_program) { InstitutionProgram.first }
+
+      it 'properly generates institution programs from weams data' do
+        create :program
+        InstitutionBuilder.run(user)
+        expect(institution_program).not_to eq(nil)
```
jholton: It's unclear to me that `InstitutionBuilder.run` caused `institution_program` to not be `nil`.
Perhaps we can more clearly test that `InstitutionBuilder.run` is causing `InstitutionProgram` to be created.
Something like:
```
     it 'properly generates institution programs from weams data' do
        create :program
        expect { InstitutionBuilder.run(user) }.to change{InstitutionProgram.count}.from(0).to(1)
      end
```

app/models/institution_builder.rb
```diff
@@ -502,4 +503,30 @@ def self.add_extension_campus_type(version_number)
...
+
+  def self.build_institution_programs(version_number)
+    str = <<-SQL
+      INSERT INTO institution_programs (
+        facility_code,
+        program_type,
+        description,
+        full_time_undergraduate,
+        graduate,
+        full_time_modifier, length_in_hours,
```
jholton: Nitpick: can we put these on two separate lines like the others?

app/models/institution_builder.rb
```diff
@@ -502,4 +503,30 @@ def self.add_extension_campus_type(version_number)
...
+        version
+      )
+      SELECT
+        facility_code,
+        program_type,
+        description,
+        full_time_undergraduate,
+        graduate,
+        full_time_modifier, length,
```
jholton: Nitpick: can we put these on two separate lines like the others?

jholton (changes requested):

app/models/institution_builder.rb
dmccurdybah: Yes!  Sorry, I was just getting addicted to green code climate checkboxes

app/models/institution.rb
dmccurdybah: Thanks!  This made me realize that I also made a mistake with the index on the migration PR (new PR coming soon...)

devin-mccurdy:
@jholton Additional DB migrations are in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/447

spec/models/institution_builder_spec.rb
```diff
@@ -891,6 +891,15 @@
       end
     end
 
+    describe 'when generating institution programs' do
+      let(:institution_program) { InstitutionProgram.first }
+
+      it 'properly generates institution programs from weams data' do
+        create :program
+        expect { InstitutionBuilder.run(user) }.to change { InstitutionProgram.count }.from(0).to(1)
```
jholton: Can we get one `expect` verifying the `version` is properly created?
I think it'd be:
`expect(institution_program.version).to eq(Version.current_preview.number)`

jholton (approved):
LGTM

PR #443, opened 2019-09-18
CT 116: Allow for new preview version to be generated without uploading a file #19667

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19667


shawkey-daniel (approved):
LGTM

sonntag-adam (approved):

app/models/upload.rb
```diff
@@ -72,6 +72,18 @@ def self.last_uploads_rows
     uploads.sort_by { |upload| upload.csv_type.downcase }
   end
 
+  def self.latest_uploads
+    preview = Version.current_preview
+    production = Version.current_production
+    return if production.blank?
+
+    created_at = preview.present? && preview.number == production.number ? production.created_at : preview.created_at
```
jholton: What is supposed to happen if `preview.present?` returns `false`?

app/models/upload.rb
```diff
@@ -72,6 +72,18 @@ def self.last_uploads_rows
     uploads.sort_by { |upload| upload.csv_type.downcase }
   end
 
+  def self.latest_uploads
```
jholton: Can we get some specs for this method please?

app/views/dashboards/index.html.erb
```diff
@@ -23,18 +23,43 @@
...
-          data: { confirm: 'Are you sure you want to create a new preview version?' },
-          class: "btn btn-warning btn-xs",
-          role: "button",
-          id: "version_build"
+    <div id="preview_dialog" title="Generate New Preview Version">
+      <div id="preview_dialog_div">
+        <p class="generate_preview_question">
+          <b>Are you sure you want to create a new preview version?</b>
+        </p>
+        <% if @latest_uploads.length > 0 %>
```
jholton: `@latest_uploads` will be `nil` if `Version.current_production` is `nil` (see `Upload.latest_uploads`)

jholton (changes requested):

app/models/upload.rb
zurbergram: Fixed logic

app/models/upload.rb
zurbergram: Created

app/views/dashboards/index.html.erb
zurbergram: Method returns empty array now if production is blank

app/models/upload.rb
jholton: Can we rename this?  I just noticed there is a `.last_uploads` too (used in this method), and so I naturally wonder what the difference is between `last uploads` and `latest uploads`.
How about we rename this something like.  `since_last_version_creation` (`Upload.since_last_version_creation`)?  Thoughts?

app/models/upload.rb
```diff
@@ -72,6 +72,18 @@ def self.last_uploads_rows
...
+    preview = Version.current_preview
+    production = Version.current_production
+    return [] if production.blank?
+
+    created_at = preview.present? && preview.number > production.number ? preview.created_at : production.created_at
+
+    Upload.last_uploads
```
jholton: You don't have to convert this into an array first.  You can do it all with SQL:
`Upload.last_uploads.where("updated_at > ?", created_at)`

app/models/upload.rb
zurbergram: renamed

app/models/upload.rb
zurbergram: changed

jholton (approved):
LGTM

PR #444, opened 2019-09-20
Edu Programs Migrations

devin-mccurdy:
Database migration PR for https://github.com/department-of-veterans-affairs/gibct-data-service/pull/445

PR #445, opened 2019-09-20
Add Edu Programs

devin-mccurdy:
Add support for a `edu_programs` CSV file that is imported and used along with the `programs` table when generating versioned `InstitutionProgram` records.

Database migration PR: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/444

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19668


zurbergram (approved):
LGTM

shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):
Approved by QA

app/models/institution_builder.rb
```diff
@@ -514,18 +514,45 @@ def self.build_institution_programs(version_number)
...
+        phone_number,
+        student_vet_group,
+        student_vet_group_website,
+        vet_success_name,
+        vet_success_email,
+        vet_tec_program,
+        tuition_amount,
+        length_in_weeks
+      FROM programs a INNER JOIN edu_programs b
+        ON a.facility_code = b.facility_code
```
jholton: Can a `facility_code` occur more than once in the `program.csv`?
I see in your `edu_program.csv` fixture that an `edu_program.csv` upload can.
So, for example, if there are four rows in `program.csv` for `facility_code` 13000313 and there are four rows in `edu_program.csv` for the same `facility_code`, you're going to get 16 rows inserted into `institution_programs` with all the various combinations.
Is what you want?

dmccurdybah: Yes, institutions can have multiple programs.  Because this is also joining on the program name (`programs.description` and `edu_programs.vet_tec_program`) you wouldn't get 16 in the case of your example.  Assuming that the 4 rows in each joined table had 4 unique names that matched, you'd get 4.

jholton: Ah yup.  Somehow I missed that.

jholton (approved):
LGTM.

PR #446, opened 2019-09-20
CT 116: Validate Program Data upload to GIDS #19669

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19669

## Screenshots
Screen Shot 2019-09-20 at 3 47 34 PM
Screen Shot 2019-09-20 at 3 43 27 PM


shawkey-daniel (changes requested):
It would be nice to build the error messages from constants and use that in the validators and the unit tests.

shawkey-daniel (approved):
LGTM

app/validators/edu_program_validator.rb
```diff
@@ -0,0 +1,23 @@
...
+    "The Facility Code & VET TEC Program (Program Name) combination is not unique:
+#{record.facility_code}, #{record.vet_tec_program}"
+  end
+
+  def self.missing_facility_code_error_msg(record)
+    "The Facility Code #{record.facility_code} is not contained within the most recently uploaded weams.csv"
+  end
+
+  def validate(record)
+    if EduProgram.where(['vet_tec_program = ? AND facility_code = ? AND id != ?',
```
jholton: Can we clean this up for readability?  Suggestion:
```
duplicates_exist = EduProgram.where(['vet_tec_program = ? AND facility_code = ? AND id != ?', record.vet_tec_program, record.facility_code, record.id]).any?
record.errors[:base] << EduProgramValidator.non_unique_error_msg(record) if duplicates_exist

facility_not_in_weam = Weam.where(['facility_code = ?', record.facility_code]).empty?
record.errors[:base] << EduProgramValidator.missing_facility_code_error_msg(record) if facility_not_in_weam
```

app/validators/program_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+    "The Facility Code & Description (Program Name) combination is not unique:
+#{record.facility_code}, #{record.description}"
+  end
+
+  def self.missing_facility_code_error_msg(record)
+    "The Facility Code #{record.facility_code} is not contained within the most recently uploaded weams.csv"
+  end
+
+  def validate(record)
+    if Program.where(['description = ? AND facility_code = ? AND id != ?',
```
jholton: Can we clean this up similar to the other suggestion?

app/models/edu_program.rb
```diff
@@ -21,4 +21,5 @@ class EduProgram < ActiveRecord::Base
   }.freeze
 
   validates :facility_code, presence: true
+  validates_with EduProgramValidator, on: :load_csv
```
jholton: Why `on: :load_csv`?  Shouldn't this validation be relevant all the time?

app/models/program.rb
```diff
@@ -15,4 +15,5 @@ class Program < ActiveRecord::Base
   }.freeze
 
   validates :facility_code, presence: true
+  validates_with ProgramValidator, on: :load_csv
```
jholton: Why `on: :load_csv`?  Shouldn't this validation be relevant all the time?

app/controllers/uploads_controller.rb
```diff
@@ -41,14 +41,14 @@ def show
...
+    alert_and_log("Upload with id: '#{params[:id]}' not found", nil)
     redirect_to uploads_path
   end
 
   private
 
-  def alert_and_log(message)
-    Rails.logger.error message
+  def alert_and_log(message, error)
+    Rails.logger.error message + (error&.backtrace).to_s
```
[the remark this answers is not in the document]
zurbergram: This cleans up the display of error messages to the users when uploading CSVs, full stacktrace is still going into the log

app/models/edu_program.rb
zurbergram: Without this validations are being run twice in `lib/csv_helper/loader.rb` which was resulting in "missing facility code from weams" being displayed twice per relevant row on upload

The "facility code + description/program name" validations were only failing after all rows were inserted into the table since it is dependent on table data

zurbergram: Since I removed the `:validate false` from the import, validations without a context will be run on import and failing rows will end up the `results.failed_instances` array 

This is used at https://github.com/department-of-veterans-affairs/gibct-data-service/blob/5ce4a203d35accf3ae2dda413ee639a5b45303ae/lib/csv_helper/loader.rb#L66

app/validators/edu_program_validator.rb
zurbergram: changed

app/validators/program_validator.rb
zurbergram: changed

app/models/edu_program.rb
zurbergram: The application is setup in a way that only time any validations are run on csv model objects is during csv upload

app/models/program.rb
zurbergram: see comments on edu_program.rb

delli-gatti-michael (approved):
Approved by QA

app/models/edu_program.rb
jholton: tl;dr: We don't want this validation to prevent records from being persisted to the database. We are providing a validation context so we can manually run the validation post-import to generate warnings for the end user.

app/validators/edu_program_validator.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class EduProgramValidator < ActiveModel::Validator
+  def self.non_unique_error_msg(record)
```
jholton: Readability suggestion:  Can we turn this into a private instance method and move it below `validate`.
```
def validate(record)
  # code
end

private

def non_unique_error_msg(record)
  # code
end
```

app/validators/program_validator.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class ProgramValidator < ActiveModel::Validator
+  def self.non_unique_error_msg(record)
```
jholton: Readability suggestion:  Can we turn this into a private instance method and move it below `validate`.
```
def validate(record)
  # code
end

private

def non_unique_error_msg(record)
  # code
end
```

app/validators/validator_util.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ValidatorUtil
```
jholton: How about we make this a parent class, or mixin for the other validators, and name it something more specific (something like `ProgramBaseValidator`)?
```
class ProgramBaseValidator < ActiveModel::Validator
...
class EduProgramValidator < ProgramBaseValidator
...
class ProgramValidator < ProgramBaseValidator
```
Then you can turn the methods into private instance methods, and call them from the subclasses.

lib/csv_helper/loader.rb
```diff
@@ -73,5 +54,22 @@ def merge_options(options)
       options.reverse_merge(key_mapping: key_mapping, value_converters: value_converters)
              .reverse_merge(SMARTER_CSV_OPTIONS)
     end
+
+    # default validations are run during import to prevent bad data from going into table,
+    # this runs validations with a context of :load_csv
+    # validations with this context are ran now due either the validation needing data in the table being loaded into
+    # or the validation causes issues on save
+    def load_csv_validation(records, failed_instances, options)
```
jholton: Can we rename this something more descriptive like `after_import_validations`?

lib/csv_helper/loader.rb
```diff
@@ -73,5 +54,22 @@ def merge_options(options)
       options.reverse_merge(key_mapping: key_mapping, value_converters: value_converters)
              .reverse_merge(SMARTER_CSV_OPTIONS)
     end
+
+    # default validations are run during import to prevent bad data from going into table,
+    # this runs validations with a context of :load_csv
+    # validations with this context are ran now due either the validation needing data in the table being loaded into
+    # or the validation causes issues on save
```
jholton: This comment is a little confusing.  Can we clean it up. Suggestion:
```
 # Default validations are run during import, which prevent bad data from being persisted to the database.
 # This method manually runs validations that were declared with a specific validation context (:load_csv).
 # The result is warnings are generated for the end user while the data is allowed to persist to the database. 
```

lib/csv_helper/loader.rb
```diff
@@ -73,5 +54,22 @@ def merge_options(options)
...
+    def load_csv_validation(records, failed_instances, options)
+      # Since row indexes start at 0 and spreadsheets on line 1,
+      # add 1 for the difference in indexes and 1 for the header row itself.
+      row_offset = CSV_FIRST_LINE + (options[:skip_lines] || 0)
+
+      records.each_with_index do |record, index|
+        unless record.valid?(:load_csv)
+          record.errors.add(:row, "Line #{index + row_offset}")
+          failed_instances << record
```
jholton: You can change this to
```suggestion
          failed_instances << record if record.persisted?
```
This will prevent duplicates of the records that were marked invalid during the import from being added to the array.  Then you can remove the `.uniq` call in `uploads_controller#create`

lib/csv_helper/shared.rb
```diff
@@ -14,7 +14,11 @@ def display_errors_with_row
     keys = errors.keys - [:row]
 
     "Row #{row.presence || 'N/A'} : " + keys.map do |key|
-      key.to_s + ' : ' + errors[key].join(', ')
+      if key.to_s == 'base'
+        errors[key].join(', ')
+      else
+        key.to_s + ' : ' + errors[key].join(', ')
+      end
```
jholton: Refactor suggestion (DRY):
```
message = key.to_s == 'base' ? '' : "#{key.to_s} : "
message += errors[key].join(', ')
```

spec/validators/edu_program_validator_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+describe EduProgramValidator do
+  describe 'when validating for load_csv context' do
```
jholton: This validator doesn't know anything about `load_csv`.
Let's reword this something like 
```
context 'when record is unique and facility_code is present in Weam table' do
```

spec/validators/edu_program_validator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'has no errors' do
+      edu_program = create :edu_program
+
+      expect(edu_program).to be_valid
+    end
+
+    it 'has invalid facility_code & vet_tec_program error message' do
```
jholton: And this should be in its own context:
```
context 'when record does not have unique facility_code & vet_tec_program values' do
  it 'fails validation' do
```

spec/validators/edu_program_validator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(edu_program.valid?(:load_csv)).to eq(false)
+      expect(edu_program_b.valid?(:load_csv)).to eq(false)
+
+      error_messages = edu_program.errors.messages
+      expect(error_messages.any?).to eq(true)
+
+      expect(error_messages[:base]).to include(EduProgramValidator.non_unique_error_msg(edu_program))
+    end
+
+    it 'has invalid facility code error message' do
```
jholton: Same here.
And this should be in its own context:
```
context 'when a corresponding weams record does not exist' do
  it 'fails validation' do
```

spec/validators/program_validator_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+describe ProgramValidator do
+  describe 'when validating for load_csv context' do
```
jholton: Fix up wording as suggested in `edu_program_validator_spec.rb`

app/validators/edu_program_validator.rb
zurbergram: This method is used in the unit test

app/validators/program_validator.rb
zurbergram: This method is used in the unit test

app/validators/edu_program_validator.rb
zurbergram: changed

app/validators/program_validator.rb
zurbergram: changed

app/validators/validator_util.rb
zurbergram: renamed to BaseValidator since other csvs in future might need a facility_code validation

lib/csv_helper/loader.rb
zurbergram: changed

lib/csv_helper/loader.rb
zurbergram: changed

lib/csv_helper/loader.rb
zurbergram: changed to `unless record.valid?(:after_import) && !record.persisted?` instead as to prevent an additional error being added to the record if not persisted

lib/csv_helper/shared.rb
zurbergram: changed

spec/validators/edu_program_validator_spec.rb
zurbergram: changed

spec/validators/edu_program_validator_spec.rb
zurbergram: changed

spec/validators/edu_program_validator_spec.rb
zurbergram: changed

spec/validators/program_validator_spec.rb
zurbergram: changed

app/validators/base_validator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class BaseValidator < ActiveModel::Validator
+  def facility_code_in_weam?(record)
```
jholton: This one can be `private` too.  Only `validate` (on the subclass) is ever called directly

jholton (approved):
LGTM

PR #448, opened 2019-09-25
Staging CT 116: Able to close out of generating new preview version by pressing ESC key #19732

zurbergram:
## Description

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19732

### DIFF
--- a/app/assets/javascripts/preview.js
+++ b/app/assets/javascripts/preview.js
@@ -9,7 +9,7 @@ $(function() {
                 '</tr>');
         }
     });
-    var generating_preview_version = false;
+    let generating_preview_version = false;
     $( "#preview_dialog" ).dialog({
         autoOpen: false,
         modal: true,
--- a/app/assets/javascripts/preview.js
+++ b/app/assets/javascripts/preview.js
@@ -13,6 +13,7 @@ $(function() {
         autoOpen: false,
         modal: true,
         width: "auto",
+        closeOnEscape: false,
         resizable: false,
         hide: {
             effect: "fade",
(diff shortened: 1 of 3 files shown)


devin-mccurdy (approved):
LGTM

delli-gatti-michael (approved):

app/helpers/dashboards_helper.rb
```diff
@@ -12,4 +12,8 @@ def latest_upload_title(upload)
     return 'Missing required upload' if CSV_TYPES_REQUIRED_TABLE_NAMES.include?(upload.csv_type)
     'Missing upload'
   end
+
+  def can_generate_preview(preview_versions)
+    'disabled' unless preview_versions.none? || !preview_versions[0].generating?
```
jholton: How about this (easier to read):
```suggestion
    'disabled' if preview_versions[0]&.generating?
```

app/assets/javascripts/preview.js
```diff
@@ -9,22 +9,30 @@ $(function() {
                 '</tr>');
         }
     });
+    var generating_preview_version = false;
```
jholton: We should use `let` instead of `var`.
```suggestion
    let generating_preview_version = false;
```

app/helpers/dashboards_helper.rb
zurbergram: changed

app/assets/javascripts/preview.js
zurbergram: changed

jholton (approved):
LGTM

PR #452, opened 2019-10-08
Institution Programs Search

devin-mccurdy:
Created a InstitutionProgram controller to allow searching and autocomplete of InstitutionPrograms.

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19734

DB migration PR: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/455


app/models/institution_program.rb
```diff
@@ -1,4 +1,66 @@
 # frozen_string_literal: true
 
 class InstitutionProgram < ActiveRecord::Base
```
zurbergram: This should extend `ApplicationRecord`

zurbergram (approved):
LGTM

shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

app/models/institution_program.rb
```diff
@@ -1,4 +1,83 @@
...
+      .limit(limit)
+  end
+
+  # Finds exact-matching facility_code or partial-matching school and city names
+  #
+  scope :search, lambda { |search_term, _include_address = false|
+    return if search_term.blank?
+
+    clause = [
+      'institution_programs.facility_code = (:facility_code)',
```
jholton: If we are going for data normalization, we no longer need `facility_code` or `version` in the `institution_programs` table.  That data already exists in `institutions`.  I think that's all of the duplicated fields.

dmccurdybah: That sounds good. I can make the changes in this PR to stop using `institutions_programs.facility_code` and then drop that in a following migration PR.

app/models/institution_program.rb
```diff
@@ -42,7 +42,12 @@ def va_bah
   #
   def self.autocomplete(search_term, limit = 6)
```
jholton: Can we get a unit test for this please?

dmccurdybah: Can do!

app/controllers/v0/institution_programs_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+                                   .joins(:institution)
+                                   .search(@query[:name])
+                                   .order('institutions.preferred_provider DESC NULLS LAST, institutions.institution')
+      [
+        %i[program_type type],
+        %w[institutions.physical_country country],
+        %w[institutions.physical_state state],
+        %w[institutions.preferred_provider preferred_provider]
+      ].each do |filter_args|
+        filter_args << filter_args[0] if filter_args.size == 1
```
jholton: This line is unnecessary (there are no `filter_args` with a `size` of 1)

app/controllers/v0/institution_programs_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+      result = {
+        type: count_field(search_results, :program_type),
+        state: count_field(search_results, :state),
+        country: embed(count_field(search_results, :country))
+      }
+      add_active_search_facets(result)
+    end
+
+    def count_field(relation, field)
+      field_map = {}
```
jholton: You can specify a default value for new hash keys like so:
```suggestion
      field_map = Hash.new(0)
```
Then you can remove line 78, since you no longer have to check that a key is present first.

app/models/institution_program.rb
```diff
@@ -1,4 +1,90 @@
...
+
+  belongs_to :institution
+  delegate :dod_bah, to: :institution
+  delegate :preferred_provider, to: :institution
+
+  def institution_name
+    institution.institution
+  end
+
+  delegate :facility_code, to: :institution
```
jholton: Can we move this up with the other `delegate`s?

app/controllers/v0/institution_programs_controller.rb
dmccurdybah: Thanks!

jholton (approved):

PR #455, opened 2019-10-09
Institution Program Migration

devin-mccurdy:
DB migration PR for https://github.com/department-of-veterans-affairs/gibct-data-service/pull/452


zurbergram (approved):
LGTM

jholton:
Looks like a lot of duplication of data (new fields in `institution_programs`) and code (in #452).
Can't we do our searching on `institutions` and return `institution_programs` data via SQL joins?

db/migrate/20191004151515_create_versioned_school_certifying_officials_archives.rb
```diff
@@ -1,4 +1,4 @@
-class CreateVersionedSchoolCertifyingOfficialsArchives < ActiveRecord::Migration[4.2]
+class CreateVersionedSchoolCertifyingOfficialsArchives < ActiveRecord::Migration[5.0]
```
johnpaulashenfelter: Did this not work with 4.2? I don't see a reason to change it but either is fine.

db/migrate/20191009142101_add_institution_columns_to_institution_programs_archives.rb
```diff
@@ -0,0 +1,40 @@
+class AddInstitutionColumnsToInstitutionProgramsArchives < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
```
johnpaulashenfelter: Same comment as above -- not sure about the safety of disabling the transaction for a schema change (add_column) vs the index.

db/migrate/20191009142100_add_institution_columns_to_institution_programs.rb
```diff
@@ -0,0 +1,40 @@
+class AddInstitutionColumnsToInstitutionPrograms < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
```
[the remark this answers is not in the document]
johnpaulashenfelter: Did a little more digging: https://github.com/LendingHome/zero_downtime_migrations#disabling-the-ddl-transaction

johnpaulashenfelter (commented):
I think the migrations need some separation into things that should and shouldn't be transactional.

db/migrate/20191009142101_add_institution_columns_to_institution_programs_archives.rb
dmccurdybah: Sounds good.  I moved the index creations to their own PRs

db/migrate/20191004151515_create_versioned_school_certifying_officials_archives.rb
dmccurdybah: Just cleanup, really.  I'm trying to be consistent and make these match the actual version when the migration was added.

devin-mccurdy:
> Looks like a lot of duplication of data (new fields in `institution_programs`) and code (in #452).
> Can't we do our searching on `institutions` and return `institution_programs` data via SQL joins?

@jholton 

As we're building out the front end for the VET TEC programs (which is being restructured to separate programs from institutions), it make sense that Programs would be a top-level model which would have its own service.  Accessing them through Institutions doesn't feel like the correct solution.

Would you rather us create a relationship between the InstitutionProgram and Institution instead of adding the fields to the InstitutionProgram?  (I'm assuming we'd be able to flatten that in the serializer)

jholton:
> Would you rather us create a relationship between the InstitutionProgram and Institution instead of adding the fields to the InstitutionProgram? (I'm assuming we'd be able to flatten that in the serializer)

Yes, I feel like this is the best approach.  It will prevent us from duplicating data and code. It will make it clearer to new developers what the relationships are between `institutions` and `institution_programs`.

Currently these models don't follow Rails association conventions (compound, non-standard foreign keys), and thus declaring a standard `has_many`/`belongs_to` relationship is not possible right now.
You might want to look into changing that:  Add a `institutions_id` column to `institution_programs` table and populate it during `InstitutionBuilder.run`.  Then you can establish`has_many`/`belongs_to` relationship and use more ActiveRecord helpers.

Alternatively, this might be a good case for using the [Sequel](https://github.com/jeremyevans/sequel) gem to construct your SQL queries.

db/migrate/20191009142101_add_institution_columns_to_institution_programs_archives.rb
```diff
@@ -0,0 +1,5 @@
+class AddInstitutionColumnsToInstitutionProgramsArchives < ActiveRecord::Migration[5.1]
+  def change
+    add_column :institution_programs_archives, :institution_id, :integer
```
jholton: Hmmm, if we go this route, then this should be `institutions_archive_id`, so the association will be accurate (i.e. after archiving, it is now associated with `institutions_archives` and not `institutions`

dmccurdybah: Right now we're expecting the schema of the archives table to match the original and we don't expect there to be a valid relationship between the institutions_programs_archives institutions_archives data.  Does that make sense?

jholton: Then I have to ask: what is the point in archiving the `institution_programs` data if it can't be tied back to the `institutions`?

dmccurdybah: The data is still there, but it's not a rails relationship.  Right now there isn't a method to revert institutions, but if this is something that we wanted to do, it could be done (and it would be done through SQL and this wouldn't be an issue).  This is really just a method of archiving old records because there was no system and each new version generates 30k+ rows.

jholton: Right.  Now it also occurs to me that you can specify a different foreign key for the relationship between `InstitutionsArchive` and `InstitutionProgramsArchive`.

db/migrate/20191022125700_update_institution_programs.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateInstitutionPrograms < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :institution_programs, :facility_code, true
```
jholton: What's the case for needing this change?

dmccurdybah: We'll need to allow nulls so that we can stop populating this field in instutution_builder. 

Because of the way we've been asked to break up our PRs I think we'll need:

- Migration PR to allow nulls
- PR to remove field usage
- Follow up migration PR to remove column

db/migrate/20191015133100_update_institution_programs_indexes.rb
```diff
@@ -0,0 +1,17 @@
...
+  def change
+    remove_index :institution_programs, name: "index_institution_programs"
+
+    add_index :institution_programs,
+              [:institution_id],
+              using: :btree,
+              algorithm: :concurrently
+
+    add_index :institution_programs,
+              [:description, :version],
```
jholton: We're removing `version`, so we don't need to index on it.

dmccurdybah: I wasn't planning on getting rid of `version`.  Maybe I'm missing something, but it's still a search parameter used by vets-website, so I think we'd still want to use it in the index.  Also I think we'd still want it for archiving.

jholton: We're removing it for the same reason we're removing `facility_code`:  it's on the `institutions` record and will be searchable there.

jholton (approved):

PR #458, opened 2019-10-11
CT: Update security rake task to use latest gems and fix issues #19786

zurbergram:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19786

PR #463, opened 2019-10-21
Comment re-added to config.force_ssl

jbritt1:
Commenting this back out as this is a likely suspect in recent job failures:
https://dsva.slack.com/archives/C37M86Y8G/p1571682085057200

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -42,7 +42,7 @@ Rails.application.configure do
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  # config.force_ssl = true
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -42,6 +42,7 @@ Rails.application.configure do
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  # We terminate SSL before traffic gets to the gi data service elb and traffic from the elb to the service is over http. So forcing ssl will break our ELB health checks.
   # config.force_ssl = true
 
   # Use the lowest log level to ensure availability of diagnostic information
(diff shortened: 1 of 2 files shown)


tskinn:
@zurbergram we want to this change of yours https://github.com/department-of-veterans-affairs/gibct-data-service/pull/458/files#diff-1d98f0039bbc6df92adee882ed99d993L45. It appears to be breaking new deployments due to health check failures.

We terminate SSL before traffic gets to the gi data service elb and traffic from the elb to the service is over http. So forcing ssl is breaking our ELB health checks as they are returning a 301 to redirect to https.

annaswims (approved):
LGTM

annaswims:
@tskinn's comment could be helpful to keep in the config.

jbritt1:
@annaswims I added Taylor's comment, but I do not have access to see why it couldn't be built. Any insight?

annaswims:
brakeman, our security scanners is not liking the change.

make  `config/brakeman.ignore`
````
{
  "ignored_warnings": [
    {
      "warning_type": "Missing Encryption",
      "warning_code": 109,
      "fingerprint": "6a26086cd2400fbbfb831b2f8d7291e320bcc2b36984d2abc359e41b3b63212b",
      "check_name": "ForceSSL",
      "message": "The application does not force use of HTTPS: `config.force_ssl` is not enabled",
      "file": "config/environments/production.rb",
      "line": 1,
      "link": "https://brakemanscanner.org/docs/warning_types/missing_encryption/",
      "code": null,
      "render_path": null,
      "location": null,
      "user_input": null,
      "confidence": "High",
      "note": "We terminate SSL before traffic gets to the gi data service elb and traffic from the elb to the service is over http."
    }
  ],
  "updated": "2019-10-21 15:40:56 -0400",
  "brakeman_version": "4.6.1"
}
````

**Comment by jbritt1** [comment]
Ah, this is the message that likely drove the change in the first place I assume.

File: config/environments/production.rb [hunk 1]
```diff
@@ -42,7 +42,8 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  # We terminate SSL before traffic gets to the gi data service elb and traffic from the elb to the service is over http. So forcing ssl will break our ELB health checks.
+  # config.force_ssl = true
```
wyattwalter: @annaswims I notice the last part of this is enabling Rails to "use secure cookies". Do you know if we would need to do anything different to make sure cookies have the secure attribute set? I'm not sure how to login to this app, otherwise I could check quick.

PR #481, opened 2019-10-29
Address RSpec/ExampleLength rubocop errors in GIDS

hughes-dustin:
## Description
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19830

## Testing done
Dev tested

## Acceptance criteria
- [ ] There are no longer any of the following offenses:
     - RSpec/ExampleLength: # Offense count: 40

## Definition of done


sonntag-adam (approved):

spec/models/institution_builder_spec.rb
```diff
@@ -53,32 +60,31 @@
...
 
       it 'does not change the institutions or versions if not successful' do
         allow(described_class).to receive(:add_crosswalk).and_raise(StandardError, 'BOOM!')
         create :version
-        version = Version.current_preview
-
         described_class.run(user)
         expect(Institution.count).to be_zero
-        expect(Version.current_preview).to eq(version)
+        expect(Version.current_preview).to eq(Version.current_preview)
```
shawkey-daniel: Are you actually checking anything here?

shawkey-daniel (changes requested):

spec/models/institution_builder_spec.rb
hughes-dustin: No it doesn't. I've put a fix in.

shawkey-daniel (approved):
LGTM

spec/controllers/v0/institution_programs_controller_spec.rb
```diff
@@ -3,34 +3,43 @@
...
+
+  def check_response_preview_number(response, version_number)
+    body = JSON.parse response.body
+    expect(body['meta']['version']['number'].to_i).to eq(version_number)
+  end
+
+  def create_preview_version
+    create(:version, :production)
+    create(:version, :preview)
+  end
```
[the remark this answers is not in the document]
hughes-dustin: Good find. I implemented the change.

spec/models/institution_builder_spec.rb
```diff
@@ -21,18 +26,20 @@
...
         expect(version).not_to eq(old_version)
         expect(version.production).to be_falsey
         expect(version).not_to be_generating
       end
 
+      it 'returns the new preview version record if sucessful' do
+        create :version
+        old_version = Version.current_preview
+        version = described_class.run(user)[:version]
+        expects_successful_preview(version, old_version)
```
jholton: The overview of this cop is
> A long example is usually more difficult to understand. Consider extracting out some behaviour, e.g. with a `let` block, or a helper method.

I'm fine with extracting commonly used expectations to methods, but I think here this does not help make the example easier to understand.  Since this is the only spec that uses this method, I just have to go look in two different places to understand the whole spec.

I think this cop also assumes we are writing specs with one expectation each.  This is generally regarded as best practice.  Here we have 4 expectations.  However, breaking this up into four tests adds a lot of overhead (i.e. the objects would have to be created in the database 4 times), so I'm not necessarily a stickler for that "best practice".

I think we should move the expectations back into this spec and bump the max number of lines for this cop in `.rubocop.yml` to.....8?  That gives us room since we often have multiple expectations in a spec, yet still will let us know that a spec might be getting unwieldy. 
Thoughts?

hughes-dustin: My team and I have discussed that changing the line limit to 8 will be the best move. I'm going to go back and refactor these files and then get back with you.

annaswims (approved):
LGTM

PR #501, opened 2019-11-26
For CSV upload, add batch_size to CSV active_record#import

zurbergram:
## Description
Add a configurable batch size property to the active_record#import statement for CSV uploads
Bumping property value from 5000 to 50,000 as file that is requiring this change has almost 400,000 rows and next largest file only has around 70,000 rows
## Testing done
- local testing

## Acceptance criteria
- [ ] can upload a CSV through GIDs frontend
- [ ] can "publish to production" 
- [ ] sitemap generation succeeds in the server log during "publish to production"
- [ ] can export a CSV type through GIDs frontend

## Definition of done

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,5 +1,5 @@
 active_record:
-  batch_size: 5000
+  batch_size: 50000
 
 archiver:
   archive: true
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,5 +1,7 @@
 active_record:
-  batch_size: 50000
+  batch_size:
+    find_each: 5000
+    import: 50000
 
 archiver:
   archive: true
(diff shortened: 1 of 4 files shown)


config/settings.yml
```diff
@@ -1,5 +1,5 @@
 active_record:
-  batch_size: 5000
+  batch_size: 50000
```
annaswims: I think we'll need to take a look at performance, but I have no reason to believe that a batch size of 50k is preferable.  That would lead to roughly 65mb insert statements

annaswims: These graphs might help us arrive at a number http://grafana.vfs.va.gov/d/rk7GzYxWk/ww-gids-outage-postmortem?orgId=1&var-datasource=Prometheus%20(Utility)&from=1574350105214&to=1574357153057

shawkey-daniel (approved):
LGTM

annaswims (approved):

PR #502, opened 2019-11-26
Handle InstitutionProgram duplicate on version generation

devin-mccurdy:
## Description

The program source data has both duplicate programs and multiple versions of the same program (where the names are the same but with different lengths).  This isn't supported so the current solution will pull in those duplicates and multiples grouped by name and default the length to `0`

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/20037

## Testing done

## Screenshots
N/A

## Acceptance criteria
- [x] When a provider has multiple programs with the same name, the program is only displayed once in the VET TEC search results.

## Definition of done


delli-gatti-michael (approved):

shawkey-daniel (approved):
LGTM

jholton:
Running this SQL insert locally.  The old version took about 6ms.  The new version takes about 12 seconds. 
Is that an acceptable increase in processing time?

devin-mccurdy:
> Running this SQL insert locally. The old version took about 6ms. The new version takes about 12 seconds.
> Is that an acceptable increase in processing time?

Yes, it should be. The import typically only happens a few times a week.

jholton:
How do you determine a duplicate `institution_programs` record?  What fields have to be identical?

devin-mccurdy:
> How do you determine a duplicate `institution_programs` record? What fields have to be identical?

We're considering any `institution_program` records with the same institution and program name to be duplicates.

jholton:
That's what I suspected.
You could leave the insert unchanged, then delete the duplicates after the fact:
```sql
delete from institution_programs where id not in (
  SELECT min(id) from institution_programs group by description, institution_id
);
```
It'll keep the insert more readable, and save you about 12 seconds per build.

devin-mccurdy:
> That's what I suspected.
> You could leave the insert unchanged, then delete the duplicates after the fact:
> 
> ```sql
> delete from institution_programs where id not in (
>   SELECT min(id) from institution_programs group by description, institution_id
> );
> ```
> 
> It'll keep the insert more readable, and save you about 12 seconds per build.

The reason that we are including the duplicates this way is that we want to set the values of their program length values to 0.  The program data includes programs with the same name but different lengths and this is the requested solution since we don't have a good way to choose which one to keep vs delete.

Also I should add that this is a temporary solution until the data can eventually be de-duped.

jholton:
You can still do that and save the complexity and time:
```sql

update institution_programs set length_in_hours = 0 where id in (
  select min(id) from institution_programs group by description, institution_id having count(*)>1
);

delete from institution_programs where id not in (
  SELECT min(id) from institution_programs group by description, institution_id
);
```

devin-mccurdy:
> You can still do that and save the complexity and time:
> 
> ```sql
> update institution_programs set length_in_hours = 0 where id in (
>   select min(id) from institution_programs group by description, institution_id having count(*)>1
> );
> 
> delete from institution_programs where id not in (
>   SELECT min(id) from institution_programs group by description, institution_id
> );
> ```

Yep, I can do that.  I didn't want to put rows into the table and then immediately delete them, but if that isn't an issue this works.

shawkey-daniel:
This change works as is, but would be a problem, if we ever added a unique constraint to the table.

jholton:
> This change works as is, but would be a problem, if we ever added a unique constraint to the table.

True.  In that case, one possible tool could be to use an intermediary temporary table to make the UPDATE/DELETE actions before inserting into `institution_programs`.

jholton (approved):
LGTM

PR #503, opened 2019-12-02
fix codelclimate file name

annaswims:
yaml files should be `.yml` not `.yaml` . fixes #496


annaswims:
This codeclimate build still failed, but I think this file needs to be in master for the config to take effect.

LindseySaari (approved):
👍

PR #504, opened 2019-12-02
GIDS: Enhance the data file upload process. #20052

zurbergram:
## Description
As a GIDS user, I would like to be able to upload files as they are received without having to manipulate them in any way so that extra steps are removed from my process.

https://github.com/department-of-veterans-affairs/va.gov-team/issues/4035

## Acceptance criteria
- [x] 1. Files that are delimited by the following values can be uploaded:
     - comma ,
     - pipe |
- [x] 2. Files that have the following extensions can be uploaded:
     - .csv
     - .txt
- [x] 3. Files that are exported use the default column separator specified for the CVS type in https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/config/csv_file_defaults.yml
- [x] 4. When uploading a file with an opening double quote (") in a column but not a closing double quote (") an error message is displayed letting user know that there is bad data and it is related to missing closing double quote (")

## Definition of done


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

app/models/upload.rb
```diff
@@ -40,7 +40,7 @@ def csv_type_check?
   end
 
   def check_for_headers
-    return unless upload_file && csv_type && skip_lines
+    return unless upload_file && csv_type && skip_lines && col_sep
```
kfrz: Chaining `&&` like this makes me want to see a dedicated method call, like `return unless condition_is_met`

kfrz (approved):
This looks good to go, thanks for adding specs! 

I did make one comment re: a stylistic nitpick.

For completeness' sake, @zurbergram would you mind linking issue and updating PR body w/ AC items checked etc.

app/models/upload.rb
zurbergram: added method

PR #507, opened 2019-12-04
Improve performance of Program CSV after import validations

zurbergram:
## Description

Currently after import validations creates 2 sql statements per row in the uploaded CSV. This refactor uses 2 sql statements instead to get the rows along with their csv_row number that fail the after import validations. This should improve the performance of after import validations for Program CSV

Database migration PR https://github.com/department-of-veterans-affairs/gibct-data-service/pull/509

https://github.com/department-of-veterans-affairs/va.gov-team/issues/4037

## Testing done
- local testing with approx 400K row file and small 15 line file

## Acceptance criteria
- [x] can upload a Program CSV 
- [x] see https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19669 

## Definition of done


shawkey-daniel (approved):
LGTM

app/models/program.rb
```diff
@@ -16,5 +16,58 @@ class Program < ApplicationRecord
 
   validates :facility_code, presence: true
   validates :program_type, inclusion: { in: InstitutionProgram::PROGRAM_TYPES }
-  validates_with ProgramValidator, on: :after_import
+
+  def self.after_import_batch_validations(_records, failed_instances, _row_offset)
```
jholton: Can we remove the unused parameters?

app/models/program.rb
```diff
@@ -16,5 +16,58 @@ class Program < ApplicationRecord
...
+  end
+
+  def self.missing_facility_in_weam
+    str = <<-SQL
+      SELECT programs.csv_row, programs.facility_code
+      FROM programs LEFT OUTER JOIN weams ON programs.facility_code = weams.facility_code
+      WHERE weams.facility_code IS NULL
+    SQL
+    sql = Program.send(:sanitize_sql, [str])
+    Program.connection.execute(sql)
```
jholton: Can be replaced with:
```ruby
Program.joins('LEFT OUTER JOIN weams ON programs.facility_code = weams.facility_code').where(weams: { facility_code: nil})
```
Then you can use the objects above.

app/models/program.rb
zurbergram: Removed

app/models/program.rb
```diff
@@ -16,5 +16,58 @@ class Program < ApplicationRecord
...
+  def self.after_import_batch_validations(_records, failed_instances, _row_offset)
+    duplicate_facility_description_results.each do |record|
+      message = line_number(record['csv_row']) + non_unique_error_msg(record['facility_code'], record['description'])
+      warning = { index: record['csv_row'], message: message }
+      failed_instances << warning
+    end
+
+    missing_facility_in_weam.each do |record|
+      program = Program.new(record)
+      message = line_number(program.csv_row) + BaseValidator.missing_facility_error_msg(program)
```
jholton: Can just use `record` without instantiating new objects here with the change suggested below.

app/models/program.rb
zurbergram: changed

app/models/program.rb
zurbergram: changed

app/models/program.rb
```diff
@@ -16,5 +16,52 @@ class Program < ApplicationRecord
...
+        GROUP BY
+          UPPER(facility_code),
+          UPPER(description)
+        HAVING COUNT(*) > 1
+        ) dupes on UPPER(programs.facility_code) = dupes.facility_code
+        AND UPPER(programs.description) = dupes.description
+    SQL
+
+    sql = Program.send(:sanitize_sql, [str])
+    Program.connection.execute(sql)
```
jholton: We can make this more Rails-y too like so:
```ruby
subquery = Program.select('UPPER(facility_code) as facility_code, UPPER(description) as description').group('UPPER(facility_code), UPPER(description)').having('count(*) > 1')
Program.joins("INNER JOIN (#{subquery.to_sql}) dupes on UPPER(programs.facility_code) = dupes.facility_code AND UPPER(programs.description) = dupes.description")
```
And you'll have `Program` objects returned to use above instead of hashes.

zurbergram: changed

lib/csv_helper/loader.rb
```diff
@@ -19,9 +19,22 @@ def load(file, options = {})
...
       results = klass.import records, ignore: true, batch_size: Settings.active_record.batch_size.import
-      after_import_validations(records, results.failed_instances, options)
+
+      # using index of -1 since these rows failed during save to the table and not during after_import_validations
+      failed_instances = results.failed_instances
+                                .map { |result| { index: -1, message: result.display_errors_with_row } }
+      after_import_validations(records, failed_instances, options)
+
+      results.failed_instances = failed_instances
       results
```
jholton: I don't like that we've now changed `failed_instances` from an array of AR objects (e.g. `Program` objects) to an array of hashes with warning data at this point.  I know we don't actually use those objects today but it might trip up someone in the future.
I'd rather either monkey patch `ActiveRecord::Import::Result` or just return our warnings array with the `results`

```ruby
return [results, import_warnings]
```

I also think you should rename the array from `failed_instances` to something like `import_warnings` since they are not actually failed instances.

lib/csv_helper/loader.rb
```diff
@@ -19,9 +19,22 @@ def load(file, options = {})
...
       records = []
 
-      records = klass == Institution ? load_csv_with_version(file, records, options) : load_csv(file, records, options)
+      records = if klass == Program
+                  load_csv_with_row(file, records, options)
+                elsif klass == Institution
+                  load_csv_with_version(file, records, options)
+                else
+                  load_csv(file, records, options)
+                end
```
jholton: This is looking like a good candidate for `case`
```ruby
records = case klass
          when Program
             load_csv_with_row(file, records, options)
          when Institution
             load_csv_with_version(file, records, options)
          else
             load_csv(file, records, options)
          end
```

zurbergram: has to be `case name` and then use `.name` for when statements
changed

lib/csv_helper/loader.rb
zurbergram: I have renamed the array that is being passed around for `after_import_batch_validations` to `validation_warnings`.

I have changed the object that is in `results.failed_instances` and `validation_warnings` to ` be `{index: => i, record: => r}` so that the ActiveRecord object is still being returned within `results.failed_instances`

Is returning the ActiveRecord object as part of the objects in `result.failed_instances` ok instead of changing the return type of the method?

lib/csv_helper/loader.rb
```diff
@@ -19,9 +19,23 @@ def load(file, options = {})
     def load_records(file, options)
       records = []
 
-      records = klass == Institution ? load_csv_with_version(file, records, options) : load_csv(file, records, options)
+      records = case name
```
jholton: Should this be
```suggestion
      records = case klass.name
```
?

zurbergram: `klass` is `name.constantize` and comes from https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/lib/csv_helper/shared.rb#L22

jholton: Ah, I see, thanks.

jholton: Wait, I now notice that in that `Shared` module there is use of a `:row` key in errors.  I don't think we need any of this in regards to changing `failed_instances` and keeping track of a warning's index.  We can use this existing `:row` attribute.
Can you explain the idea behind the `index` in the warning hashes, so I'm sure I understand why you added it?

zurbergram: For Programs it is storing the `csv_row` value in the table and then retrieving the results from 2 SQL queries. Since 2 SQL queries are being used the warning messages won't be in order by csv row when displayed to the end user, currently only Program CSV has a `csv_row` column so an index property is still needed so that `loader` code can be generic.

jholton: They can be ordered by row.  See [here.](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/513/files#r354539143).
Working for all CSV types as far as I can tell.

zurbergram: I have returned `results.failed_instances` to returning ActiveRecord objects and instead put the index value as new property of the errors hash object and reference that for sorting purposes

jholton: I still don’t understand. Why do you have to sort on index when every error has a row number you can sort on?

zurbergram: Was trying to preserve the validation message "Line x" but after investigating output was actually "Row Line x : blah blah blah", removed `:index` property and change `:row` to be an integer to be sorted by

lib/csv_helper/loader.rb
```diff
@@ -19,9 +19,22 @@ def load(file, options = {})
...
+                when Institution.name
+                  load_csv_with_version(file, records, options)
+                else
+                  load_csv(file, records, options)
+                end
+
       results = klass.import records, ignore: true, batch_size: Settings.active_record.batch_size.import
+
+      # using index of -1 since these rows failed during save to the table and not during after_import_validations
+      results.failed_instances.each { |result| result.errors.add(:row, -1) }
```
jholton: This is not necessary.  All the `failed_instances` are assigned a row on line 96 of this file.
The only time that won't happen is for other `Program` validation failures because you return early from the method after running  `validator_klass.after_import_batch_validations`.
Remove this early return and those will be assigned row numbers also.  See [my PR](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/513/files)'s renamed `after_import_validations_and_row_assignment` method.
If you _do_ return early, the missing row number will be [treated as 'N/A'](https://github.com/department-of-veterans-affairs/gibct-data-service/blob/edcdd3ccdfb5ee802810010d9018c8061099a72e/lib/csv_helper/shared.rb#L16), and when you sort in the controller will still appear first.  Thus the row of `-1` is still unnecessary.

app/validators/program_validator.rb
```diff
@@ -1,17 +1,36 @@
 # frozen_string_literal: true
 
 class ProgramValidator < BaseValidator
-  def validate(record)
-    duplicates_exist = Program.where(['description = ? AND facility_code = ? AND id != ?',
-                                      record.description, record.facility_code, record.id]).any?
-    record.errors[:base] << non_unique_error_msg(record) if duplicates_exist
+  def self.after_import_batch_validations(validation_warnings)
```
jholton: Since we changed back to AR objects, can we rename this parameter `failed_instances`?

app/validators/program_validator.rb
```diff
@@ -1,17 +1,36 @@
...
 
-    facility_code_in_weam?(record)
+    missing_facility_in_weam.each do |record|
+      record.errors[:base] << missing_facility_error_msg(record)
+      add_warning_message(record, validation_warnings)
+    end
   end
 
-  private
+  def self.add_warning_message(record, validation_warnings)
```
jholton: Can we rename this method to something like `add_record_to_failed_instances` (and rename parameters as appropriate) since it's not actually adding a warning message?

zurbergram: renamed

app/validators/program_validator.rb
zurbergram: changed

lib/csv_helper/loader.rb
zurbergram: The early return in `after_import_validations` is because the goal is to either do batch validations or use Active Record Validations with a context of `after_import`

zurbergram: removed line and added `.to_i` to https://github.com/department-of-veterans-affairs/gibct-data-service/blob/a20c6de4e5fa5650f46bb56f0ac99bd2dd9cd128/app/controllers/uploads_controller.rb#L23

app/controllers/uploads_controller.rb
```diff
@@ -20,7 +20,8 @@ def create
       failed = load_csv.failed_instances
       @upload.check_for_headers
 
-      validation_warnings = failed.map(&:display_errors_with_row).sort
+      validation_warnings = failed.sort { |a, b| a.errors[:row]&.first.to_i <=> b.errors[:row]&.first.to_i }
```
jholton: The safe navigation operator is unnecessary.  You will always be calling `.first` on an array (i.e. no chance it can be `nil`)

zurbergram: removed

lib/csv_helper/loader.rb
jholton: > The early return in `after_import_validations` is because the goal is to either do batch validations or use Active Record Validations with a context of `after_import`

I just played around with the 400K row file and observed on my machine that it did add ~10 seconds to the request if you let it continue past the current `return`.  So I agree it's good to try to reduce/eliminate that.
The reason I was advocating for it is that if a `Program` fails an import validation (e.g. `validates :facility_code, presence: true`), then when the view loads the user will see 
> Row N/A : facility_code : can't be blank

Letting the `records.each_with_index` block run would populate the row for those objects, but it's not worth the time cost.  There is another solution that I will propose in another code comment momentarily.

app/validators/program_validator.rb
```diff
@@ -1,17 +1,36 @@
 # frozen_string_literal: true
 
 class ProgramValidator < BaseValidator
-  def validate(record)
-    duplicates_exist = Program.where(['description = ? AND facility_code = ? AND id != ?',
-                                      record.description, record.facility_code, record.id]).any?
-    record.errors[:base] << non_unique_error_msg(record) if duplicates_exist
+  def self.after_import_batch_validations(failed_instances)
```
jholton: At this point we have just entered the `after_import_batch_validations` method.
`failed_instances` contains all records that failed ActiveRecord validations.  For `Program` those are
```ruby
validates :program_type, inclusion: { in: InstitutionProgram::PROGRAM_TYPES }
```
This is a good opportunity to add the `row` to those records' `errors` before we start adding more records to `failed_instances` with our bulk validations.  This ensures that these failed instances will display a row number in their warning messages to the end user.
```suggestion
  def self.after_import_batch_validations(failed_instances)
    failed_instances.each do |record|
      record.errors.add(:row, record.csv_row)
    end 
```

zurbergram: added

zurbergram: This will only add `:row` values for the records that failed on ActiveRecord::Import#import for Program CSV

jholton: Correct, which is what `failed_instances` contains at this point in execution.

zurbergram: I have moved the call into `loader.rb` and it uses `try` so that all CSV types will get the `:row` hash property for rows that do not fail validations in `after_import_validations`

app/validators/program_validator.rb
```diff
@@ -1,17 +1,36 @@
 # frozen_string_literal: true
 
 class ProgramValidator < BaseValidator
```
jholton: Trying to figure out the best change to make here.  `ProgramValidator` doesn't need to be a subclass of `ActiveModel::Validator` anymore
`BaseValidator#facility_code_in_weam`is only used by `EduProgramValidator`, so it can be moved there.
I'm thinking:
* make `EduProgramValidator`'s parent class `ActiveModel::Validator`
* move `#facility_code_in_weam` to `EduProgramValidator`
* remove `ProgramValidator` parent class declaration
* change `BaseValidator` to `SharedMessages` with `self.missing_facility_error_msg` to be called by `EduProgramValidator` and `ProgramValidator`.  This part I'm least confident in, but can't think of a much cleaner way to organize it.
Thoughts?

lib/csv_helper/loader.rb
```diff
@@ -19,9 +19,23 @@ def load(file, options = {})
...
+                  load_csv_with_version(file, records, options)
+                else
+                  load_csv(file, records, options)
+                end
+
       results = klass.import records, ignore: true, batch_size: Settings.active_record.batch_size.import
+
+      results.failed_instances.each do |record|
+        record.errors.add(:row, record.try(:csv_row))
+      end
```
jholton: Why the move?  I think it should be moved back.
* This only works for `Program` objects
* Every other import has the row populated by at `after_import_validations`, line 97

app/validators/program_validator.rb
zurbergram: changed

lib/csv_helper/loader.rb
zurbergram: Moved this to only occur when making a call to validator's `after_import_batch_validations` method, this keeps the call generic and only runs it when needed

jholton: Yeah, this makes sense to me assuming more "batch validations" for big files is coming.

lib/csv_helper/loader.rb
```diff
@@ -99,6 +89,20 @@ def after_import_validations(records, failed_instances, options)
...
     end
 
+    def run_after_import_batch_validations?(failed_instances)
+      # this a call to custom batch validation checks for large import CSVs
+      validator_klass = "#{klass.name}Validator".safe_constantize
+      run_validations = validator_klass.present? && defined? validator_klass.after_import_batch_validations
+
+      if run_validations
+        failed_instances.each do |record|
+          record.errors.add(:row, record.try(:csv_row))
```
jholton: Should we change it from `record.try(:csv_row)` to `record.csv_row`?
Assuming more "batch validation" uploads/CSVs are going to be added, and they will all have a `csv_row` column on their associated tables, we want this to raise an exception if it doesn't exist (i.e. the column hasn't been added to the table).

zurbergram: changed

jholton (approved):
I think we're there!
Please check AC and Description of Done boxes as appropriate.

PR #510, opened 2019-12-04
Reuse preview version as Production version on push to production

hughes-dustin:
## Description
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19875

Previously, we were creating a new entry in GIDS version table when we pushed preview version to production. Instead of creating a new `production version`, we want to just set the `production` flag to true on the current `preview version`. 

## Testing done
Dev tested.

## Acceptance criteria
- [ ] The same data set that is created on preview generation is used as the published version.

## Definition of done


shawkey-daniel (approved):
LGTM

sonntag-adam (approved):

jholton:
Why the change?  Any idea why it was done this way in the first place?
What is the benefit?
Any drawbacks to the change?

hughes-dustin:
> Why the change? Any idea why it was done this way in the first place?
> What is the benefit?
> Any drawbacks to the change?

This is the first step to a larger effort to normalize the archive tables to their parent tables. The duplicating of versions makes it impossible for us to establish a proper parent/child relationship.

jholton:
1.

2.

3.

4.

5.

6.

app/controllers/dashboards_controller.rb
```diff
@@ -54,9 +54,9 @@ def push
     if version.blank?
       flash.alert = 'No preview version available'
     else
-      production_version = Version.create(number: version.number, production: true, user: current_user)
+      version.update(production: true)
 
-      if production_version.persisted?
+      if version.persisted?
```
jholton: This conditional will now return `true` 100% of the time since `version` is an already existing (persisted) `version`.

app/models/version.rb
```diff
@@ -28,6 +28,10 @@ def self.previous_production
   end
 
   def self.current_preview
+    if Version.current_production && Version.preview.newest.first.created_at < Version.current_production.created_at
+      return Version.current_production
```
jholton: This adds confusion.  The method is named `current_preview`, so I would expect it to always return a `preview` or perhaps `nil`

app/controllers/dashboards_controller.rb
hughes-dustin: I removed the condition check.

app/models/version.rb
hughes-dustin: The way it functioned previously was it would return the latest preview version whether if that version was a duplicate of the production or not. I figured that by refactoring it to return the production version when there are no new preview versions, we would still retain how `current_preview` was being used. This is necessary as `current_preview` returns the `updated_at` date we need to determine which uploads are being used for generating the new preview version. 

I could move that logic to `since_last_preview_version` in `upload.rb` instead of leaving it in `version.rb`. @jholton what are your thoughts?

app/models/version.rb
```diff
@@ -28,7 +28,9 @@ def self.previous_production
   end
 
   def self.current_preview
-    Version.preview.newest.first
+    return nil if current_production && preview.newest.first.created_at < current_production.created_at
+
+    preview.newest.first
```
jholton: Are there plans to clean up `versions`?  Just wondering if there will ever be a case where there are no preview `versions`.
I think a safe way to write this would be:
```ruby
  preview.newest.where('created_at > ?', current_production.created_at).first
```
Can there ever not be a `current_production`?

hughes-dustin: Upon building the db, we will always have one preview version and one current production `version`. So I don't think we will ever have a case where there are no preview or production `versions`.

I've put in your suggested change and tested it. It looks good to me.

jholton: Got it.  So then I thought, "why does seeding the data create two `versions`?".
I looked at `02_data.rb` and it is acting in the old style:  It creates a preview version.  Then it creates a production version with the same version number.
Can we change that so that it is consistent with this new paradigm and just updates the preview `version` to production?

jholton: Then this method definition would work with something like this:
```ruby
    cp = preview.newest
    cp = cp.where('created_at > ?', current_production.created_at) if current_production
    cp.first
```

hughes-dustin: That makes sense. Changes are in.

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -148,18 +148,14 @@ def load_table(klass, options)
       end
 
       context 'when successful' do
-        it 'adds a new version record' do
-          SiteMapperHelper.silence do
-            expect { post(:push) }.to change(Version, :count).by(1)
-          end
+        before do
```
jholton: Empty `before` block?

hughes-dustin: Sorry, left over code from when I was trying to fix broken tests. Changes are pushed in.

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -148,36 +148,31 @@ def load_table(klass, options)
       end
 
       context 'when successful' do
-        it 'adds a new version record' do
+        it 'sets the new production version' do
           SiteMapperHelper.silence do
-            expect { post(:push) }.to change(Version, :count).by(1)
+            expect { post(:push) }.to change(Version, :count).by(0)
```
jholton: Given that it `sets the new production version`, can we have an expectation to test that `version.production` changes from `false` to `true`?

spec/controllers/dashboards_controller_spec.rb
```diff
@@ -148,36 +148,31 @@ def load_table(klass, options)
...
+          allow(version).to receive(:update).and_return(version)
+          allow(version).to receive(:persisted?).and_return(false)
         end
 
-        it 'when is not successful returns an error message' do
-          allow(Version).to receive(:create).and_return(Version.new)
-          post(:push)
-          expect(flash.alert).to eq('Production data not updated, remains at previous production version')
+        it 'does not add a new version' do
+          expect { post(:push) }.to change(Version, :count).by(0)
```
jholton: Not sure what we're testing here.
`Version.count` doesn't change if it's successful _or_ fails.
How can it fail?  
`version.update(production: true)` should never fail, and if it does it will fail silently.
Looks like only if `Version.current_preview` is `blank?` does our outcome change.
I think we should just test that path.  
What do you think?

spec/controllers/dashboards_controller_spec.rb
hughes-dustin: Updated the test to expect that `current_production` is nil before `push()`, and that `current_production` is equal to `current_preview` and that `production` is `true`.

spec/controllers/dashboards_controller_spec.rb
hughes-dustin: I've pulled out the test. It looks like we are already covering that condition on `line 137`.

spec/models/gibct_site_mapper_spec.rb
```diff
@@ -3,17 +3,17 @@
...
   let(:production_institution_fc) { '00000002' }
 
   before do
+    create :version, :production
+    version = create :version, :preview
     File.delete(sitemaps_path) if File.exist?(sitemaps_path)
-
     %i[preview production].each { |p| create :version, p }
-    create :institution, version: preview_version.number, facility_code: preview_institution_fc
+    create :institution, version: version.number, facility_code: preview_institution_fc
```
jholton: Now it looks like we are creating unused `versions`.  2 on lines 12, 13 and 2 on line 15.
I think if you undo all these changes and just swap the creation order of `preview` and `production` on line 15, everything will work.
Remember, your `current_preview` has to be newer than your `current_production`. So change the creation order and that state will be achieved.

hughes-dustin: Changes are in.

jholton (approved):
LGTM

PR #511, opened 2019-12-04
Update schema.rb to be consistent with staging and production

annaswims:
connected to staging and production servers and ran 'bundle exec rails db:schema:dump' the output was identical and did not match what we have in master

I'm updating schema.rb to accurately reflect what is in staging/production


LindseySaari (approved):
👍

PR #519, opened 2019-12-13
Bah 4063 Database Specific Changes 

elder-joseph:
## Description
As a developer, I need to add a parent / child relationship so that the database is more normalized.

This PR is specific to the GIDS db source code, there is a separate PR associated with the app changes for 4063: #518 

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/4063

## Testing done
Testing passes locally 

## Screenshots
N/A

## Acceptance criteria
- [x] Version table is created.
- [x] Has_many relationship is established from version to institution.
- [x] Has_many relationship is established from version to zipcode_rate.
- [x] Archiver is configured to archive institution and zipcode_rate data.

## Definition of done
- [x] Events are logged appropriately
- [x] Documentation has been updated, if applicable
- [x] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [x] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #522, opened 2019-12-16
Crosswalk Issues DB Migration

devin-mccurdy:
## Description

Create the `crosswalk_issues` table and indexes that will be used when creating `crosswalk_issues` records.

This table is currently only a cross-reference table, but will be updated in an upcoming issue to include user-entered "fixes" that will be leveraged by `InstitutionBuilder`.

Migration PR for: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/523

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/4041

## Testing done
Tested locally and QA review

## Screenshots
N/A

## Acceptance criteria
- [x] `crosswalk_issues` table is created

## Definition of done


zurbergram (approved):

shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

db/migrate/20191219210002_create_crosswalk_issues.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCrosswalkIssues < ActiveRecord::Migration[5.2]
+  def change
+    create_table :crosswalk_issues do |t|
+      t.integer :weam_id
+      t.integer :crosswalk_id
+      t.integer :ipeds_hd_id
```
jholton: These are foreign keys, right?
Might I suggest:
```ruby
      t.belongs_to :weam, foreign_key: true
      t.belongs_to :crosswalk, foreign_key: true
      t.belongs_to :ipeds_hd, foreign_key: true
```

db/schema.rb
```diff
@@ -521,7 +521,7 @@
     t.boolean "stem_indicator", default: false
     t.string "campus_type"
     t.string "parent_facility_code_id"
-    t.bigint "version_id"
+    t.integer "version_id"
```
jholton: Are these accidental reversions?  I _think_ it should be `bigint`

db/schema.rb
```diff
@@ -1630,12 +1630,13 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.integer "version"
-    t.bigint "version_id"
+    t.integer "version_id"
```
jholton: Are these accidental reversions?  I _think_ it should be `bigint`

dmccurdybah: Yes, accidental, but I'm not sure the cause of this... I don't think this schema matched the migrations.

dmccurdybah: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/519/files

It's added [here](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/519/files) as an `integer`.  I think the intent of the PR was to add a reference between `institution_archives` and `versions` but the migration doesn't actually do that (but the schema does)

jholton: Ugh, you're right.  I see what happened.
First, as background, since Rails 5.1, [primary and foreign keys have been `bigint` instead of `integer`](https://github.com/rails/rails/pull/26266)
In the linked PR ([this commit](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/519/commits/dda2012bc87cc043b30c6c075b938cecac37196c)), @elder-joseph added foreign keys to the two 'archive' tables in question.
Then I [approved the PR](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/519#pullrequestreview-337280796).
Then @elder-joseph [changed the migrations to now just add the foreign key field as an `interger`](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/519/commits/b8e7c02818b6bd64fd3256a70e625b16c5be7f66).  I understand why he did this: it has been decided that the 'archive' tables don't need indexes.
The problem is, he then merged the PR without re-running the migrations after the change.
Thus, the schema specifies `bigint` and the migration specifies `integer`.
So when you @dmccurdybah ran the migrations on your machine it changed the field type to `integer` because that's what the migration specifies.
So now we have a potential problem.  Has the migrations in question been run in deployed `dev`, `staging`, and/or `production`? 

**Please**, if you push changes after a PR approval, re-request a review before you merge.

dmccurdybah: Thank, yeah, I understand the issue and this definitely shouldn't have happened.

The migrations haven't been ran on `dev`, `staging` or `production` yet. Is it acceptable to you if we update the existing migrations in a separate PR?

jholton: Yeah, that's fine with me since we're a small team.  Will just have to make sure all the other engineers are aware so they can ensure that their local databases are accurate.

jholton (approved):
LGTM

PR #523, opened 2019-12-16
Crosswalk issues list

devin-mccurdy:
## Description

Find and display Crosswalk Issues in a single page list. This will be used to check Weams data against Education's IpedsHD data that doesn't have an existing Crosswalk record linking them.  The intention is that end users who maintain the Crosswalk file will be able to use this list to assist in finding missing crosswalk linkages. Fixing these issues will be a manual process until we design and implement a solution which will allow them to resolve these issues directly from the UI.

The new `crosswalk_issues` table is rebuilt whenever the Weams, Crosswalks, or IpedsHD file is imported

Migration PR: Migration PR for: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/522

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/4041

## Testing done
Tested locally and QA review

## Screenshots
image

## Acceptance criteria
- [x] Displays the following fields for any flagged crosswalk issues:
    - Weams Institution Name
    - Weams Facility Code
    - Weams ipeds
    - Weams ope
    - IpedsHd unitid
    - IpedsHd opeid
    - Crosswalk ipeds
    - Crosswalk ope
- [x] Crosswalk Issue records are created when there is any of the following conditions:
    - There is an exact match (non-case sensitive) in the "unitd" OR "opeid" of the "IpedsHD" file and the "iped" OR "ope" of the "Crosswalk" file
    - There is an exact match (non-case sensitive) in the "unitd" OR "opeid" of the "IpedsHD" file and the "iped" OR "ope" of the "WEAMS" file

## Definition of done


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

app/controllers/crosswalks_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class CrosswalksController < ApplicationController
```
jholton: Shouldn't this (and the corresponding views) be `crosswalk_issues` instead of `crosswalks`?

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class CrosswalkIssue < ApplicationRecord
+  belongs_to :crosswalk
+  belongs_to :ipeds_hd
+  belongs_to :weam
```
jholton: Don't forget the corresponding `has_many` declarations.

app/controllers/crosswalks_controller.rb
dmccurdybah: The list displayed is of crosswalk_issues, but the plan for this page is to allow the end user to create crosswalk records, which will be the eventual replacement of the crosswalk CSV (which also resolves the crosswalk_issues).

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  # rubocop:disable Metrics/MethodLength
+  def self.rebuild
+    sql = <<-SQL
```
jholton: This makes my brain hurt.  But before I dig in:
> There is an exact match (non-case sensitive) in the "instnm" of the "IpedsHD" file and the "institution name" of the "Weams" file

Is this scenario addressed in the SQL?

config/routes.rb
```diff
@@ -21,6 +21,8 @@
     get '(:csv_type)' => 'uploads#new', on: :new, as: ''
   end
 
+  get '/crosswalks' => 'crosswalks#index'
```
jholton:
```suggestion
  resources :crosswalks, only: :index
```

app/models/crosswalk_issue.rb
dmccurdybah: Sorry, this was removed from the story AC in testing and I forgot to remove it from the PR after it'd been removed from the code.

app/controllers/uploads_controller.rb
```diff
@@ -87,12 +87,18 @@ def call_load
     file = @upload.upload_file.tempfile
     skip_lines = @upload.skip_lines.try(:to_i)
     col_sep = @upload.col_sep
+
+    CrosswalkIssue.delete_all if [Crosswalk, IpedsHd, Weam].include?(klass)
```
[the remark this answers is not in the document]
dmccurdybah: This has to be done before `klass.load` it attempts to delete the contents of the uploading model table, which throws an exception because we added foreign keys to `crosswalk_issues`.

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,46 @@
...
+          weams.cross IS NOT NULL
+          AND ipeds_hds.cross IS NOT NULL
+          AND crosswalks.cross IS NOT NULL
+          AND weams.cross = ipeds_hds.cross
+          AND weams.cross = crosswalks.cross
+        )
+        AND NOT(
+          weams.cross IS NULL
+          AND weams.ope IS NULL
+          AND ipeds_hds.ope IS NULL
```
jholton: if `weams.cross` is `null` and `weams.ope` is `null` then `ipeds_hds.ope` will be `null` 100% of the time since the `JOIN` would have failed.  Thus, you don't need this line.

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,46 @@
...
+          AND crosswalks.cross IS NULL
+          AND crosswalks.ope IS NULL
+        )
+        AND NOT(weams.campus_type IS NOT NULL AND UPPER(weams.campus_type) = 'E')
```
jholton: I don't think you have to check for `NULL` first.
```suggestion
        AND NOT UPPER(weams.campus_type) = 'E'
```

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,46 @@
...
+        weams.id,
+        crosswalks.id,
+        ipeds_hds.id
+      FROM weams
+        LEFT OUTER JOIN ipeds_hds ON  weams.cross = ipeds_hds.cross
+          OR weams.ope = ipeds_hds.ope
+        LEFT OUTER JOIN crosswalks ON weams.facility_code = crosswalks.facility_code
+      WHERE
+        (institution_of_higher_learning_indicator = true OR non_college_degree_indicator = true)
+        AND NOT(
```
jholton: Can you tell me what these `AND NOT` conditionals are doing in relation to the Acceptance Criteria?  

Are we eliminating valid records as opposed to selecting invalid records?

dmccurdybah: We're looking for partial matches only, so we are excluding cases where:

- `weams` records fully matches across `ipeds_hds`, and `crosswalks`
- `weams` records without any partial matches across `ipeds_hds`, and `crosswalks`

If you're questioning the story AC I can ask to have it updated to reflect this.

jholton: I was just trying to mentally map the AC to the SQL and having trouble.

app/models/crosswalk_issue.rb
dmccurdybah: Yes, the join would have failed, but because it's an outer join, the result would still include those `weams` rows that failed to join unless we exclude them.

app/models/crosswalk_issue.rb
dmccurdybah: Yes, I completely understand how this might seem a little convoluted but we're trying to get the customer a view into this data so they can more easily review the "issues" as first step towards starting the discussion about how these might be resolved. Right now this is a manual process on their and and we're trying to be as "agile" as we can.

There's also a follow up PR that is waiting on this one which adds some additional cases.

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,46 @@
...
+          weams.cross IS NOT NULL
+          AND ipeds_hds.cross IS NOT NULL
+          AND crosswalks.cross IS NOT NULL
+          AND weams.cross = ipeds_hds.cross
+          AND weams.cross = crosswalks.cross
+        )
```
jholton: In English, this `AND NOT` removes all results where `weams.cross` is populated AND `ipeds_hds.cross` is populated AND `crosswalks.cross ` is populated AND all three values are the same.  ALL of those conditions have to be true.

What about the scenario where  your LEFT OUTER JOINS joined on no `ipeds_hds` or `crosswalks` records.  In that case you'd have a `null` `ipeds_hds.cross` and `null` `crosswalks.cross`, and those results would stay included.  Your `crosswalks_issues` table would end up with a record that had a `weams_id` populated but a `null` `crosswalks_id` and `null` `ipeds_hds_id`.
Is that OK?

Or do you want:
```SQL
AND NOT(
          weams.cross IS NOT NULL
          AND weams.cross = ipeds_hds.cross
          AND weams.cross = crosswalks.cross
        )
```

app/models/crosswalk_issue.rb
dmccurdybah: It looks like this was necessary and I'm going to have to figure out why =)

dmccurdybah: Ah, we do want to allow nulls, but explicitly exclude `campus_type` of 'E' which we have to check for `null`.

This seems clearer

```SQL
AND (
          weams.campus_type IS NULL 
          OR (weams.campus_type IS NOT NULL AND UPPER(weams.campus_type) != 'E')
        )
```

app/models/crosswalk_issue.rb
dmccurdybah: Yes, this is what they've asked for because it's an "issue" they'd need to resolve.

app/models/crosswalk_issue.rb
jholton: Right.  I'm only talking about line 36.  Line 34 and 35 take care of the `weams` rows.
It's a nitpick: I'm just saying that (I think) you will get the same results with or without line 36, because you will never have a result where `weams.cross` is `null` and `weams.ope` is `null` but `ipeds_hds.ope` is present. So you don't need to check for the value of `ipeds_hds.ope`.  Again, not a big deal.

app/models/crosswalk_issue.rb
```diff
@@ -0,0 +1,46 @@
...
+        AND NOT(
+          weams.cross IS NULL
+          AND weams.ope IS NULL
+          AND ipeds_hds.ope IS NULL
+          AND crosswalks.cross IS NULL
+          AND crosswalks.ope IS NULL
+        )
+        AND (weams.campus_type IS NULL OR UPPER(weams.campus_type) != 'E')
```
jholton: Nitpick: `is distinct from` exist for this use case.
> not equal, treating null like an ordinary value

[link](https://www.postgresql.org/docs/11/functions-comparison.html)
```suggestion
        AND UPPER(weams.campus_type) is distinct from 'E'
```

jholton (approved):
LGTM. Thanks for helping me understand the goals wit the SQL.

PR #534, opened 2020-01-02
GIDS: Use Scorecard API to replace flat-file upload

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/4380

As an EDU team member, I would like GIDS to pull scorecard data via API so I no longer have to upload a CSV into GIDS.
## Supporting Artifacts: 
1. The API and related documentation can be found here: https://collegescorecard.ed.gov/data/documentation/

## Testing done
- local testing

## Screenshots
Screen Shot 2020-01-16 at 2 51 52 PM

Screen Shot 2020-01-16 at 2 48 08 PM

Screen Shot 2020-01-16 at 2 51 48 PM

## Acceptance criteria
- [ ] 1. The user can manually update the Scorecard data via API.
- [ ] 2. The data retrieved via API matches the data that is uploaded manually.
- [ ] 3. The API key is configurable.

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #536, opened 2020-01-06
Add lib/common/client code from vets-api (no REDIS) to use for API calls within GIDS

zurbergram:
## Description
This PR is just for adding `lib/common/client` code without Redis needed for 
https://github.com/department-of-veterans-affairs/va.gov-team/issues/4380
The code was copied from https://github.com/department-of-veterans-affairs/vets-api/tree/master/lib/common

Main PR is https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534

## Acceptance criteria
- [ ]

## Definition of done


shawkey-daniel (approved):
LGTM

kfrz:
@zurbergram if you're going to wholesale lift an entire `lib/` subdir the *least* you might want to consider is copying the specs over.


I am going to suggest *not* approving this for a few major reasons:

- There are no specs added at all to illustrate what code changes will do in this PR
- There are no acceptance criteria to explain why you need all the library
- The `vets-api` `lib/` folder is certainly robust, though that doesn't mean it's appropriate for every use case, especially when it comes to a distinct application.


I'd be happy to help break down some first-principles with y'all if you are finding implementation of an API client wrapper to be a challenge. Let me know how we can help.

kfrz (changes requested):
See comment on main thread. I would like to see a) specifications and b) not a wholesale copy-paste of what is potentially useless code in your context.

zurbergram:
> @zurbergram if you're going to wholesale lift an entire `lib/` subdir the _least_ you might want to consider is copying the specs over.
> 
> I am going to suggest _not_ approving this for a few major reasons:
> 
> * There are no specs added at all to illustrate what code changes will do in this PR
> * There are no acceptance criteria to explain why you need all the library
> * The `vets-api` `lib/` folder is certainly robust, though that doesn't mean it's appropriate for every use case, especially when it comes to a distinct application.
> 
> I'd be happy to help break down some first-principles with y'all if you are finding implementation of an API client wrapper to be a challenge. Let me know how we can help.

I was going to do the specs in a separate PR due to the size of this one

Updated description of https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534

Since vets-api already had a built out library for making rest calls using faraday and gibct-data-service now needs the same capabilities I figured copying over an already functioning library that is frequently used was the best approach

zurbergram:
> See comment on main thread. I would like to see a) specifications and b) not a wholesale copy-paste of what is potentially useless code in your context.

Do you want me to go through this PR and only include files that are directly used by code in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534 ?

zurbergram:
Did `git reset --hard master` to reset branch so commit history with all the things was removed, this is the minimum code from https://github.com/department-of-veterans-affairs/vets-api/tree/master/lib/common/client and https://github.com/department-of-veterans-affairs/vets-api/tree/master/lib/common/exceptions needed to perform rest call in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534

zurbergram:
@kfrz I have only copied over what is needed (along with existing vets-api unit tests) for https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534 to function

zurbergram:
@department-of-veterans-affairs/backend-review-group This PR is ready for a re-review

lib/common/client/base.rb
```diff
@@ -0,0 +1,152 @@
...
+          # not using breakers for Redis at this time
+          # if handlers.include?(Breakers::UptimeMiddleware)
+          #  return connection if handlers.first == Breakers::UptimeMiddleware
+          #
+          #  raise BreakersImplementationError, 'Breakers should be the first middleware implemented.'
+          # else
+          #  warn("Breakers is not implemented for service: #{config.service_name}")
+          # end
+
+          connection
```
kfrz: If you're not using code, delete it rather than comment it out will make for a cleaner history..

lib/common/client/base.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+
+      def request(method, path, params = {}, headers = {}, options = {}) # rubocop:disable Metrics/MethodLength
+        sanitize_headers!(method, path, params, headers)
+        raise_not_authenticated if headers.keys.include?('Token') && headers['Token'].blank?
+        connection.send(method.to_sym, path, params) do |request|
+          request.headers.update(headers)
+          options.each { |option, value| request.options.send("#{option}=", value) }
+        end.env
+      rescue Common::Exceptions::BackendServiceException => e
```
kfrz: Can you explain the reason you need to coerce your API's service exceptions?

kfrz (approved):
Approving - there's nothing particularly dangerous about this changeset, though I would probably choose to approach the objective with a different tactic that introduces fewer potentially unused codepaths.

lib/common/client/configuration/base.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        # @!group Class Attributes
+        # @!attribute [rw]
+        # headers to include in all requests
+        class_attribute :base_request_headers
+
+        self.open_timeout = 15
+        self.read_timeout = 15
+        self.request_types = %i[get put post delete].freeze
+        self.user_agent = 'Vets.gov Agent'
```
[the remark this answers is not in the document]
zurbergram: Changed to "VA.gov Agent"

lib/common/client/base.rb
zurbergram: removed

lib/common/client/base.rb
zurbergram: This is same as behavior at https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/base.rb#L107

This specific rescue creates or uses a ServiceException in the module of class making the request which allows for easier debugging when an issue occurs as the exception then appears to be coming from the correct module instead of a generic one

lib/common/client/middleware/response/json_parser.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Common
+  module Client
+    module Middleware
+      module Response
+        class JsonParser < Faraday::Response::Middleware
+          WHITESPACE_REGEX = /\A^\s*$\z/.freeze
+          MHV_SUCCESS_REGEX = /^success/i.freeze
```
jholton: You're not communicating with MHV, right?

lib/common/client/base.rb
```diff
@@ -0,0 +1,134 @@
...
+    # Raised if the Faraday adapter is a Faraday::Adapter::HTTPClient and the
+    # Common::Client::Middleware::Request::RemoveCookies middleware in not included.
+    #
+    class SecurityError < StandardError
+    end
+
+    ##
+    # Raised when the breakers Faraday middleware is not first in the stack.
+    #
+    class BreakersImplementationError < StandardError
```
jholton: GIDS doesn't use `breakers`.
There are a few other `breakers` references in the PR too.

jholton (commented):
Noticed a couple unneeded pieces.

lib/common/client/middleware/response/json_parser.rb
zurbergram: Removing in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534

lib/common/client/base.rb
zurbergram: Removing in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/534

PR #542, opened 2020-01-14
Bah info box issues: extra bullet point, wrong value in the Column Separator drop-down on upload results in "Something Went Wrong" error message

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/4817
https://github.com/department-of-veterans-affairs/va.gov-team/issues/4818

## Testing done
- local testing

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -121,8 +121,8 @@ class UploadsController < ApplicationController
   def requirements_messages
     messages = validation_messages
     # this a call to custom validators that are not listed inside the class
-    custom_validator_messages = "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
-    messages.push(*custom_validator_messages)
+    custom_batch_validator_messages = "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
+    messages.push(*custom_batch_validator_messages)
 
     messages.uniq # remove duplicates for objects that have validate: #{klass.name}Validator
   end
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -7,9 +7,8 @@ class UploadsController < ApplicationController
 
   def new
     @upload = new_upload(params[:csv_type])
-    csv_type_check = @upload.csv_type_check?
-
-    if csv_type_check
+    
+    if @upload.csv_type_check?
       @requirements = requirements_messages
       return
     end
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -120,15 +120,6 @@ class UploadsController < ApplicationController
   end
 
   def requirements_messages
-    messages = validation_messages
-    # this a call to custom validators that are not listed inside the class
-    custom_validator_messages = "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
-    messages.push(*custom_validator_messages)
-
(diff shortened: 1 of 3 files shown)


devin-mccurdy (approved):
LGTM

sonntag-adam (approved):

delli-gatti-michael (approved):
lgtm

app/controllers/uploads_controller.rb
```diff
@@ -125,7 +124,7 @@ def requirements_messages
     custom_validator_messages = "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
     messages.push(*custom_validator_messages)
```
jholton: These aren't necessary anymore are they?  If you remove them, then you won't be inserting duplicate messages and you won't need to call `.uniq` on `messages` on line 127, right?

app/controllers/uploads_controller.rb
```diff
@@ -139,8 +138,14 @@ def validation_messages
         generic_requirement_message('These columns can only contain numeric values: ', validations)
       when ActiveRecord::Validations::UniquenessValidator
         generic_requirement_message('These columns should contain unique values: ', validations)
+      when "#{klass.name}Validator".safe_constantize
+        "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
+      else
+        validations.class.to_s
```
jholton: Why?  Do you want to just print the name of a validator class to the screen?

app/controllers/uploads_controller.rb
zurbergram: https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/weam.rb and https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/program.rb need this line

https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/edu_program.rb is the exception which when you viewed this CSV was creating the empty bullet point

app/controllers/uploads_controller.rb
zurbergram: Adding default behavior, if removing this might have to add a `.select(&:present)` to messages in `requirements_messages` to remove blank lines

jholton: Isn't having to `.select(&:present)` better than ending up with a bullet point in the view that says something like 
* ActiveRecord::Validations::WhateverValidator

That's what you'll get if the `else` path occurs.

app/controllers/uploads_controller.rb
```diff
@@ -139,8 +138,14 @@ def validation_messages
...
       end
-    end
+    end.flatten
```
jholton: Basically, I think you can undo all these `requirements`-related changes and change this line to 
```suggestion
    end.compact
```
And you're good to go.  No having to deal with duplicated messages, blank bullet points or arcan 'ActiveRecord::Validations::WhateverValidator' bullet points appearing.

jholton: To clarify you just need `.compact` instead of `.select(&:present?)` because the default response from a `case` statement is `nil`.  `.compact` removes `nil`s from an array.

zurbergram: changed

app/controllers/uploads_controller.rb
zurbergram: removed

jholton (approved):

PR #545, opened 2020-01-22
Update CI Docker image to use postgresql 9.6

johnpaulashenfelter:
## Description

The production environment is moving to Postgresql 9.6 over the coming 30 days. Running CI on 9.6 is a simple first step to ensure there are no surprises in the code from the changeover.

## Testing done

Will verify the container build is successfully running 9.6.

## Acceptance criteria
- [x] CI server uses postgres 9.6 server instead of 9.4

## Definition of done


annaswims:
🎉 Should we wait until right before the upgrade to merge this, just in case someone tries to use a 9.6 only feature prematurely? Is that even a thing?  I don't remember what 9.4->9.6 gives us.

johnpaulashenfelter:
@annaswims there's a *lot* of point releases to read through notes, but a good summary is https://wiki.postgresql.org/wiki/NewIn96 and https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.5

One strategy would be keep 9.4 on *dev* machines too to avoid using anything. 9.6 running on CI catches any issues btw 9.4 (local) and 9.6 (CI). In any case, even just keeping this branch up-to-date runs the code against 9.6

zurbergram:
Should the following also be updated:
- https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/docker-compose.yml 
- https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/config/initializers/strong_migrations.rb#L3

johnpaulashenfelter:
@zurbergram I Believe the docker-compose is used for starting up the app on staging/production (and dev if you use it there) so it doesn't have much to do with the CI environment.

The goal here is to have *CI* running on 9.6 to make sure your specs continue to pass and to hopefully give you a warning from CI if you do something on 9.4 that's not going to work on 9.6 (unlikely, but remotely possible).

CI already successfully passes your test suite since this branch builds successfully. I checked that 9.6 _was_ the version being run on this branch as expected and it is.

I'm 100% open to this team managing the transition to pg9.6 however you want, knowing that at some point in the next ~30d *production* switches over.

My advice is to get CI running 9.6 as an early warning system. And I'd consider running any significant new raw SQL query you build in both 9.4 and 9.6 if you want to be 100% comfortable -- that may or may not affect any work you have in the time period we're talking about.

Depending on how you do your development, you can run both 9.4 and 9.6 on the same machine through the Postgres.app or `homebrew` on a Mac and its pretty easy to swap the version in the docker container (though time consuming for the downloads). Or one of your team can run 9.4 and another 9.6 -- lots of options.

But in any case, based on the *specs* from the CI build for this branch/PR you're fine for 9.6 right now

johnpaulashenfelter:
Closing since long-since handled.

PR #550, opened 2020-01-28
Bah 4038 Refactor archiver to handle parent/child relationships

elder-joseph:
## Description
As a developer, I need to refactor the version archiver to handle data that is nested in a parent / child relationship so that additional enhancements can be made.

https://github.com/department-of-veterans-affairs/va.gov-team/issues/4038

App and Database changes.

Database specific changes for 4038 can be found here:
#551 (merged)
#571 (merged)
#576 (merged)

## Testing done
All CI tests pass locally, along with dev testing of functionality.

## Screenshots
N/A

## Acceptance criteria
- [x] Archiver is configurable to archive tables with version id.
- [x] Archiver is configurable to archive tables modeled as children to tables with version id.
- [x] Archiver moves previous versioned data to configured corresponding archive tables.

## Definition of done


.rubocop.yml
```diff
@@ -6,18 +6,18 @@ require:
 
 AllCops:
   Include:
-    - '**/Gemfile'
-    - '**/config.ru'
-    - '**/Rakefile'
-    - '**/*.rake'
-    - '**/*.rb'
+    - "**/Gemfile"
```
zurbergram: Is this a result of a setting in VSCode causing this to happen?

app/models/gibct_site_mapper.rb
```diff
@@ -32,8 +32,8 @@ def generate_sitemap(version)
     SitemapGenerator::Sitemap.create do
       add '/search', priority: 0.9, changefreq: 'monthly'
 
-      Institution.version(version).find_each(batch_size: Settings.active_record.batch_size.find_each) do |institution|
-        add "/profile/#{institution.facility_code}", priority: 0.8, changefreq: 'weekly'
+      Institution.version_id(version).find_each(batch_size: Settings.active_record.batch_size.find_each) do |i|
```
zurbergram: why rename the variable?

.rubocop.yml
elder-joseph: I believe so, will try disabling my plugin for formatting

app/models/gibct_site_mapper.rb
elder-joseph: This was renamed to avoid another rubocop length instance

db/migrate/20200107150000_add_institution_columns_to_school_certifying_officials.rb
```diff
@@ -0,0 +1,5 @@
+class AddInstitutionColumnsToSchoolCertifyingOfficials < ActiveRecord::Migration[5.2]
```
zurbergram: Where is the database specific PR?

db/migrate/20200107150001_create_school_certifying_officials_archives.rb
```diff
@@ -0,0 +1,15 @@
+class CreateSchoolCertifyingOfficialsArchives < ActiveRecord::Migration[5.2]
```
zurbergram: Where is the database specific PR?

db/migrate/20200107150004_remove_institutions_version_constraint.rb
```diff
@@ -0,0 +1,8 @@
+
```
zurbergram: Where is the database specific PR?

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -55,7 +55,7 @@ def preview_body(body)
   context 'with autocomplete results' do
     it 'returns collection of matches' do
       create(:version, :production)
-      create_list(:institution, 2, :start_like_harv)
+      create_list(:institution, 2, :start_like_harv, version_id: Version.current_production.id)
```
zurbergram: Can you add this to the relevant factory instead of here and other places in this file?

spec/models/institution_tree_spec.rb
```diff
@@ -3,21 +3,23 @@
 require 'rails_helper'
 
 RSpec.describe InstitutionTree, type: :model do
-  let(:version) { create(:version, :production) }
+  let(:version) { Version.current_preview }
```
zurbergram: why change this? or not use 
```
create (:version, :preview)
``` 
?

spec/models/institution_tree_spec.rb
```diff
@@ -3,21 +3,23 @@
...
-      create(:institution, facility_code: '103', parent_facility_code_id: '100', campus_type: 'N')
-      create(:institution, facility_code: '104', parent_facility_code_id: '102', campus_type: 'E')
-      create(:institution, facility_code: '105', parent_facility_code_id: '103', campus_type: 'E')
-      create(:institution, facility_code: '106', parent_facility_code_id: '103', campus_type: 'E')
-      create(:institution, facility_code: '107', parent_facility_code_id: '103', campus_type: 'E')
-      create(:institution, facility_code: '108', parent_facility_code_id: '100', campus_type: 'E')
-      create(:institution, facility_code: '109')
-      create(:institution, facility_code: '110', parent_facility_code_id: '100', campus_type: 'E', version: 2)
+      create :version, :production
+      create(:institution, facility_code: '100', campus_type: 'Y', version_id: Version.current_production.id)
```
zurbergram: Can you add this to the relevant factory instead of here and other places in this file?

spec/models/institution_spec.rb
```diff
@@ -148,20 +148,11 @@ def create_institution(institution, description)
...
-    context 'with version' do
-      it 'retrieves institutions by a specific version number' do
-        i = create_list :institution, 2, version: 1
-        j = create_list :institution, 2, version: 2
-
-        expect(described_class.version(i.first.version)).to match_array(i.to_a)
-        expect(described_class.version(j.first.version)).to match_array(j.to_a)
-      end
-
-      it 'returns blank if a nil or non-existent version number is supplied' do
```
zurbergram: why remove this? should this be reworked to be similar test but for version_id instead?

app/controllers/v0/institution_programs_controller.rb
```diff
@@ -9,7 +9,7 @@ def autocomplete
       @data = []
       if params[:term].present?
         @search_term = params[:term]&.strip&.downcase
-        @data = InstitutionProgram.version(@version[:number]).autocomplete(@search_term)
+        @data = InstitutionProgram.autocomplete(@search_term)
```
zurbergram: since this is being removed should `InstitutionProgram::autocomplete` be updated to account for the version?

app/models/institution_builder.rb
```diff
@@ -603,22 +643,27 @@ def self.build_versioned_school_certifying_official(version_number)
...
+      INNER JOIN institutions i ON s.facility_code = i.facility_code
+      INNER JOIN versions v ON v.number = ?
+      WHERE v.id = i.version_id
+      AND s.institution_id IS NOT NULL;
     SQL
+    sql = SchoolCertifyingOfficial.send(:sanitize_sql, [str, version_number])
+    SchoolCertifyingOfficial.connection.execute(sql)
+  end
 
+  def self.add_sco_institution_id(version_number)
```
zurbergram: unit test?

spec/models/institution_builder_spec.rb
```diff
@@ -948,13 +949,5 @@ def check_ipeds_ic_ay
         expect(institutions.find_by(facility_code: '1VZZZZZZ').approved).to be_truthy
       end
     end
-
-    describe 'when generating versioned school certifying official table' do
```
zurbergram: should this test be rewritten to account for the new behavior of `build_school_certifying_officials` instead of being removed?

app/controllers/v0/institution_programs_controller.rb
elder-joseph: Any suggestions on the best way to do so here?

zurbergram (approved):

delli-gatti-michael (approved):

delli-gatti-michael:
Local testing complete and PR approved

PR:
https://github.com/department-of-veterans-affairs/gibct-data-service/pull/550#pullrequestreview-352425918

app/controllers/v0/api_controller.rb
```diff
@@ -20,6 +20,7 @@ def resolve_version
 
       @version = {
         number: version.number,
+        id: version.id,
```
jholton: Do we need this in response data?

jholton: If not, [here's](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/558) a refactor to separate a working `@version` object from the metadata.  With that you can just use `@version.id` wherever you need to, and `version` details in metadata stays the same.

app/controllers/v0/institution_programs_controller.rb
```diff
@@ -9,7 +9,9 @@ def autocomplete
       @data = []
       if params[:term].present?
         @search_term = params[:term]&.strip&.downcase
-        @data = InstitutionProgram.version(@version[:number]).autocomplete(@search_term)
+        @data = InstitutionProgram.joins('INNER JOIN versions v ON v.id = institutions.version_id')
+                                  .where('v.number = ?', @version.number)
+                                  .autocomplete(@search_term)
```
jholton: Why make this change?  Are you planning to remove `version` from `institution_programs`?

elder-joseph: @jholton by leveraging the `version_id` of the parent institution, `version` is no longer really necessary for `institution_programs`. This would also be the case for `school_certifying_officials`, which has a similar relationship to `institution`. 

It seems that `version` could be removed from both, but I wasn't entirely sure if it would be a good idea to add a migration for its removal.

jholton: Makes sense.  I think it's a good idea to remove the column when we're sure it's not being used anywhere.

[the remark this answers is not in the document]
elder-joseph: Not a problem, I had a chance to look over the PR and it seems to make sense to do it that way.

elder-joseph: I agree, it probably is best to go ahead and add a migration to remove that unused column. I will also have to do some clean up on the unit tests which use version there.

elder-joseph: Hi @jholton, after discussing this with some of the folks on my team, we decided it would be best to go ahead and plan to do this in a separate issue where we would also remove the other deprecated `version` columns.

jholton: Sounds good to me.

jholton (commented):
I'm not all the way through it, but I have to step away from my computer.  Here are some comments so far.

app/controllers/v0/institutions_controller.rb
```diff
@@ -44,7 +45,7 @@ def show
 
     # GET /v0/institutions/20005123/children
     def children
-      children = Institution.version(@version[:number])
+      children = Institution.joins(:version)
```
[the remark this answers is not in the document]
elder-joseph: Hi @jholton thanks for the feedback. Not sure how I missed that for these instances. I've gone ahead and pushed a commit for three occurrences.

app/models/institution_tree.rb
```diff
@@ -38,43 +37,40 @@ def main_campus(institution)
...
-          UNION
-            SELECT
-              i.campus_type, i.facility_code, i.parent_facility_code_id
-            FROM institutions i
-            INNER JOIN related_up r ON r.parent_facility_code_id = i.facility_code
-            WHERE i.version = ?
-        ) SELECT * FROM related_up where campus_type = 'Y'
+      WITH RECURSIVE related_up AS(
+        SELECT campus_type, facility_code, parent_facility_code_id
+        FROM institutions WHERE facility_code = ?
```
jholton: I think you're still going to want the `version` clause:
```suggestion
        FROM institutions WHERE facility_code = ? AND version_id = ?
```

app/models/institution_tree.rb
```diff
@@ -38,43 +37,40 @@ def main_campus(institution)
...
 
     def descendants(ancestor)
       str = <<-SQL
         WITH RECURSIVE related_down AS(
-          SELECT facility_code
-          FROM institutions
-          WHERE facility_code = ? AND version = ?
+          SELECT i.facility_code
+          FROM institutions i
+          WHERE i.facility_code = ?
```
jholton: Same here:
```suggestion
          WHERE i.facility_code = ? AND version_id = ?
```

app/models/institution_tree.rb
```diff
@@ -38,43 +37,40 @@ def main_campus(institution)
...
           UNION
             SELECT i.facility_code
             FROM institutions i
             INNER JOIN related_down r ON i.parent_facility_code_id = r.facility_code
-            WHERE i.version = ?
+            INNER JOIN versions v ON v.id = i.version_id
```
jholton: You're currently not filtering on `version` in this SELECT.  Shouldn't this be
```suggestion
            WHERE i.version_id = ?
```

app/models/institution_builder.rb
```diff
@@ -603,22 +643,27 @@ def self.build_versioned_school_certifying_official(version_number)
...
-        last_name,
-        title,
-        phone_area_code,
-        phone_number,
-        phone_extension,
-        email,
-        ?
-      FROM school_certifying_officials
+        i.id
+      FROM school_certifying_officials s
```
jholton:
```
INSERT INTO school_certifying_officials
...
SELECT DISTINCT
...
FROM school_certifying_officials
```
Insert records into a table from itself?

app/serializers/institution_profile_serializer.rb
```diff
@@ -105,7 +105,7 @@ def yellow_ribbon_programs
   end
 
   def versioned_school_certifying_officials
-    object.school_certifying_officials.map do |sco|
+    object.school_certifying_officials.select { |sco| sco.institution_id == object.id }.map do |sco|
```
jholton: 🤔  I'd expect `school_certifying_officials` association with `institutions` to be based on `institution_id` already.
Why is it based on `facility_code`?

app/models/archiver.rb
```diff
@@ -14,9 +17,15 @@ def self.archive_previous_versions
 
     begin
       ApplicationRecord.transaction do
-        ARCHIVE_TYPES.each do |archivable|
-          create_archives(archivable[:source], archivable[:archive], previous_version, production_version)
-          archivable[:source].where('version >= ? and version < ?', previous_version, production_version).delete_all
+        ARCHIVE_TYPES_BY_PARENT_ID.each do |archivable|
+          create_archives_by_parent_id(archivable[:source], archivable[:archive], previous_version, production_version)
+          delete_by_parent_id(archivable[:source], archivable[:archive], previous_version, production_version)
```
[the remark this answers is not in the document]
elder-joseph: Hi @jholton, 

I pulled in your suggested changes for archiver here then tested them a bit, but noticed some strange behavior with some of the related specs. In **archive_previous_versions**, `production_version` and `previous_version` correspond to `version.number`.

I came up with this:
`archivable[:source].joins(institution: :version).where('number >= ? AND number <?', previous_version, production_version).delete_all`

and a similar Active Query for the other `.delete all`, which seems to resolve the issues I was running into. 

Thoughts on this?

app/models/gibct_site_mapper.rb
```diff
@@ -32,7 +32,10 @@ def generate_sitemap(version)
     SitemapGenerator::Sitemap.create do
       add '/search', priority: 0.9, changefreq: 'monthly'
 
-      Institution.version(version).find_each(batch_size: Settings.active_record.batch_size.find_each) do |institution|
+      # Institution.joins("INNER JOIN versions v ON institutions.version_id = #{version}")
+      #            .find_each(batch_size: Settings.active_record.batch_size.find_each) do |institution|
+      Institution.joins('INNER JOIN versions v ON institutions.version_id = ', version.to_s)
```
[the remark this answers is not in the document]
elder-joseph: Pushed a commit to correct this and all of the suggested `institution_tree.rb` changes

app/models/archiver.rb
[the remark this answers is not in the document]
jholton: Those two queries should produce very similar SQL.  What exactly is the issue?
Using `InstitutionProgram` as the `archivable[:source]` and random, believable version numbers, here's how they compare:

Yours:
```
InstitutionProgram.joins(institution: :version).where('number >= ? AND number <?', 4, 5).delete_all
# =>  InstitutionProgram Destroy (1.4ms)  DELETE FROM "institution_programs" WHERE "institution_programs"."id" IN (
  SELECT "institution_programs"."id"
  FROM "institution_programs" 
  INNER JOIN "institutions" ON "institutions"."id" = "institution_programs"."institution_id"
  INNER JOIN "versions" ON "versions"."id" = "institutions"."version_id"
  WHERE (number >= 4 AND number <5)
)
```

Mine:
```
irb(main):010:0> InstitutionProgram.joins(institution: :version).where(institutions: { version_id: 4..5 - 1}).delete_all
# =>  InstitutionProgram Destroy (2.2ms)  DELETE FROM "institution_programs" WHERE "institution_programs"."id" IN (
  SELECT "institution_programs"."id"
  FROM "institution_programs"
  INNER JOIN "institutions" ON "institutions"."id" = "institution_programs"."institution_id"
  INNER JOIN "versions" ON "versions"."id" = "institutions"."version_id"
  WHERE "institutions"."version_id" BETWEEN $1 AND $2)  [["version_id", 4], ["version_id", 4]]
```
Looks like the only difference is 
`WHERE (number >= 4 AND number <5)`
vs 
`WHERE "institutions"."version_id" BETWEEN $1 AND $2)  [["version_id", 4], ["version_id", 4]]`

I don't have a strong preference.  They will perform the same.  I am curious what problem you ran into given how similar they are.

app/serializers/institution_profile_serializer.rb
elder-joseph: That is a lot cleaner to update the association for institution and remove the mapping. I updated this in a commit I have pushed. Look it over when you get a chance and let me know what you think.

app/models/institution_builder.rb
elder-joseph: That is a bit clunky, but it seemed to accomplish what I was going for at the time.

I refactored it a little bit to what I have below in the latest commit, which seems a bit less clunky.
```
INSERT INTO school_certifying_officials(
          facility_code,
          institution_name,
          priority,
          first_name,
          last_name,
          title,
          phone_area_code,
          phone_number,
          phone_extension,
          email,
          institution_id)
       SELECT
        i.facility_code,
        i.institution,
        s.priority,
        s.first_name,
        s.last_name,
        s.title,
        s.phone_area_code,
        s.phone_number,
        s.phone_extension,
        s.email,
        i.id
      FROM institutions i
      INNER JOIN school_certifying_officials s ON i.facility_code = s.facility_code
      INNER JOIN versions v ON v.number = ?
      WHERE v.id = i.version_id
      AND i.id IS NOT NULL;
```

jholton: OK, let me step through the life of this table.

I upload a `sco.csv` with two records:
sco_start

Now I generate a new preview version and check my table again:
sco_run1
What happened?
`build_school_certifying_officials` added two new records with the `institution_id` populated as expected.  Then `add_sco_institution_id` ran and updated the original two records with an `institution_id`, so now we have duplicate data.

Then I generate another preview version, and check my table again.
sco_run2
Instead of adding two new rows for the new version, it doubled the number of rows.
What happened?
Your SQL joins `school_certifying_officials` to `institutions` on `facility_code`.  All 4 existing records join on that condition and thus 4 new records were inserted instead of the expected 4.

Finally, I upload my`sco.csv` file again.
re-up

Whoops!  When you upload a file it deletes all the existing data before importing the new data.  Looks like your still going to need a second table.

elder-joseph: Hi @jholton, 

I have actually run into this duplication issue before. I believe when I updated the query yesterday I dropped off the `UNIQUE` keyword from the query which is how I believe I had resolved this in the past. I will update the query adding that back in to see if resolves this duplication issue.

jholton: Don't miss my last point.  Fixing the duplication won't matter if the table gets wiped out when someone uploads a new csv.  😃

elder-joseph: Hey @jholton,

My apologies, I meant to loop back with you here. That point was not lost, I just pushed that commit correcting the duplication issue for testing purposes.

After discussing a plan forward with some of my team members I am working on re-implementing that functionality to the second table `versioned_school_certifying_officials`.

I saw exactly the behavior you described earlier and am working on a fix now 👍

jholton: Sounds good; just wanted to make sure.  I think a second table is the way to go.

elder-joseph: Pushed an update that reconfigures `versioned_school_certifying_officials` to work with the refactored archiver.

app/models/archiver.rb
```diff
@@ -28,14 +37,26 @@ def self.archive_previous_versions
...
-  def self.create_archives(source, archive, previous_version, production_version)
-    str = <<-SQL
+  def self.create_archives_by_parent_id(source, archive, previous_version, production_version)
+    str = <<~SQL
       INSERT INTO #{archive.table_name}
-        SELECT *
-          FROM #{source.table_name}
-          WHERE version >= ? and version < ? ;
+      SELECT s.* FROM #{source.table_name} s
+      JOIN institutions i ON s.institution_id = i.id
```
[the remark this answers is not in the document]
jholton: Same idea for the `delete_all` statement above:
```ruby
source = if archivable[:source].has_attribute?(:institution_id)
  archivable[:source].joins(institution: :version)
else
  archivable[:source].joins(:version)
end
source.where('number >= ? AND number <?', previous_version, production_version).delete_all
```

spec/models/institution_builder_spec.rb
```diff
@@ -947,13 +948,14 @@ def check_ipeds_ic_ay
...
-      it 'properly generates versioned school certifying official from school certifying official' do
-        create :school_certifying_official, facility_code: '2V000105'
-        expect { described_class.run(user) }.to change(VersionedSchoolCertifyingOfficial, :count).from(0).to(1)
-        expect(VersionedSchoolCertifyingOfficial.first.version).to eq(Version.current_preview.number)
+      describe 'when generating school certifying official table' do
+        it 'properly generates school certifying official with instituion_id' do
+          create :school_certifying_official, facility_code: '1ZZZZZZZ'
+          expect(VersionedSchoolCertifyingOfficial.count).to eq(0)
+          described_class.run(user)
+          expect(VersionedSchoolCertifyingOfficial.last.institution_id).to be_truthy
```
jholton: Suggestion:
```ruby
expect { described_class.run(user) }.to change(VersionedSchoolCertifyingOfficial, :count).from(0).to(1)
expect(VersionedSchoolCertifyingOfficial.last.institution_id).to be_present
```

app/models/school_certifying_officials_archive.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class SchoolCertifyingOfficialsArchive < SchoolCertifyingOfficial
```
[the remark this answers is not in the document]
jholton: I see you just pushed a commit to drop the table.

app/models/institution_builder.rb
elder-joseph: That probably makes sense to go ahead and do, I'll add that migration to do so.

elder-joseph:
@jholton I just pushed changes based off of your suggested refactors for `institution_tree.rb`, `institution_builder.rb`, the `create_archives` and `delete_all` queries of `archiver.rb`, updated `institution_builder_spec.rb `, and added a migration to remove `institution_id` from `school_certifying_officials` since it is not needed.

app/models/institution_builder.rb
```diff
@@ -453,21 +454,22 @@ def self.add_yellow_ribbon_programs(version_number)
         (SELECT
           i.version, i.id, yrs.degree_level, yrs.division_professional_school,
             yrs.number_of_students, yrs.contribution_amount, NOW(), NOW()
-          FROM institutions as i, yellow_ribbon_program_sources as yrs
+          FROM institutions as i, yellow_ribbon_program_sources as yrs, versions
           WHERE i.facility_code = yrs.facility_code
-          AND i.version = ?);
+          AND versions.number = #{version_number}
+          AND versions.id = i.version_id);
```
jholton: I think this can also be reduced to 
```
          FROM institutions as i, yellow_ribbon_program_sources as yrs
          WHERE i.facility_code = yrs.facility_code
          AND i.version_id = #{version_id};
```

Maybe in a separate PR, but `YellowRibbonProgram` looks like a candidate to reduce/remove it's `version` attribute since it `belongs_to` `institution`.
You'll have to create a migration to stop requiring it be populated:
```
  def change
    change_column_null :yellow_ribbon_programs, :version, true
  end
```
Then you can stop populating it, and eventually remove it.

app/models/institution_builder.rb
```diff
@@ -565,14 +566,15 @@ def self.build_institution_programs(version_number)
...
+        INNER JOIN edu_programs e ON p.facility_code = e.facility_code
           AND LOWER(description) = LOWER(vet_tec_program)
           AND vet_tec_program IS NOT NULL
-        INNER JOIN institutions c ON a.facility_code = c.facility_code
-          AND c.version = ?
-          AND c.approved = true;
+        INNER JOIN institutions i ON p.facility_code = i.facility_code
+        INNER JOIN versions v ON v.number = ?
+        WHERE v.id = i.version_id
+          AND i.approved = true;
```
jholton: Another candidate to remove the `versions` join:
```
        INNER JOIN edu_programs e ON p.facility_code = e.facility_code
          AND LOWER(description) = LOWER(vet_tec_program)
          AND vet_tec_program IS NOT NULL
        INNER JOIN institutions i ON p.facility_code = i.facility_code
        WHERE i.version_id = #{version_id}
          AND i.approved = true;
```

app/models/institution_builder.rb
```diff
@@ -615,8 +617,11 @@ def self.build_versioned_school_certifying_official(version_number)
         phone_number,
         phone_extension,
         email,
-        ?
-      FROM school_certifying_officials
+        i.id
+      FROM school_certifying_officials s
+      INNER JOIN versions v ON v.number = ?
+      INNER JOIN institutions i ON i.facility_code = s.facility_code
+      WHERE v.id = i.version_id
```
jholton: Last candidate for removing the `versions` join:
```
      INNER JOIN institutions i ON i.facility_code = s.facility_code
      WHERE i.version_id = #{version_id}
```

db/migrate/20200212205137_add_institution_id_to_versioned_school_certifying_officials_archives.rb
```diff
@@ -0,0 +1,5 @@
+class AddInstitutionIdToVersionedSchoolCertifyingOfficialsArchives < ActiveRecord::Migration[5.2]
+  def change
+    add_column :versioned_school_certifying_officials_archives, :institution_id, :bigint
+  end
+end
```
jholton: This is unnecessary.  The previous migration (`20200212204337_add_institution_id_to_versioned_school_certifying_officials.rb `) takes care of this.

db/migrate/20200213231744_remove_institution_id_from_school_certifying_officials.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveInstitutionIdFromSchoolCertifyingOfficials < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured { remove_column :school_certifying_officials, :institution_id, :bigint }
```
jholton: Why do you need `safety_assured`?

app/models/school_certifying_officials_archive.rb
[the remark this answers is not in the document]
jholton: Don't forget to remove this file (`school_certifying_officials_archive.rb`)

jholton (commented):
We'll also have to figure out the correct order to run the migrations

app/models/institution_builder.rb
elder-joseph: I had actually made this refactor, but it caused the majority of the `institution_builder_spec` queries to bomb out, so I left it out of the refactor for now. 

I also noticed `YellowRibbonProgram` here using the `version` attribute and had similar feelings regarding that, but I think that would probably be best handled in a separate PR.

db/migrate/20200213231744_remove_institution_id_from_school_certifying_officials.rb
elder-joseph: When running `bundle exec rake db:migrate` without wrapping 

`remove_column :school_certifying_officials, :institution_id, :bigint`

I get a message regarding strong migrations 
`=== Dangerous operation detected #strong_migrations ===`
Which ends the migrations and and recommends wrapping this block with a `safety_assured { ... } block`, which allowed the migrations to be run successfully.

db/migrate/20200212205137_add_institution_id_to_versioned_school_certifying_officials_archives.rb
elder-joseph: I think that we actually do need this migration, since `versioned_school_certifying_officials` and `versioned_school_certifying_officials_archives` are both created before `institution_id` is added to `versioned_school_certifying_officials`, `institution_id` will not be automatically added to `versioned_school_certifying_officials_archives`.

elder-joseph:
Hey @jholton,

I just pushed commits which made 2 of your suggested query refactors in `institution_builder.rb`. I agree that it would be a good idea to refactor the 3rd `YellowRibbonProgram`, but after discussing with my team a bit, we think this should probably be done as a separate story where `YellowRibbonProgram` itself is refactored, similar to how you described above.

Also removed `school_certifying_officials_archive.rb`.

db/migrate/20200213231744_remove_institution_id_from_school_certifying_officials.rb
jholton: Does it give more info?  why is it a dangerous operation?

db/migrate/20200212205137_add_institution_id_to_versioned_school_certifying_officials_archives.rb
jholton: Ooooops, I totally misread the table name.  Sorry!

db/migrate/20200213231744_remove_institution_id_from_school_certifying_officials.rb
jholton:
```

ActiveRecord caches attributes which causes problems
when removing columns. Be sure to ignore the column:

class SchoolCertifyingOfficial < ApplicationRecord
  self.ignored_columns = ["institution_id"]
end

Deploy the code, then wrap this step in a safety_assured { ... } block.

class StrongMigrations::Checker < ActiveRecord::Migration[5.2]
  def change
    safety_assured { remove_column :school_certifying_officials, :institution_id, :bigint }
  end
end
```

So, it wants you to add
```ruby
self.ignored_columns = ["institution_id"]
```
to `SchoolCertifyingOfficial`, then deploy.  Then run the migration that drops the column.  Then you can remove 
```ruby
self.ignored_columns = ["institution_id"]
```

db/migrate/20200212205137_add_institution_id_to_versioned_school_certifying_officials_archives.rb
elder-joseph: Not a problem! With the relatively long and similar names, it is easy to do so.

spec/models/shared_examples/shared_examples_for_archivable_by_parent_id.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.shared_examples 'an archivable model' do |options|
+RSpec.shared_examples 'an archivable model by parent id' do |options|
```
jholton: Can we revert the changes in this file? (i.e. can this file and `shared_examples_for_archivable_by_version_id.rb ` be merged together?)
What is the difference between the files?

db/migrate/20200213231744_remove_institution_id_from_school_certifying_officials.rb
elder-joseph: I see, I will ignore the column for now and temporarily remove that migration.

Once deployed I can submit another PR to add the migration back in and remove `self.ignored_columns = ["institution_id"]`

spec/controllers/v0/institutions_controller_spec.rb
```diff
@@ -300,15 +300,15 @@ def check_boolean_facets(facets)
     end
 
     it 'returns profile details' do
-      school = create(:institution, :in_chicago)
-      get(:show, params: { id: school.facility_code, version: school.version })
+      school = create(:institution, :in_chicago, version_id: Version.current_production.id)
+      get(:show, params: { id: school.facility_code, version_id: school.version_id })
```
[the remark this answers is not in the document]
elder-joseph: If I do not supply `version_id` to the institution being created here and remove `version_id` from the `get`, 

my response comes back as a 404 where `"The record identified by 00000002 could not be found"`.

That is the case for all three of these occurrences in this file.

spec/models/shared_examples/shared_examples_for_archivable_by_parent_id.rb
elder-joseph: I actually had considered this and implemented a solution that combined the files when refactoring the archiver sql and active record queries based off of your suggestion. The issue I ran into was that while the two types of models are similar the setup and tests vary slightly and require some additional conditional logic, since one type relies on `version_id` while the other uses `institution_id`.

I wasn't sure what would be the lesser of two evils in this scenario, adding conditional logic for a single test for all 4 models, or having 2 tests for the two types of models as it is currently.

Thoughts?

spec/controllers/v0/institutions_controller_spec.rb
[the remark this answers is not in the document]
jholton: I understand why you need to provide it in the `create(:institution, ...) call, but it's you can remove `version_id: school.version_id` from the `get` request.  It's doing nothing. 
A `version_id` parameter is not used anywhere (see `api_controller#resolve_version` to see how the `version` parameter is used).

To confirm, I removed it locally from the specs and they pass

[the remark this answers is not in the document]
elder-joseph: Noted, removed these three instances and pushed an updated commit here.

spec/models/shared_examples/shared_examples_for_archivable_by_version_id.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    Archiver.archive_previous_versions
+
+    expect(original_type.count).to eq(count_total)
+    expect(archived_type.count).to eq(archive_count)
+
+    if Version.current_preview
+      expect(original_type.where('version_id = ? OR version_id = ?', current_production_id, current_preview_id).size)
+        .to eq(count_greater_equal_production)
+      expect(archived_type.where('version_id != ? OR version_id != ?', current_production_id, current_preview_id).size).to eq(archive_count)
```
jholton: Are we trying to show here that `archived_type` does not have current versions?
We could do `expect(archived_type.where('version_id IN (?, ?)', current_production_id, current_preview_id).size).to eq(0)`
This also does not account for multiple new preview versions.

spec/models/shared_examples/shared_examples_for_archivable_by_version_id.rb
```diff
@@ -0,0 +1,124 @@
...
+    else
+      expect(original_type.where('version_id = ?', current_production_id).size)
+        .to eq(count_greater_equal_production)
+      expect(archived_type.where('version_id != ?', current_production_id).size).to eq(archive_count)
```
jholton: Same comments apply for these two.  They are doing the same thing except for scenarios where there is no `Version.current_preview`.

So, overall, it looks like the goal of these specs are to make sure that current production/preview versions are not in the archive tables and vice versa.  Do you agree?

I've been messing around with this locally, so I'll open a PR to this branch shortly that illustrates changes along these lines.

spec/factories/institution_programs.rb
```diff
@@ -32,23 +32,23 @@
     end
 
     trait :in_nyc do
-      institution { create(:institution, physical_city: 'NEW YORK', physical_country: 'USA', physical_state: 'NY') }
+      institution { create(:institution, physical_city: 'NEW YORK', physical_country: 'USA', physical_state: 'NY', version_id: Version.last.id) }
```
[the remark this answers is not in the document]
elder-joseph: I agree that this is less than ideal. I have taken note of this and agree that it would be a good idea to refactor this when possible.

spec/models/shared_examples/shared_examples_for_archivable_by_version_id.rb
elder-joseph: I agree, as long as the tests provide a reasonable level of confidence that production/preview versions are in the correct place before and after the archiver runs, then I think we accomplish this goal.

I am taking a look over your PR related to this now.

jholton:
Do you have a separate PR for the migrations?  The two PRs listed in the description don't cover the remaining migrations.

elder-joseph:
@jholton I do have a separate PR #576 to handle the updated migrations. Just updated the description to include it

jholton (approved):

PR #553, opened 2020-01-29
Improve institutions query performance

devin-mccurdy:
## Description
Remove unnecessary casing of `institutions.institution` and `institutions.city` (those fields are imported as uppercase) in institution search.

Related DB migration PR: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/557

## Testing done

## Screenshots
N/A

## Acceptance criteria
- [x] Institutions end point queries are optimized

## Definition of done


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

app/controllers/v0/institutions_controller.rb
```diff
@@ -114,17 +114,17 @@ def search_results
...
       result = {
         category: {
           school: institution_types.except(Institution::EMPLOYER).inject(0) { |count, (_t, n)| count + n },
           employer: institution_types[Institution::EMPLOYER].to_i
         },
         type: institution_types,
-        state: search_results.filter_count(:state),
-        country: embed(search_results.filter_count(:country)),
+        state: count_field(search_results, :state),
+        country: embed(count_field(search_results, :country)),
```
jholton: Did you test (measure) these changes?
This change makes requests much slower, depending on the number of results.
If I visit http://localhost:3000/v0/institutions (no search terms) on my machine, the requests are 5x - 10x SLOWER.
That's worst case scenario, but the more results terms the worse it gets.
These changes do not improve speed at any search results threshold.

Here are two examples each of the new way and old way when I visit http://localhost:3000/v0/institutions:

Old results:
Completed 200 OK in **902ms** (Views: 171.5ms | ActiveRecord: 668.0ms)
Completed 200 OK in **419ms** (Views: 26.8ms | ActiveRecord: 382.0ms)

New Results:
Completed 200 OK in **5002ms** (Views: 38.5ms | ActiveRecord: 541.9ms)
Completed 200 OK in **4581ms** (Views: 41.2ms | ActiveRecord: 678.4ms)

You removed a SQL count statement and replaced it with a Ruby function that has to instantiate every search result object and loop through them (not just results for the current page; _all_ matching results).

SQL is (generally) faster than Ruby.  We need to optimize our SQL, and remove SQL if we can, but replacing SQL with Ruby is almost never the right answer.

app/controllers/v0/institutions_controller.rb
```diff
@@ -24,7 +24,7 @@ def autocomplete
     def index
       @meta = {
         version: @version,
-        count: search_results.count,
+        count: search_results.load.size,
```
jholton: This causes a SQL SELECT that loads returns every search result.  
The `COUNT` select that this replaces is much faster.
These search results are not used anywhere else (if you revert the `count_field` changes below; see my comment there), so the previous `.count` is much preferred.

app/controllers/v0/institution_programs_controller.rb
```diff
@@ -23,7 +23,7 @@ def autocomplete
     def index
       @meta = {
         version: @version,
-        count: search_results.count,
+        count: search_results.load.size,
```
jholton: This makes sense because `search_results` are used in the `count_field` method.  But given my other comments, I suspect the best path would be to use `search_results.filter_count` method (as used in `institutions_controller`) instead of `count_field` in `facets`, and revert this to `search_results.count`

jholton (changes requested):
Please measure these changes first.
Just look at your Rails log output.  There's a lot of info there.

For example, I noticed that `institution_programs` search requests are clearly generating an N+1 queries scenario.
Changing [this line](https://github.com/department-of-veterans-affairs/gibct-data-service/blob/64beff3399c34f857a58ba3f67836b0bd7881de9/app/controllers/v0/institution_programs_controller.rb#L50) to `.includes(:institution)` cuts the request time by close to half.

app/controllers/v0/institutions_controller.rb
dmccurdybah: Yes, I tested the performance, but with specific search terms and got much smaller result sets and got much different results.I don't think we have a route that would allow searching without any terms, but I am seeing the performance issues with terms that return huge result sets ("name=university" unsurprisingly has a large amount of records).

jholton (approved):

PR #559, opened 2020-02-04
add associations between `Version` and `Institution`

jholton:
## Description
add associations between `Version` and `Institution`
refactor `InstitutionProgram.autocomplete`

### DIFF
--- a/app/controllers/v0/institution_programs_controller.rb
+++ b/app/controllers/v0/institution_programs_controller.rb
@@ -9,9 +9,7 @@ module V0
       @data = []
       if params[:term].present?
         @search_term = params[:term]&.strip&.downcase
-        @data = InstitutionProgram.joins('INNER JOIN versions v ON v.id = institutions.version_id')
-                                  .where('v.number = ?', @version.number)
-                                  .autocomplete(@search_term)
+        @data = InstitutionProgram.autocomplete(@search_term, @version)
       end
       @meta = {
         version: @version,
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -156,6 +156,7 @@ class Institution < ApplicationRecord
   has_many :institution_programs, -> { order(:description) }, inverse_of: :institution, dependent: :nullify
   has_many :school_certifying_officials, -> { order 'priority, last_name' },
            primary_key: :facility_code, foreign_key: 'facility_code', inverse_of: :institution
+  belongs_to :version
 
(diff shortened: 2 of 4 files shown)


app/models/institution.rb
```diff
@@ -156,6 +156,7 @@ class Institution < ApplicationRecord
   has_many :institution_programs, -> { order(:description) }, inverse_of: :institution, dependent: :nullify
   has_many :school_certifying_officials, -> { order 'priority, last_name' },
            primary_key: :facility_code, foreign_key: 'facility_code', inverse_of: :institution
+  belongs_to :version
```
jholton: The migration for the association has been run, but they were never set up in the models.

app/models/version.rb
```diff
@@ -3,6 +3,7 @@
 require 'securerandom'
 
 class Version < ApplicationRecord
+  has_many :institutions, dependent: :nullify
```
jholton: Set up association.

app/controllers/v0/institution_programs_controller.rb
```diff
@@ -9,9 +9,7 @@ def autocomplete
       @data = []
       if params[:term].present?
         @search_term = params[:term]&.strip&.downcase
-        @data = InstitutionProgram.joins('INNER JOIN versions v ON v.id = institutions.version_id')
-                                  .where('v.number = ?', @version.number)
-                                  .autocomplete(@search_term)
+        @data = InstitutionProgram.autocomplete(@search_term, @version)
```
jholton: Some of the ActiveQuery chains were in `autocomplete` and some were here.  I moved them all to `autocomplete`

app/models/institution_program.rb
```diff
@@ -43,14 +43,12 @@ def va_bah
...
-                         "#{search_term}%"
-                       )
-                       .group('institution_programs.id, institutions.facility_code, description')
-                       .limit(limit)
+  def self.autocomplete(search_term, version, limit = 6)
+    select(:id, 'institutions.facility_code as value', 'description as label')
+      .joins(institution: :version)
+      .where(institutions: { version: version })
+      .where('lower(description) LIKE (?)', "#{search_term}%")
+      .limit(limit)
```
jholton: Now that we have the associations set up, we can use more ActiveQuery and less raw SQL

elder-joseph (approved):

PR #573, opened 2020-02-11
Add the matching score to the matching Ipeds_HD table

hughes-dustin:
## Description
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/5726

The user should be able to see the matching scores of each entry in the ipeds_hd table so they have a better idea of how closely they match to the weams data. 

## Testing done
dev tested

## Screenshots
image

## Acceptance criteria
- [x] A column displaying the % match is displayed as the first column in the potential matches table.
- [x] The table is sorted descending by % match.

## Definition of done


app/controllers/crosswalk_issues_controller.rb
```diff
@@ -47,17 +47,30 @@ def find_matches
...
+        SELECT *, GREATEST(similarity(institution, #{escaped_institution}),
+                      similarity((city||state||zip||addr), '#{address_data_to_match}')
+                  , similarity((city||state||zip||addr), '#{physical_address_data}')) AS match_score
+        FROM ipeds_hds
+        WHERE (similarity(institution,  #{escaped_institution}) > 0.5
+          OR similarity((city||state||zip||addr), '#{address_data_to_match}') > 0.3
+          OR similarity((city||state||zip||addr), '#{physical_address_data}') > 0.3)
+        ORDER BY GREATEST(GREATEST(similarity(institution, #{escaped_institution}),
+                                         similarity((city||state||zip||addr), '#{address_data_to_match}')),
+                                         similarity((city||state||zip||addr), '#{physical_address_data}')) DESC
```
shawkey-daniel: Since the greatest was evaluated in the select clause, I don't think you need to evaluate greatest again. ORDER BY match_score DESC should work.

shawkey-daniel (changes requested):

app/controllers/crosswalk_issues_controller.rb
hughes-dustin: Good point. I made the change.

app/controllers/crosswalk_issues_controller.rb
```diff
@@ -47,17 +47,21 @@ def find_matches
...
-    sanitize_order = IpedsHd.sanitize_sql_for_order("similarity(institution, #{escaped_institution}) DESC,
-                                         similarity((city||state||zip||addr), '#{address_data_to_match}') DESC,
-                                         similarity((city||state||zip||addr), '#{physical_address_data}') DESC")
 
-    @ipeds_hd_arr = IpedsHd.where(query, "%#{@issue.weam.institution}%",
-                                  "%#{address_data_to_match}%",
-                                  "%#{physical_address_data}%")
-                           .order(sanitize_order)
+    str = <<-SQL
+        SELECT *, GREATEST(similarity(institution, #{escaped_institution}),
```
shawkey-daniel: I should have mentioned this on my last review. Please explicitly list the fields for the select instead of using the asterisk. Then it will be good to go.

shawkey-daniel (approved):
LGTM

app/controllers/crosswalk_issues_controller.rb
hughes-dustin: Got it. The changes are in along with removing the pointless mapping method.

sonntag-adam (approved):

app/controllers/crosswalk_issues_controller.rb
```diff
@@ -47,17 +47,20 @@ def find_matches
...
+        SELECT id, ipeds_hds.cross, institution, addr, state, city, zip, ope, GREATEST(similarity(institution, #{escaped_institution}),
+                      similarity((city||state||zip||addr), '#{address_data_to_match}')
+                  , similarity((city||state||zip||addr), '#{physical_address_data}')) AS match_score
+        FROM ipeds_hds
+        WHERE (similarity(institution,  #{escaped_institution}) > 0.5
+          OR similarity((city||state||zip||addr), '#{address_data_to_match}') > 0.3
+          OR similarity((city||state||zip||addr), '#{physical_address_data}') > 0.3)
+        ORDER BY match_score DESC
+    SQL
+    sql = Institution.send(:sanitize_sql, str)
```
erluti: `santize_sql` is already available on a Model, so send isn't necessary.  I think it makes more sense to call it against the model you're working with, so `IpedsHd.sanitize_sql(str)` makes a bit more sense to me.
I was reading up on the docs for `sanitize_sql` https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Sanitization/ClassMethods.html and I don't think it's really necessary here.  I think it makes more sense to use the `quote` method like `escaped_institution` uses above.  Then you could even just call the `str` variable `sql` and leave out this sanitize line.

app/views/crosswalk_issues/find_matches.html.erb
```diff
@@ -29,19 +29,21 @@
             <th>Address</th>
             <th>OPE</th>
             <th>IPE</th>
+            <th>Matching Score</th>
             <th>Actions</th>
             </thead>
             <tbody>
-            <% @ipeds_hd_arr.each do |iped| %>
+            <% @ipeds_hd_arr.each do |i| %>
```
erluti: Why was this variable changed to a less descriptive name?

philipwritescode (changes requested):
I have just a few questions, but this is a solid approach

app/controllers/crosswalk_issues_controller.rb
hughes-dustin: I believe we need the `sanitize_sql ` call. For one thing, CodeClimate finds a possible sql injection issue without it. Which means we wouldn't be able to merge in the changes. From what I gathered, the fix needed for this issue is to sanitize the string. 

The use of `Institution` was a typo on my part. You are right, we should definitely use the model we are working in, though I don't think it really matters as long as it's a child of `ActiveRecord`.

app/views/crosswalk_issues/find_matches.html.erb
hughes-dustin: Ha! I honestly don't remember making that change. Good catch!

app/controllers/crosswalk_issues_controller.rb
erluti: That's sort of annoying of code climate, because these values are from a model and not from user input.  But I'm new to VSP, so I didn't know it was going to flag that.

erluti: I think `sanitize_sql` is intended to work only on a `WHERE` clause, so I'd personally be cautious about applying it to the whole thing.

philipwritescode (approved):
Personally I wouldn't trust `sanitize_sql` on the whole query because the spec say it's just for where clauses, but if it works it works.
LGTM

app/controllers/crosswalk_issues_controller.rb
hughes-dustin: I'll point this out to my team for the refactor story we have for next sprint. Maybe we can find an alternative approach from `sanitize_sql`.

PR #575, opened 2020-02-12
Bah uploadtransaction 5361

ksantiagoBAH:
## Description
As a developer, I need to make the GIDS file upload process transactional so that old source data is not deleted until the upload has been successful.

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/5361

## Testing done
Local testing completed

## Acceptance criteria
1. If a file upload fails
- [x] The previously uploaded data is available in the corresponding table.
- [x] The previously uploaded data is exportable.
2. If a file upload is successful
- [x] The previously uploaded data is removed from the corresponding table.
- [x] The previously uploaded data is not included in the CSV export.

## Definition of done


shawkey-daniel (approved):
LGTM

sonntag-adam (approved):

spec/models/shared_examples/shared_examples_for_loadable.rb
```diff
@@ -43,6 +43,12 @@
         expect(results.num_inserts).to eq(1)
         expect(results.ids.length).to eq(1)
       end
+
+      it 'does not delete previous record if upload is invalid' do
```
erluti: This is not a failing test before the changes.  Using `byebug` I noticed that `described_class.count` is 5 , which is not what `initial_result_count` is set to.  The `change` matcher expects the change to happen inside the block given to `expect`, and by assigning a value to `results` means no action is being taken and this expect will never see a change.

Also, `transaction` doesn't rollback unless there's an exception, so this test should probably use something like `allow(described_class).to receive(:load_csv_file).and_raise("error in csv")` before `load` is called so that there's an exception.

lib/csv_helper/loader.rb
```diff
@@ -10,13 +10,17 @@ module Loader
...
+      klass.transaction do
+        delete_all
+        load_csv_file(file, options)
+      end
     end
 
     def load_from_api(results, options = {})
-      delete_all
-      load_records(results, options)
+      klass.transaction do
```
erluti: there should also be a test to go with this change

philipwritescode (changes requested):
looks like a good change, but there are some concerns around the associated tests.

spec/models/shared_examples/shared_examples_for_loadable.rb
```diff
@@ -43,6 +43,19 @@
         expect(results.num_inserts).to eq(1)
         expect(results.ids.length).to eq(1)
       end
+
+      it 'does rolls back to the old table content if the upload is invalid' do
```
shawkey-daniel: Change to "does roll back ..."

shawkey-daniel (approved):
LGTM

sonntag-adam (approved):

lib/csv_helper/loader.rb
```diff
@@ -10,13 +10,21 @@ module Loader
...
 
     def load(file, options = {})
-      delete_all
-      load_csv_file(file, options)
+      klass.transaction do
+        delete_all
+        load_csv_file(file, options)
+      end
+    rescue StandardError => e
+      "Error #{e.message}"
```
erluti: Are you sure `load` should return a string when there's an exception?  Looking in the uploads_controller, I think a string would cause an error when `data.ids` is called in `call_load`.  I think it would probably be okay to let the error just bubble up since the uploads controller has some rescues.  Might be good to add a test for this issue in the uploader_controller spec.  This would probably also mean changing the test for this to expect the block to `raise_error` and then instead of using the change matcher just wrap it yourself with `initial_count = described_class.count` and after the error using `expect(described_class.count).to eq initial_count`

spec/models/shared_examples/shared_examples_for_loadable.rb
```diff
@@ -43,6 +43,28 @@
         expect(results.num_inserts).to eq(1)
         expect(results.ids.length).to eq(1)
       end
+
+      it 'does roll back to the old table content if the upload is invalid' do
+        allow(described_class).to receive(:load_csv_file).and_raise('error in csv')
+        before_count = described_class.count
+        begin
+          described_class.load(csv_file_invalid, load_options)
```
erluti: you can avoid the begin/rescue with `expect{described_class.load(csv_file_invalid, load_options)}.to raise_error`

philipwritescode (approved):
LGTM!

PR #578, opened 2020-02-20
Refactor `archivable` shared rspec examples

jholton:
## Description
Feature branch refactor PR


spec/models/shared_examples/shared_examples_for_archivable.rb
```diff
@@ -82,35 +87,27 @@ def check_log_for_error(error_message)
...
-  def create_preview_version
-    create :version, :preview
-    create :institution, version_id: current_preview_id
-    create factory, institution_id: Institution.last.id
-  end
 
-  def current_preview_id
-    Version.current_preview.id
+    expect(original_type.distinct.pluck(:version_id)).to include(*[Version.current_production.id, Version.current_preview&.id].compact)
+    expect(archived_type.pluck(:id)).not_to include(*original_type.pluck(:id))
```
jholton: Check that `archived` and `current` records were separated. (i.e. no `current` records in `archived`).
I consider this test "OK, but not ideal".  The records with `institutions` associations can't be tied back to `version` right now due to missing associations among `archives` tables.

elder-joseph (approved):
After looking this over a bit, I think that this refactor reasonably addresses the same general spirit of the the previous unit tests, given the shift to `version_id`. 

I agree that it would be nice once we are able to establish the appropriate associations, to include in that refactor effort the test for records with `institutions` associations.

elder-joseph (approved):

PR #584, opened 2020-02-21
Kelsonic 5597 create yr programs endpoint

kelsonic:
## Description
**Issue:** https://github.com/department-of-veterans-affairs/va.gov-team/issues/5597

This PR implements the endpoint `GET https://{BASE_URL}/v0/yellow_ribbon_programs?page=1&per_page=30&sort_by=number_of_students&sort_direction=desc&city=boulder&country=usa&contribution_amount=unlimited&number_of_students=unlimited&school_name_in_yr_database=university&state=co` that returns back YR Program Source entities.

## Testing done
Wrote all of the tests I could think of, but if I should add more please just let me know! 🙂

## Screenshots
image

## Acceptance criteria
- [x] Create `GET https://{BASE_URL}/v0/yellow_ribbon_programs?page=1&per_page=30&sort_by=number_of_students&sort_direction=desc&city=boulder&country=usa&contribution_amount=unlimited&number_of_students=unlimited&school_name_in_yr_database=university&state=co` endpoint.
- [x] Write tests for the new endpoint and related code.

## Definition of done


app/models/yellow_ribbon_program_source.rb
```diff
@@ -4,48 +4,89 @@ class YellowRibbonProgramSource < ApplicationRecord
...
-    'sfr email address' => { column: :sfr_email_address },
-    'initials yr processor' => { column: :initials_yr_processor },
+    'updated for 2011-2012' => { column: :updated_for_2011_2012, converter: BooleanConverter },
+    'withdrawn' => { column: :withdrawn, converter: BooleanConverter },
     'year of yr participation' => { column: :year_of_yr_participation },
-    'notes' => { column: :notes }
   }.freeze
 
-  validates :facility_code, presence: true
+  # validates :city, presence: true
```
kelsonic: I left these commented as their presence, while not needed, would certainly be useful since much of our design spec depends on them.

spec/factories/yellow_ribbon_program_source.rb
```diff
@@ -2,15 +2,26 @@
...
+    school_name_in_yr_database { 'University'}
+    state { 'CO' }
+    street_address { '1234 University Drive'}
+    zip { '80303' }
 
     trait :institution_builder do
       facility_code { '1ZZZZZZZ' }
     end
+
+    trait :in_florence do
```
kelsonic: Tribute @ncksllvn lol

spec/serializers/yellow_ribbon_program_source_serializer_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:yellow_ribbon_program_source) { create :yellow_ribbon_program_source }
+  let(:data) { JSON.parse(subject)['data'] }
+  let(:attributes) { data['attributes'] }
+
+  it 'includes city' do
+    expect(attributes['city']).to eq(yellow_ribbon_program_source.city)
+  end
+
+  it 'includes contribution_amount' do
+    expect(attributes['contribution_amount']).to eq(yellow_ribbon_program_source.contribution_amount)
```
kelsonic: This test is currently failing as `contribution_amount` is a string instead of (what it should be) an integer.

app/controllers/v0/yellow_ribbon_programs_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      query = params.deep_dup
+      query.tap do
+        query[:name].try(:strip!)
+        query[:name].try(:downcase!)
+      end
+    end
+
+    def search_results
+      @query ||= normalized_query_params
+      YellowRibbonProgramSource.search(@query[:name])
```
zurbergram: I would recommend using https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/yellow_ribbon_program.rb instead as this table is linked to the relevant institution and if needed Ruby on Rails can be used to retreive institution data

app/models/yellow_ribbon_program_source.rb
zurbergram: If these fields are always populated in the CSV and are needed for you work i would uncomment these
Additionally you can list multiple fields per validates, see https://guides.rubyonrails.org/active_record_validations.html#presence

app/controllers/v0/yellow_ribbon_programs_controller.rb
kelsonic: Hmm, `YellowRibbonProgramSource` seems to have exactly what I need atm 🤔

zurbergram: The issue with using `yellow_ribbon_program_source` is that it is blown away every time a CSV is successfully uploaded meaning the data will change on the fly whenever someone updates the CSV instead of only when a “preview generation” and “publish to production” occur

zurbergram: If `YellowRibbonProgram` is missing needed data create a [Active Record Migration](https://edgeguides.rubyonrails.org/active_record_migrations.html) while following [Strong Migrations Gem](https://github.com/ankane/strong_migrations) to add the missing columns to the table

Then in https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L449 update the SQL to insert the data from `yellow_ribbon_program_source` into `yellow_ribbon_program`

kelsonic: Perfect, thanks for this suggestion! I updated the PR to only use `YellowRibbonProgram` instead of `YellowRibbonProgramSource` now 🙂

app/models/yellow_ribbon_program_source.rb
kelsonic: These are definitely not needed anymore since we're relying on the `yellow_ribbon_program.rb` model now.

kelsonic:
Couldn't quite discern what Jenkins' error was here:

image

app/models/yellow_ribbon_program_source.rb
zurbergram: Validates only run for tables that are populated by CSV import as https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L449 uses pure SQL and does not trigger validates on objects

If you would like to validate that these fields have values then `yellow_ribbon_program_source` model needs the validates, however if you are going to ignore rows without these values at https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L449 then the validates are not needed

zurbergram:
> Couldn't quite discern what Jenkins' error was here:
> 
> image

Make sure to run `bundle exec rake ci` as this will run the linter, security and spec tasks one of which is causing the Jenkins failure

app/models/yellow_ribbon_program.rb
```diff
@@ -1,4 +1,52 @@
...
+  validates :city, presence: true
+  validates :contribution_amount, numericality: true
+  validates :degree_level, presence: true
+  validates :division_professional_school, presence: true
+  validates :facility_code, presence: true
+  validates :number_of_students, numericality: true
+  validates :school_name_in_yr_database, presence: true
+  validates :state, presence: true
+  validates :street_address, presence: true
+  validates :zip, presence: true
```
zurbergram: Validates only run for tables that are populated by CSV import as https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L449 uses pure SQL and does not trigger validates on objects

If you would like to validate that these fields have values then yellow_ribbon_program_source model needs the validates, however if you are going to ignore rows without these values at https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L449 then the validates are not needed

zurbergram: https://guides.rubyonrails.org/active_record_validations.html#when-does-validation-happen-questionmark

app/models/yellow_ribbon_program_source.rb
```diff
@@ -4,48 +4,89 @@ class YellowRibbonProgramSource < ApplicationRecord
...
+  validates :facility_code, presence: true
   validates :number_of_students, numericality: true
-  validates :contribution_amount, numericality: true
+  # validates :school_name_in_yr_database, presence: true
+  # validates :state, presence: true
+  # validates :street_address, presence: true
+  # validates :zip, presence: true
+
+  # Finds exact-matching facility_code or partial-matching school and city names
+  scope :search, lambda { |search_term|
```
zurbergram: is this needed since switching to `yellow_ribbon_program`?

app/models/yellow_ribbon_program_source.rb
```diff
@@ -4,48 +4,89 @@ class YellowRibbonProgramSource < ApplicationRecord
...
+
+    where(
+      sanitize_sql_for_conditions(
+        [clause.join(' OR '),
+         search_term: "%#{search_term}%"]
+      )
+    )
+  }
+
+  scope :filter, lambda { |field, value|
```
zurbergram: is this needed since switching to `yellow_ribbon_program`?

app/models/yellow_ribbon_program_source.rb
```diff
@@ -4,48 +4,89 @@ class YellowRibbonProgramSource < ApplicationRecord
...
+    when 'false', 'no'
+      where.not(field => true)
+    else
+      where(field => value)
+    end
+  }
+
+  scope :version, ->(n) { where(version: n) }
+
+  scope :filter_count, lambda { |field|
```
zurbergram: is this needed since switching to `yellow_ribbon_program`?

db/migrate/20200224205220_add_columns_to_yellow_ribbon_program.rb
```diff
@@ -0,0 +1,35 @@
+class AddColumnsToYellowRibbonProgram < ActiveRecord::Migration[5.2]
```
zurbergram: Please make a separate PR with just this migration and the `db/schema.rb` change which will be merged into master first.

app/models/yellow_ribbon_program_source.rb
kelsonic: `search` is used by the controller, but the others can be omitted. Nice catch, thanks! 🙂

app/models/yellow_ribbon_program_source.rb
kelsonic: Nope! Nice catch 😎

app/models/yellow_ribbon_program_source.rb
kelsonic: Nope! Nice catch 😎

app/models/yellow_ribbon_program.rb
kelsonic: Hmm, so technically we can handle the scenario if all of these fields are missing on the frontend, so I think I can remove all of the validations tbh.

app/models/yellow_ribbon_program_source.rb
zurbergram: Isn't the controller at line `YellowRibbonProgram.search(@query[:name])` using a different model object than this one?

kelsonic: Ah whoops, sorry I was looking at `models/yellow_ribbon_program.rb`. I'll omit this as it's not needed as well as the filter lambdas on `yellow_ribbon_program_source` 👍

app/models/yellow_ribbon_program.rb
```diff
@@ -1,4 +1,44 @@
...
+  delegate :country, to: :institution
+
+  validates :contribution_amount, numericality: true
+  validates :degree_level, presence: true
+  validates :division_professional_school, presence: true
+  validates :facility_code, presence: true
+  validates :number_of_students, numericality: true
+
+  # Finds exact-matching facility_code or partial-matching school and city names
+  scope :search, lambda { |query|
```
kelsonic: This needs to be refactored at some point for code cleanliness 🙂

kelsonic:
@department-of-veterans-affairs/backend-review-group This PR is ready for review! 🙂

ncksllvn (approved):

app/controllers/v0/yellow_ribbon_programs_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      # Derive the order properties.
+      order_properties = {}
+      order_properties[sanitized_sort_by] = sanitized_sort_direction
+
+      # Sort the results.
+      results.order(order_properties)
+    end
+
+    def sanitized_sort_by
+      YellowRibbonProgram.last.attributes.include?(@query[:sort_by]) ? @query[:sort_by] : :school_name_in_yr_database
```
jholton: This suggested change:
* Doesn't require a record to exist in the table
* Saves a SQL query
```suggestion
      YellowRibbonProgram.new.attributes.include?(@query[:sort_by]) ? @query[:sort_by] : :school_name_in_yr_database
```

kelsonic: Awesome! Thank you!! Making the update now

app/models/institution_builder.rb
```diff
@@ -450,10 +450,30 @@ def self.add_yellow_ribbon_programs(version_number)
...
+            yrs.new_school, yrs.notes, yrs.open_ended_agreement,
+            yrs.public_private, yrs.school_name_in_weams,
+            yrs.school_name_in_yr_database, yrs.sco_email_address, yrs.sco_name,
+            yrs.sco_telephone_number, yrs.sfr_email_address, yrs.sfr_name,
+            yrs.sfr_telephone_number, yrs.state, yrs.street_address,
+            yrs.updated_for_2011_2012, yrs.withdrawn, yrs.year_of_yr_participation,
+            yrs.zip, NOW(), NOW()
           FROM institutions as i, yellow_ribbon_program_sources as yrs, versions
           WHERE i.facility_code = yrs.facility_code
           AND versions.number = #{version_number}
```
jholton: Given recent updates, you can now change this `add_yellow_ribbon_programs` method to accept `version.id` instead of `version.number` and eliminate a join with the `versions` table:
```
          FROM institutions as i, 
          INNER JOIN yellow_ribbon_program_sources as yrs
          ON i.facility_code = yrs.facility_code
          WHERE i.version_id = #{version_id}
```

app/models/yellow_ribbon_program.rb
```diff
@@ -1,4 +1,44 @@
...
+          city: "%#{query['city']}%", # (includes)
+          country: (query['country']).to_s, # (is equal)
+          school_name_in_yr_database: "%#{query['school_name_in_yr_database']}%", # (includes)
+          state: (query['state']).to_s # (is equal)
+        ]
+      )
+    )
+  }
+
+  scope :version, ->(n) { where(version: n) }
```
jholton: Is this used anywhere?

app/models/yellow_ribbon_program_source.rb
```diff
@@ -4,48 +4,50 @@ class YellowRibbonProgramSource < ApplicationRecord
...
 
-  validates :facility_code, presence: true
+  validates :contribution_amount, numericality: true
   validates :degree_level, presence: true
   validates :division_professional_school, presence: true
+  validates :facility_code, presence: true
   validates :number_of_students, numericality: true
-  validates :contribution_amount, numericality: true
+
+  scope :version, ->(n) { where(version: n) }
```
jholton: It doesn't look like this model has a `version` attribute.

jholton (changes requested):

spec/controllers/v0/yellow_ribbon_programs_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+require 'rails_helper'
+
+RSpec.describe V0::YellowRibbonProgramsController, type: :controller do
+  context 'when determining version' do
+    it 'uses a production version as a default' do
+      create(:version, :production)
+      create(:yellow_ribbon_program)
+      get(:index)
+      expect(response.content_type).to eq('application/json')
+      expect(response).to match_response_schema('yellow_ribbon_program')
```
jholton: What is this testing?  You're not checking the version returned by the results.

jholton:
```ruby
    it 'uses a production version as a default' do
      create(:version, :production)
      preview = create(:version, :preview)
      create(:yellow_ribbon_program)
      create(:yellow_ribbon_program, version: preview.number)
      get(:index)
      body = JSON.parse response.body
      # we have one record associated with the production version, so we expect one result.
      # this fails because we are getting back both the production and preview records
      expect(body['data'].count).to eq(1)
    end
```

app/models/institution_builder.rb
kelsonic: Just to clarify, you want me to update it to the above code snippet you provided? 🙂

spec/controllers/v0/yellow_ribbon_programs_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      body = JSON.parse response.body
+      expect(body['meta']['version']['number'].to_i).to eq(Version.current_production.number)
+    end
+  end
+
+  context 'when searching' do
+    before do
+      create(:version, :production)
```
jholton: I'd add a preview record here too to validate your result counts.

app/models/institution_builder.rb
jholton: This one is just a suggestion. Saves you a join.  
```ruby
  def self.run_insertions(version)
    ...
    add_yellow_ribbon_programs(version.id)
    ...
```
```ruby
def self.add_yellow_ribbon_programs(version_id)
```
And end the SQL with
```
          FROM institutions as i, 
          INNER JOIN yellow_ribbon_program_sources as yrs
          ON i.facility_code = yrs.facility_code
          WHERE i.version_id = #{version_id}
```

app/models/yellow_ribbon_program.rb
kelsonic: You know, I'm not really sure, I just tried to copy the pattern that was done [with the Institution model](https://github.com/department-of-veterans-affairs/gibct-data-service/blob/4fc38e21bd0e556fc6cfc9b84eef316e8f4b7e70/app/models/institution.rb) so it's likely superfluous! I'll attempt to omit it 🙂

app/models/yellow_ribbon_program_source.rb
kelsonic: Omitted! Thanks!

app/models/yellow_ribbon_program.rb
kelsonic: Omitted successfully! Thanks!

kelsonic:
Thanks for the comments/suggestions! Ready for re-review! 🎉

spec/controllers/v0/yellow_ribbon_programs_controller_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    it 'uses a production version as a default' do
+      create(:version, :production)
+      preview = create(:version, :preview)
+      create(:yellow_ribbon_program)
+      create(:yellow_ribbon_program, version: preview.number)
+      get(:index)
+      body = JSON.parse response.body
+      # we have one record associated with the production version, so we expect one result.
+      # this fails because we are getting back both the production and preview records
+      expect(body['data'].count).to eq(1)
```
jholton: It's failing.  😄 
You need to account for the version in your `YellowRibbonProgram.search`.
You might want to put back the `version` scope that you just removed.  I commented on it because it wasn't being used anywhere, but it looks like this is where it _should_ be used.
Something like `YellowRibbonProgram.version(version).search(params)`

kelsonic: Updated! 🙂

app/controllers/v0/yellow_ribbon_programs_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+        query[:sort_direction].try(:strip!)
+        query[:sort_direction].try(:downcase!)
+      end
+    end
+
+    def search_results
+      @query ||= normalized_query_params
+
+      # Derive the search results.
+      results = YellowRibbonProgram.version(@version).search(@query)
```
jholton: `@version` is a `Version` object.  The `YellowRibbonProgram#version` attribute is an integer.
So, this should be
```suggestion
      results = YellowRibbonProgram.version(@version.number).search(@query)
```
Your specs are passing because when you run them `@version.id` and `@version.number` happen to be the same integer, and Rails uses the `id` in the SQL it generates.

spec/controllers/v0/yellow_ribbon_programs_controller_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  context 'when determining version' do
+    it 'uses a production version as a default' do
+      create(:version, :production)
+      preview = create(:version, :preview)
+      create(:yellow_ribbon_program)
+      create(:yellow_ribbon_program, version: preview.number)
+      get(:index)
+      body = JSON.parse response.body
+      # we have one record associated with the production version, so we expect one result.
+      # this fails because we are getting back both the production and preview records
```
jholton: Can you remove these comments?  They aren't accurate now since the spec does not fail (which is what we want).

app/controllers/v0/yellow_ribbon_programs_controller.rb
kelsonic: So I think we want to use `@version.id`, not `@version.number`, as per one of your last comments, right? https://github.com/department-of-veterans-affairs/gibct-data-service/pull/584#discussion_r387851540

I made the update to use `version.id` explicitly in the model! 🙂

kelsonic: Updated to use `version.number` due to:

image

jholton (approved):

PR #590, opened 2020-02-26
GIDS: Institution CSV is blank on export #6201

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6201

Issue was that export of versioned tables was using `version.id` instead of `version.number` which is what is passed in to the method

```
SELECT * FROM "institutions" WHERE "institutions"."version_id" = $1  [["version_id", <version_number>]]
```
vs
```
SELECT * FROM "institutions" JOINS "versions" ON "institutions"."version_id" = "versions.id" WHERE "versions"."number" = $1  [["number", <version_number>]]
```
## Testing done
To test 
- generate a preview
- in SQL change the number value of the verisons row that was just created from preview generation so that id does not equal number for the row
- refresh dashboard, preview should now show the new "number" you set
- export the preview version

## Acceptance criteria
- [ ]

## Definition of done


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

.rubocop.yml
```diff
@@ -48,7 +48,17 @@ Style/ClassAndModuleChildren:
...
+Style/HashEachMethods:
+  Enabled: true
+# https://rubocop.readthedocs.io/en/latest/cops_style/#stylehashtransformkeys
+Style/HashTransformKeys:
+  Enabled: true
+# https://rubocop.readthedocs.io/en/latest/cops_style/#stylehashtransformvalues
+Style/HashTransformValues:
+  Enabled: true
+
+Layout/LineLength:
```
omgitsbillryan: Why the rubocop changes?

zurbergram: running `rake ci` outputted below so went ahead and configured them

```
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file:
 - Style/HashEachMethods
 - Style/HashTransformKeys
 - Style/HashTransformValues
```
https://rubocop.readthedocs.io/en/latest/cops_style/#stylehasheachmethods
https://rubocop.readthedocs.io/en/latest/cops_style/#stylehashtransformkeys
https://rubocop.readthedocs.io/en/latest/cops_style/#stylehashtransformvalues

omgitsbillryan: Ok, I see. Was rubocop updated recently or something to cause this?

omgitsbillryan: Also, I see that `Metrics/LineLength` was removed and `Layout/LineLength` was added. Was this renamed in a newer version or something?

zurbergram: https://github.com/rubocop-hq/rubocop/releases/tag/v0.78.0
```
#7542: (Breaking) Move LineLength cop from Metrics department to Layout department. (@koic)
```

When running `rake ci`
```
.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
```

zurbergram: removed

spec/models/institution_spec.rb
```diff
@@ -185,4 +185,40 @@ def create_institution(institution, description)
...
   end
+
+  describe 'when exporting' do
+    let(:mapping) { described_class::CSV_CONVERTER_INFO }
+
+    # Pull the default CSV options to be used
+    default_options = Rails.application.config.csv_defaults[described_class.name] ||
+                      Rails.application.config.csv_defaults['generic']
+    # Merge with provided options
+    load_options = default_options.each_with_object({}) { |(k, v), o| o[k.to_sym] = v; }
```
omgitsbillryan: I'm a bit confused by what you mean when you say

> Merge with provided options

It kind of looks like you're just converting the hash keys from strings to symbols, e.g.:
```
{"skip_lines"=>0, "col_sep"=>","}
  ...becomes...
{:skip_lines=>0, :col_sep=>","}
```

omgitsbillryan: In which case, you may want to use `default_options.symbolize_keys!`

zurbergram: fixed
copy paste error from `spec/models/shared_examples/shared_examples_for_loadable.rb`

institutions table is not created via CSV import but during the process in `app/models/institution_builder.rb` however without a test for exporting we missed the issue that is being fixed

spec/models/institution_spec.rb
```diff
@@ -185,4 +185,39 @@ def create_institution(institution, description)
...
+
+        rows[i].each.with_index { |value, j| attributes[mapping[header_row[j]][:column]] = value }
+        csv_record = described_class.new(attributes)
+        csv_record.derive_dependent_columns if csv_record.respond_to?(:derive_dependent_columns)
+
+        csv_test_attributes = csv_record.attributes.except('id', 'version', 'created_at', 'updated_at', 'csv_row', 'version_id')
+        test_attributes = record.attributes.except('id', 'version', 'created_at', 'updated_at', 'csv_row', 'version_id')
+        test_attributes['ope'] = "\"#{test_attributes['ope']}\"" if test_attributes['ope']
+
+        expect(csv_test_attributes).to eq(test_attributes)
```
omgitsbillryan: I personally prefer to see the expectation declared directly within the spec example, but not that important.

omgitsbillryan (approved):
The 1-line code change in the module is 👍 

However, I'm largely unable to follow / understand what's going on in that spec 😬 , but I suppose it's just a spec and I don't want to hold things up. Since the change was only 1 LoC, I would have expected that we should have been able to update an existing spec example rather than creating a new one. At the very least, I didn't expect that so much new spec plumbing work would need to get setup.

spec/models/institution_spec.rb
zurbergram: Was able to remove that line as was not necessary, it's used in shared_examples_for_loadable to merge provided options with default options but here it's not needed

PR #591, opened 2020-02-26
GIDS: Update Data Requirements Display#5752

ksantiagoBAH:
## Description
As an EDU Team Member, I need it to be clear what the data requirements are for each file when uploading so that I can address any reported issues without having to navigate to multiple pages.

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/5752

## Testing done
Local Testing Completed

## Acceptance criteria
- [x] The formatting of the data requirements help section is formatted using order lists and bullets (see Supporting Artifact # 1 for reference).
- [x] The data requirements help section is displayed on the file upload success page (see Supporting Artifact # 2 for reference).

## Definition of done


app/controllers/uploads_controller.rb
```diff
@@ -9,42 +9,48 @@ def new
...
+      csv_requirements
+
       return
     end
-
     alert_and_log(@upload.errors.full_messages.join(', '))
     redirect_to dashboards_path
   end
 
+  def csv_requirements
```
zurbergram: can you move this method down below the `private` ?

app/controllers/uploads_controller.rb
```diff
@@ -9,42 +9,48 @@ def new
...
-      @requirements = requirements_messages if @upload.csv_type_check?
       render :new
-    end
+   end
   end
 
   def show
     @upload = Upload.find_by(id: params[:id])
+
+    csv_requirements if @upload
```
zurbergram:
```suggestion
    csv_requirements if @upload.present?
```

app/controllers/uploads_controller.rb
```diff
@@ -97,32 +103,75 @@ def klass
...
+  def batch
+    custom_batch_validator_messages = {
+      message: 'Requirement Description:',
+      value: "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
+    }
+    if custom_batch_validator_messages[:value].nil?
+      {}
+    else
+      [custom_batch_validator_messages]
+    end
```
zurbergram: Change the if/else block to
```suggestion
    [custom_batch_validator_messages] if custom_batch_validator_messages[:value].present?
```

app/controllers/uploads_controller.rb
```diff
@@ -97,32 +103,75 @@ def klass
...
+  end
+
+  def validation_messages_inclusion
+    inclusion = []
+
     klass.validators.map do |validations|
       case validations
-      when ActiveRecord::Validations::PresenceValidator
-        generic_requirement_message('These columns must have a value: ', validations)
       when ActiveModel::Validations::InclusionValidator
```
zurbergram: change this to an if statement since only one case now

app/controllers/uploads_controller.rb
```diff
@@ -97,32 +103,75 @@ def klass
...
+        array = { message: affected_attributes(validations), value: inclusion_requirement_message(validations) }
+        inclusion.push(array)
       end
     end
+
+    if !inclusion.empty?
+      [inclusion]
+    else
+      inclusion = {}
+    end
```
zurbergram: Change the if/else block to

```suggestion
    [inclusion] unless inclusion.empty?
```

app/controllers/uploads_controller.rb
```diff
@@ -97,32 +103,75 @@ def klass
...
+
+    return [numericality] unless numericality[:value].empty?
+  end
+
+  def validation_messages_uniqueness
+    uniqueness = { message: 'These columns should contain unique values: ', value: [] }
+
+    uniqueness[:value] = klass_validator(ActiveRecord::Validations::UniquenessValidator)
+
+    return [uniqueness] unless uniqueness[:value].empty?
```
zurbergram:
```suggestion
   [uniqueness] unless uniqueness[:value].empty?
```

app/controllers/uploads_controller.rb
```diff
@@ -97,32 +103,75 @@ def klass
...
+
+    return [presence] unless presence[:value].empty?
+  end
+
+  def validation_messages_numericality
+    numericality = { message: 'These columns can only contain numeric values: ', value: [] }
+
+    numericality[:value] = klass_validator(ActiveModel::Validations::NumericalityValidator)
+
+    return [numericality] unless numericality[:value].empty?
```
zurbergram:
```suggestion
   [numericality] unless numericality[:value].empty?
```

app/controllers/uploads_controller.rb
ksantiagomorgan6: done

app/models/program.rb
```diff
@@ -14,6 +14,7 @@ class Program < ApplicationRecord
     'length' => { column: :length, converter: BaseConverter }
   }.freeze
 
-  validates :facility_code, :description, presence: true
+  validates :facility_code, presence: true
+  validates :description, presence: true
```
zurbergram: do these need to be on a separate line?

app/models/upload.rb
```diff
@@ -116,9 +116,14 @@ def self.fetching_for?(csv_type)
     Upload.where(ok: false, completed_at: nil, csv_type: csv_type).any?
   end
 
+  # def self.valid_col_seps
```
zurbergram: remove the commented out code

app/views/uploads/new.html.erb
```diff
@@ -9,17 +9,52 @@
...
       <%= render 'form' %>
     </div>
-    <% if @requirements&.any? %>
-      <div class="col-xs-4">
-        <div class="alert alert-info" role="alert">
-          <h4 class="alert-heading">CSV Data Requirements</h4><br />
-          <ul>
-            <% @requirements.each do |validations| %>
-              <li><%= validations %></li>
+        <% if @requirements&.any? %>
```
zurbergram: Should this be broken out into a partial since it is displayed in both new and show?

app/views/uploads/show.html.erb
```diff
@@ -66,3 +66,54 @@
     <%= @upload.comment %>
   </div>
 </div>
+
+<div class="container" style="float: left;">
```
zurbergram: Should this be broken out into a partial since it is displayed in both new and show?

spec/controllers/uploads_controller_spec.rb
```diff
@@ -79,28 +79,39 @@ def map_attributes(csv_class, requirement_class)
       requirements(csv_class, requirement_class)
         .attributes
         .map { |column| csv_class::CSV_CONVERTER_INFO.select { |_k, v| v[:column] == column }.keys.join(', ') }
-        .join(', ')
     end
 
+    # def map_attributes(csv_class, requirement_class)
```
zurbergram: removed commented out code

app/models/upload.rb
```diff
@@ -116,9 +116,14 @@ def self.fetching_for?(csv_type)
...
+  #   valid_col_seps = Settings.csv_upload.column_separators.map { |cs| "\"#{cs}\"" }.join(' and ')
+  #   "Valid column separators are: #{valid_col_seps}."
+  # end
+
   def self.valid_col_seps
```
zurbergram: how does changing this method affect the usage in the method `Upload.set_col_sep` down below in this file?

app/models/program.rb
ksantiagomorgan6: Yes there are quite a few like this, i'll fix all the models

app/controllers/uploads_controller.rb
```diff
@@ -131,12 +181,11 @@ def csv_column_name(column)
     klass::CSV_CONVERTER_INFO.select { |_k, v| v[:column] == column }.keys.join(', ')
   end
 
-  def generic_requirement_message(message, validations)
-    message + affected_attributes(validations)
+  def generic_requirement_message(validations)
```
zurbergram: this method isn't needed since it just calls `affected_attributes(validations)`

app/models/hcm.rb
```diff
@@ -14,7 +14,9 @@ class Hcm < ApplicationRecord
     'method reason desc' => { column: :hcm_reason, converter: BaseConverter }
   }.freeze
 
-  validates :ope, :hcm_type, :hcm_reason, presence: true
+  validates :ope, presence: true
```
zurbergram: undo these changes

app/models/ipeds_ic.rb
```diff
@@ -121,7 +121,10 @@ class IpedsIc < ApplicationRecord
   }.freeze
 
   validates :cross, presence: true
-  validates :vet2, :vet3, :vet4, :vet5, inclusion: { in: (-2..1) }
+  validates :vet2, inclusion: { in: (-2..1) }
```
zurbergram: undo these changes

app/models/mou.rb
```diff
@@ -16,7 +16,8 @@ class Mou < ApplicationRecord
     'approval date' => { column: :approval_date, converter: BaseConverter }
   }.freeze
 
-  validates :ope, :ope6, presence: true
+  validates :ope, presence: true
```
zurbergram: undo do these changes

app/models/program.rb
```diff
@@ -14,6 +14,7 @@ class Program < ApplicationRecord
     'length' => { column: :length, converter: BaseConverter }
   }.freeze
 
-  validates :facility_code, :description, presence: true
+  validates :facility_code, presence: true
```
zurbergram: undo do these changes

app/models/settlement.rb
```diff
@@ -14,4 +14,5 @@ class Settlement < ApplicationRecord
   }.freeze
 
   validates :cross, :settlement_description, presence: true
+  validates :settlement_description, presence: true
```
zurbergram: undo do these changes

app/models/weam.rb
```diff
@@ -74,7 +74,9 @@ class Weam < ApplicationRecord
   }.freeze
 
   has_many :crosswalk_issue, dependent: :delete_all
-  validates :facility_code, :institution, :country, presence: true
+  validates :facility_code, presence: true
```
zurbergram: undo do these changes

zurbergram (changes requested):
Please undo the changes to all the models and instead update how the method `affected_attributes` works which might require a tweak to how `validation_messages_inclusion` works as well

zurbergram (approved):

sonntag-adam (approved):

app/controllers/uploads_controller.rb
```diff
@@ -97,46 +103,83 @@ def klass
...
+  def batch
+    custom_batch_validator_messages = {
+      message: 'Requirement Description:',
+      value: "#{klass.name}Validator::REQUIREMENT_DESCRIPTIONS".safe_constantize
+    }
+    if custom_batch_validator_messages[:value].nil?
+      {}
+    else
+      [custom_batch_validator_messages]
+    end
```
omgitsbillryan: This if-else block doesn't seem to do anything. Am I missing something?

omgitsbillryan (commented):
Mind updating the description to be more specific to what changes were made in code? i.e. "added `foo` param to `bar_controller`"

Right now it looks to be a copy+paste from the associated ticket.

app/models/upload.rb
```diff
@@ -148,8 +148,8 @@ def csv_file_headers
   def set_col_sep(first_line)
     self.col_sep = Settings.csv_upload.column_separators
                            .find { |column_separator| first_line.include?(column_separator) }
-
-    error_message = "Unable to determine column separator. #{Upload.valid_col_seps}"
+    valid_col_seps = Upload.valid_col_seps[:value].map { |cs| "\"#{cs}\"" }.join(' and ')
+    error_message = "Unable to determine column separator. #{valid_col_seps}"
```
omgitsbillryan: Does `valid_col_seps` represent valid column separators? The error message language doesn't make this obvious. Maybe something like,

```
Unable to determine column separator. Valid separators = [#{valid_col_seps}].
```

app/controllers/uploads_controller.rb
```diff
@@ -9,42 +9,42 @@ def new
     @upload = Upload.from_csv_type(params[:csv_type])
 
     if @upload.csv_type_check?
-      @requirements = requirements_messages
+      csv_requirements
+
       return
```
omgitsbillryan: Not related to PR necessarily, but I'm trying to better understand.

What does this route return for an http response? Also, what does a blank `return` do from within a controller?

omgitsbillryan (commented):
I'm a bit confused. There's a lot of validation happening in the controller that I would expect to see in a validator. Is using vanilla models + validators not an option?

app/controllers/uploads_controller.rb
ksantiagomorgan6: That should of been removed, cleaned it up

app/controllers/uploads_controller.rb
ksantiagomorgan6: I updated it so it makes more sense with the if statement returning the csv_requirements

ksantiagoBAH:
> I'm a bit confused. There's a lot of validation happening in the controller that I would expect to see in a validator. Is using vanilla models + validators not an option?

I had to use these validators to get the data necessary to display correct info in the views.

app/controllers/uploads_controller.rb
```diff
@@ -109,46 +112,78 @@ def klass
...
+    numericality = { message: 'These columns can only contain numeric values: ', value: [] }
+
+    numericality[:value] = klass_validator(ActiveModel::Validations::NumericalityValidator)
+
+    [numericality] unless numericality[:value].empty?
   end
 
-  def validation_messages
+  def validation_messages_uniqueness
+    uniqueness = { message: 'These columns should contain unique values: ', value: [] }
```
omgitsbillryan: `'These columns should contain unique values: '` kind of looks like "view" code to me. If we were able to move the `message`'s over to our views, we might be able to do away with the arbitrary `{message:, value:}` data structure altogether which would the objects we're passing to our view a bit more straightforward. It would also make a lot of the other code in this controller that's responsible for creating and manipulating this data structure simpler.

omgitsbillryan (approved):
I don't want to block so, LGTM.

My main objection that this introduces a lot of bloat in the controller that looks like it doesn't belong there. I think a lot of the logic around validations could be moved to the models, and we shouldn't need to create that arbitrary `{message:, value:}` data structure. This along with some comments I made inline that I leave to your discretion.

PR #605, opened 2020-03-10
Bah 6666

elder-joseph:
## Description
As a developer, I need to improve the performance of the institution_builder queries in order to enhance the performance and stability of the application.

Note: 
```
Instead of 2 loops could switch last part of query in https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/models/institution_builder.rb#L172 to be an IN statement
``` 
was not implemented since the loop is how the hierarchy for `accreditation_type` needs to be set in that particular order, which cannot be done with and `IN` statement.

[Originating Issue](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/6666)

## Testing done
Testing passes locally

## Screenshots
N/A

## Acceptance criteria
- [x] Institution_builder queries have been optimized.

## Definition of done


db/migrate/20200310150006_add_indexes_to_institution_builder.rb
```diff
@@ -0,0 +1,10 @@
+class AddIndexesToInstitutionBuilder < ActiveRecord::Migration[5.2]
```
zurbergram: Should these be broken into a migration per table?

elder-joseph: I can break them out per table.

app/models/institution_builder.rb
```diff
@@ -516,19 +515,18 @@ def self.build_zip_code_rates_from_weams(version_number)
...
-      v.number,
       #{conn.quote(timestamp)},
       #{conn.quote(timestamp)},
-      v.id
-      FROM weams INNER JOIN versions v ON v.number = ?
+      #{version_id}
+      FROM weams
     WHERE country = 'USA'
       AND bah IS NOT null
       AND dod_bah IS NOT null
```
shawkey-daniel: Don't we need to filter by version_id?

shawkey-daniel (changes requested):

app/models/institution_builder.rb
elder-joseph: Do you mean adding something like `AND version_id = #{version_id}`

I don't think that is actually necessary here since we are passing in `version_id` as a parameter here to set it for `zipcode_rates` and `weams` does not have a `version_id`

shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

johnpaulashenfelter (approved):
👍 

That's a lot of migrations to remember to run on deploy.

PR #630, opened 2020-03-26
CF: Mapping Caution Flag Content #7151

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/7151

Database PR: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/634
Gem PR: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/633
https://github.com/ulfurinn/wongi-engine/

## Testing done
- locally run `bundle install` and `bundle exec rake db:migrate`
- locally run `bundle exec rake db:seed` to create the mapping rules
- locally run gibct-data-service, vets-api, and vets-website
- http://localhost:3001/gi-bill-comparison-tool/profile/31002011 or http://localhost:3001/gi-bill-comparison-tool/ and look for institutions with caution flags
- Can check in local database for any caution flags that didn't have a rule applied to them by doing (the not like is because of https://github.com/department-of-veterans-affairs/va.gov-team/issues/7440)
```
select reason
	from caution_flags cf join versions v on v.id = cf.version_id
	where title is null
	and v.number = <A VERSION NUMBER>
	and reason not like '%,%'
```
## Screenshots
Screen Shot 2020-04-01 at 11 54 04 AM

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #634, opened 2020-04-01
CF: Mapping Caution Flag Content: Database changes

zurbergram:
## Description
Main PR: https://github.com/department-of-veterans-affairs/gibct-data-service/pull/630

## Acceptance criteria
- [ ]

## Definition of done


shawkey-daniel (approved):
LGTM

db/migrate/20200327142848_create_caution_flag_rules.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.timestamps
+    end
+
+    add_foreign_key :caution_flag_rules, :rules, validate: false
+  end
+
+  def down
+    drop_table :caution_flag_rules
+
+    remove_foreign_key :caution_flag_rules, :rules
```
jholton, line 18: If you try to rollback you get an error.  If you're dropping the table, you don't need to also manually remove the foreign key.

db/migrate/20200327142848_create_caution_flag_rules.rb
```diff
@@ -0,0 +1,20 @@
+class CreateCautionFlagRules < ActiveRecord::Migration[5.2]
+  def up
+    create_table :caution_flag_rules do |t|
+      t.integer :rule_id
```
jholton, line 4: Rails makes these foreign keys `bigint`s now.  And this doesn't get you an index on `rule_id`.  Try this instead:
```ruby
    create_table :caution_flag_rules do |t|
      t.belongs_to :rule, foreign_key: true
      t.string :title
      t.string :description
      t.string :link_text
      t.string :link_url
      t.timestamps
    end
```
and you don't need the `add_foreign_key` below either.

zurbergram: changed

db/migrate/20200327142848_create_caution_flag_rules.rb
zurbergram: oh cause the order is wrong, removed manually removing the foreign key

jholton (approved):

PR #649, opened 2020-04-27
CF: Production duplicates of caution flag#8413

zurbergram:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8413

## Testing done
- go to http://localhost:4000/v0/institutions/21823010

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/app/models/caution_flag.rb
+++ b/app/models/caution_flag.rb
@@ -2,8 +2,8 @@
 
 class CautionFlag < ApplicationRecord
   belongs_to :institution, counter_cache: :count_of_caution_flags
-  scope :unique_flag, -> {
-    select("DISTINCT title, description, link_text, link_url")
+  scope :unique_flag, lambda {
+    select('DISTINCT title, description, link_text, link_url')
   }
 
   def self.map(version_id)
--- a/app/models/caution_flag.rb
+++ b/app/models/caution_flag.rb
@@ -2,6 +2,9 @@
 
 class CautionFlag < ApplicationRecord
   belongs_to :institution, counter_cache: :count_of_caution_flags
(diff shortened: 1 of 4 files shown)


shawkey-daniel (approved):
LGTM

app/models/caution_flag.rb
```diff
@@ -2,6 +2,9 @@
 
 class CautionFlag < ApplicationRecord
   belongs_to :institution, counter_cache: :count_of_caution_flags
+  scope :unique_flag, lambda {
```
erluti, line 5: I have a couple thoughts about the name for this scope.  I think it should be plural because it returns more than one in the relation.  Secondly I think it should have the word 'distinct' in it because using a `DISTINCT` in a query affects other things you could do to the relation, like call `count`.  So my recommendation would be to call this scope `distinct_flags`

app/models/caution_flag.rb
```diff
@@ -2,6 +2,9 @@
...
+    select('DISTINCT title, description, link_text, link_url')
```
erluti, line 6: I think this would be more rails-ish as `select('title, description, link_text, link_url').distinct`

philipwritescode (commented):
This is looking good.  I have a couple recommendations on the scope name and I would also like to see a test for it in the CautionFlag specs.

app/models/caution_flag.rb
zurbergram: changed

app/models/caution_flag.rb
zurbergram: changed

philipwritescode (approved):
looks good

PR #678, opened 2020-07-02
add physical zip and ialias criteria to free text institution search 

hughes-dustin:
## Description
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/10617

## Testing done
Dev tested.

## Screenshots
image

image

## Acceptance criteria
- [ ] The 'ialias' from the IPEDSHD is searchable within the Comparison Tool.
- [ ] The 'physical institution zip code' from the WEAMS data is searchable within the Comparison Tool.

## Definition of done

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -237,7 +237,7 @@ class Institution < ApplicationRecord
       'facility_code = (:facility_code)',
       'institution LIKE (:upper_search_term)',
       'city LIKE (:upper_search_term)'
-     ]
+    ]
 
     if fuzzy_search
       clause << "SIMILARITY(institution, :search_term) > #{Settings.institution_name_similarity_threshold}"
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -236,12 +236,17 @@ class Institution < ApplicationRecord
     clause = [
       'facility_code = (:facility_code)',
       'institution LIKE (:upper_search_term)',
-      'city LIKE (:upper_search_term)'
-    ]
+      'city LIKE (:upper_search_term)',
+      'ialias = (:search_term)',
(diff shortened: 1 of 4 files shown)


app/models/institution.rb
```diff
@@ -242,6 +242,9 @@ def self.autocomplete(search_term, limit = 6)
     if fuzzy_search
       clause << "SIMILARITY(institution, :search_term) > #{Settings.institution_name_similarity_threshold}"
       clause << "SIMILARITY(city, :search_term) > #{Settings.institution_city_similarity_threshold}"
+      clause << 'zip = (:search_term)'
+      clause << 'ialias LIKE (:search_term)'
+      clause << 'ialias LIKE upper(:search_term)'
```
shawkey-daniel, line 247: Is this needed? I didn't think it was case-sensitive.

shawkey-daniel (changes requested):

app/models/institution.rb
dmccurdybah: You can use `:upper_search_term` instead of `upper(:search_term)` and we could also import it as upper case and not need to check twice

hughes-dustin: There are cases where `ialias` data is stored in all caps like "CSU". Without the the upper, we won't get any results if a user searches for something like "csu".  For some reason, `upper_search_term` does not return any results from `ialias` matches when searching "csu".``

hughes-dustin: changes are in.

shawkey-daniel (approved):
LGTM

sonntag-adam (approved):

philipwritescode (approved):
LGTM

PR #679, opened 2020-07-02
Institution fuzzy search updates

devin-mccurdy:
## Description
Performance updates for institution "fuzzy search".

This is behind the `gibct_search_enhancements` feature flag, so updates aren't affecting the experience of production users. 

[ZenHub Issue](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/10613)

## Testing done

## Screenshots
N/A

## Acceptance criteria
- [x] Fuzzy Search works as expected

## Definition of done

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -156,7 +156,8 @@ class Institution < ApplicationRecord
     },
     'complies_with_sec_103' => { column: :complies_with_sec_103, converter: BooleanConverter },
     'solely_requires_coe' => { column: :solely_requires_coe, converter: BooleanConverter },
-    'requires_coe_and_criteria' => { column: :requires_coe_and_criteria, converter: BooleanConverter }
+    'requires_coe_and_criteria' => { column: :requires_coe_and_criteria, converter: BooleanConverter },
+    'poo status' => { column: :poo_status, converter: BaseConverter }
   }.freeze
 
   has_many :caution_flags, -> { distinct_flags }, inverse_of: :institution, dependent: :destroy
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -227,36 +227,31 @@ class Institution < ApplicationRecord
       .where('institution LIKE (?)', "#{search_term.upcase}%")
       .limit(limit)
   end
-
   # Finds exact-matching facility_code or partial-matching school and city names
   #
   scope :search, lambda { |search_term, include_address = false, fuzzy_search = false|
     return if search_term.blank?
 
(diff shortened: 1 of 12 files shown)


shawkey-daniel (approved):
LGTM

sonntag-adam (approved):

johnpaulashenfelter (approved):
if it handles your search, then 👍 :)

app/models/institution.rb
```diff
@@ -233,32 +233,29 @@ def self.autocomplete(search_term, limit = 6)
...
+      clause << 'ialias LIKE :upper_search_term'
+    else
+      clause << 'institution LIKE :upper_search_term'
+      clause << 'city LIKE :upper_search_term'
     end
 
     if include_address
       3.times do |i|
-        clause << "lower(address_#{i + 1}) LIKE (:lower_search_term)"
+        clause << "lower(address_#{i + 1}) LIKE :lower_search_term"
```
LindseySaari, line 250: Were the parenthesis causing issues here? I see that they're added on line 254 instead

dmccurdybah: That was just cleanup.  I'm not sure why every `like` had parenthesis, but now they should only be used where they're actually needed

app/models/institution.rb
```diff
@@ -233,32 +233,29 @@ def self.autocomplete(search_term, limit = 6)
...
-      'city LIKE (:upper_search_term)'
-    ]
+    clause = ['facility_code = :facility_code']
 
     if fuzzy_search
-      clause << "SIMILARITY(institution, :search_term) > #{Settings.institution_name_similarity_threshold}"
-      clause << "SIMILARITY(city, :search_term) > #{Settings.institution_city_similarity_threshold}"
-      clause << 'zip = (:search_term)'
-      clause << 'ialias LIKE (:upper_search_term)'
+      clause << 'institution % :upper_search_term'
```
LindseySaari, line 239: I see that the SIMILARITY and LIKE clauses for city and institution were replaced with the % wildcard character. Are the similarity thresholds no longer needed?

dmccurdybah: `SIMILARITY` compares the entire search term with the column value, instead of fully leveraging `pg_trgm`. We'll be adding the similarity thresholds back in, but outside of that clause so that we're only enforcing it on the subset of institutions that already match, instead of checking the similarity of ALL of them (which was causing the search results to exceed the staging timeout values)

PR #693, opened 2020-07-29
Change when fuzzy search with pgtrm is used

zurbergram:
## Description
Initial implementation wasn't correctly behind feature flag and didn't account for several cases.  

This only allows usage of fuzzy search if the combined count of conditions below is 0: 
-  a facility_code that equals search_term
- an institution name that starts with search_term
- an institution name that equals search_term
- a city that equals search_term
- an ialias that contains search_term

This also updates the non-fuzzy search search to use same where statements as the count

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -113,7 +113,6 @@ Metrics/ModuleLength:
     - "app/models/institution_builder.rb"
 
 Metrics/BlockLength:
-  Max: 40
   Exclude:
     - "spec/**/*"
     - "rakelib/**/*.rake"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -113,6 +113,7 @@ Metrics/ModuleLength:
     - "app/models/institution_builder.rb"
 
 Metrics/BlockLength:
+  Max: 40
   Exclude:
     - "spec/**/*"
     - "rakelib/**/*.rake"
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -267,7 +267,7 @@ class Institution < ApplicationRecord
                                        upper_search_term: search_term.upcase,
                                        upper_contains_term: "%#{search_term.upcase}%",
                                        lower_contains_term: "%#{search_term.downcase}%",
-                                       starts_with_term: "%#{search_term.upcase}",
+                                       starts_with_term: "#{search_term.upcase}%",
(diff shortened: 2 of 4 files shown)


shawkey-daniel (approved):
LGTM

.rubocop.yml
```diff
@@ -113,6 +113,7 @@ Metrics/ModuleLength:
     - "app/models/institution_builder.rb"
 
 Metrics/BlockLength:
+  Max: 40
```
omgitsbillryan, line 116: What's going on here?

app/models/institution.rb
```diff
@@ -230,31 +237,37 @@ def self.autocomplete(search_term, limit = 6)
...
 
-    clause = ['facility_code = :facility_code']
+    clause = ['facility_code = :upper_search_term']
 
-    if fuzzy_search
-      clause << 'SIMILARITY(institution, :search_term) > :name_threshold'
-      clause << 'SIMILARITY(city, :search_term) > :city_threshold'
-      clause << 'zip = :search_term'
+    if fuzzy_search_flag
+      if use_fuzzy_search
```
omgitsbillryan, line 248: I know I'm not very familiar with gibct, but what's the difference between `app/models/institution.rb` & `use_fuzzy_search`?

.rubocop.yml
zurbergram: just increasing value since default of 25 was causing lint failures

omgitsbillryan (approved):

app/models/institution.rb
zurbergram: `fuzzy_search_flag` comes from a Feature Flag. It is based on the status of https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/features.yml#158 when used in the frontend at https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/gi/containers/SearchPage.jsx

`use_fuzzy_search?` is an additional check that checks if the counts of certain criteria combined is higher than 0 and causes it to use more effective and efficient SQL when searching

zurbergram: the status of the feature flag is passed down as a parameter to gibct because feature flags are not a part of this repo as they are in vets-api

PR #714, opened 2020-09-01
Ratings upload

devin-mccurdy:
## Description
Add the ability to upload and export a school ratings CSV file.  This isn't used for anything yet, but will be fully implemented in future PRs.

Note: Will be merged after DB mibration PR: [713](https://github.com/department-of-veterans-affairs/gibct-data-service/pull/713)

[ZenHub Issue](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/12962)

## Testing done

## Screenshots
Dashboard:
image

Upload page:
image

Successful upload:
image

Validation Message:
image

## Acceptance criteria
- [x] A "school_ratings.csv" with the following data elements can be uploaded to GIDS:
  - a. rater_id (REQUIRED string), Note: The format of this isn't known, so string gives us the most flexibility.
  - b. facility_code (REQUIRED string)
  - c. overall_experience (int)
  - d. quality_of_classes (int)
  - e. online_instruction (int)
  - f. job_preparation (int)
  - g. gi_bill_support (int)
  - h. veteran_community (int)
  - i. marketing_practices (int)
  - j. rated_on (REQUIRED datetime), Note: rankers can rank institutions more than once, and this will be used to make sure we're only counting their most recent ranking.
- [x] Data rows are validated against the specifications in AC1:
  - a. If there is a data type mismatch the following message is displayed: "Row [Number]: [column name] : is not a number"
  - b. If there is a data type mismatch the following message is displayed: "Row [Number]: [column name] : can't be blank"
  - c. If the data does not meet the specifications, the row is not imported.
  - d. If a row contains multiple validation errors, they are all displayed on a single line.

## Definition of done


app/models/school_rating.rb
```diff
@@ -0,0 +1,20 @@
...
+    'rater id' => { column: :rater_id, converter: BaseConverter },
+    'facility code' => { column: :facility_code, converter: FacilityCodeConverter },
+    'overall experience' => { column: :overall_experience, converter: SchoolRatingConverter },
+    'quality of classes' => { column: :quality_of_classes, converter: SchoolRatingConverter },
+    'online instruction' => { column: :online_instruction, converter: SchoolRatingConverter },
+    'job preparation' => { column: :job_preparation, converter: SchoolRatingConverter },
+    'gi bill support' => { column: :gi_bill_support, converter: SchoolRatingConverter },
+    'veteran community' => { column: :veteran_community, converter: SchoolRatingConverter },
+    'marketing practices' => { column: :marketing_practices, converter: SchoolRatingConverter },
+    'rated on' => { column: :rated_on, converter: DateTimeConverter }
```
shawkey-daniel, line 16: Would 'rated date' be less ambiguous?

shawkey-daniel (changes requested):

app/models/school_rating.rb
dmccurdybah: I can change CSV column name, but I'd like the column name to match other rails columns...  which looks like this should be `rated_at` (instead of `rated_on`) to match `created_at` and `updated_at`

shawkey-daniel: I'm good with that

delli-gatti-michael (approved):

zurbergram (approved):

shawkey-daniel (approved):
LGTM

app/models/converters/date_time_converter.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class DateTimeConverter < BaseConverter
+  def self.convert(value)
+    value = super(value)
+
+    return nil if value.blank?
+
+    begin
+      DateTime.parse(value)
```
erluti, line 10: is there a reason this is not `Time.zone.parse(value)`?

app/models/converters/school_rating_converter.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SchoolRatingConverter < BaseConverter
+  def self.convert(value)
+    # allow non-numeric values and let validation handle it
+    return value unless value.is_a? Numeric
+
+    value = value.to_i
+    value = 5 if value > 5
```
erluti, line 9: take it or leave it, but written like this you can sometimes avoid the second check:
```
if value > 5
  value = 5
elsif value <= 0 # or < 1
 value = nil
end

File: spec/models/converters/date_time_converter_spec.rb [hunk 4]
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DateTimeConverter do
+  subject { described_class }
```
erluti, line 6: this is unused.  I think conventionally the tests should be using `subject` instead of repeating `described_class` below

spec/models/converters/date_time_converter_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  it 'converts strings to dates' do
+    date_string = '2020-01-01T12:05:02+08:00'
+    expect(described_class.convert(date_string)).to eq(DateTime.parse(date_string))
```
erluti, line 10: unless DateTime is preferable in DateTimeConverter, this should also be `Time.zone.parse`

spec/models/converters/school_rating_converter_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SchoolRatingConverter do
+  subject { described_class }
```
erluti, line 6: also not using `subject` in this test

philipwritescode (commented):
A few comments and questions

spec/models/converters/date_time_converter_spec.rb
dmccurdybah: Yes, you're right, I don't think this is necessary.

I think I can just remove this converter and let rails handle it and we'll be fine. For the data we have (which is just test data in a format we came up with since we have zero idea of what will eventually be provided), I think we're safe with that.

spec/models/converters/date_time_converter_spec.rb
dmccurdybah: I can remove line 6 and just use `described_class`, but in order to use `subject` explicitly we'd have to disable the rubocop `RSpec/NamedSubject` rule or it'd throw a linter error.  If you think that's a better solution I can make a separate PR to update `.rubocop.yml`, but I'd rather not have this wait on that if possible.

spec/models/converters/school_rating_converter_spec.rb
dmccurdybah: I can remove line 6 and just use described_class, but in order to use subject explicitly we'd have to disable the rubocop RSpec/NamedSubject rule or it'd throw a linter error. If you think that's a better solution I can make a separate PR to update .rubocop.yml, but I'd rather not have this wait on that if possible.

app/models/converters/date_time_converter.rb
dmccurdybah: I don't think we need this converter at all =)

philipwritescode (approved):
LGTM!

PR #732, opened 2020-09-12
Fix common word removal from processed institution search term

zurbergram:
## Description

https://github.com/department-of-veterans-affairs/va.gov-team/issues/13408

Fix the removal of common words, the fix ensures that for "maryland" "and" is not replaced out of the word

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -236,8 +236,8 @@ class Institution < ApplicationRecord
 
     processed_search_term = search_term.dup
     Settings.search.common_word_list.each do |word|
-      #  word.match(/[a-z]/i) is for characters like "&" and "-",
-      # the word boundary regex does not work with these characters
+      # word.match(/[a-z]/i) check is because of characters like "&" and "-",
+      # the word boundary regex "\b" does not replace these characters
       if word.match(/[a-z]/i)
         processed_search_term.gsub!(/\b#{Regexp.escape(word)}\b/i, '')
       else
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -237,7 +237,7 @@ class Institution < ApplicationRecord
     processed_search_term = search_term.dup
     Settings.search.common_word_list.each do |word|
       # word.match(/[a-z]/i) check is because of characters like "&" and "-",
-      # the word boundary regex "\b" does not replace these characters
+      # the word boundary regex "\b" does not work with these characters
       if word.match(/[a-z]/i)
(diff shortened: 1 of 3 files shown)


shawkey-daniel (approved):
LGTM

app/models/institution.rb
```diff
@@ -236,8 +236,13 @@ def self.institution_search_term(search_term)
 
     processed_search_term = search_term.dup
     Settings.search.common_word_list.each do |word|
-      processed_search_term.gsub!(/\b#{Regexp.escape(word)}\b/i, '') if word.match(/[a-z]/i)
-      processed_search_term.gsub!(/#{Regexp.escape(word)}/i, '')
+      # word.match(/[a-z]/i) check is because of characters like "&" and "-",
+      # the word boundary regex "\b" does not work with these characters
+      if word.match(/[a-z]/i)
```
erluti, line 241: This definitely works.
It struck me a little odd though because this regex is essentially, "does it have a letter in the string?"  But this is meant to be testing for specific strings that are single special characters.  So I took a step back, and I think this code could better use regular expressions to clean up the search terms.
I do introduce a separate settings entry for the single characters, but this builds one regular expression that can be run on the `processed_search_term` one time instead of looping through words.
```
COMMON_REMOVAL_REGEXP = Regexp.new(
  (
    Settings.search.common_character_list.map { |char| Regexp.escape(char) } +
    Settings.search.common_word_list.map { |word| "\b#{Regexp.escape(word)}\b" }
  ).join('|'),
  'i')
...
  def self.institution_search_term(search_term)
    return if search_term.blank?

    processed_search_term = search_term.gsub(COMMON_REMOVAL_REGEXP, '')
    
    return search_term.dup if processed_search_term.blank?

    processed_search_term.strip
  end
```
I'd put the constant at the top of this file so it's built once, and then every call of `institution_search_term` could benefit from the speed of regular expressions.

philipwritescode (approved):
This works so I don't want to be blocking, but I have left a comment on another possible approach

app/models/institution.rb
zurbergram: this would produce regex similar to what is in the image?
I'm trying to avoid the case where the "and" in "maryland" is replaced by the regex.

image


This is the other issue that was encountered is the `\b&\b` does not match the `&` in "university of maryland & and"

Screen Shot 2020-09-14 at 10 46 19 AM

https://rubular.com/

PR #751, opened 2020-10-26
GIDS: Add the ability to upload the accreditation excel file #13863

zurbergram:
## Description

https://github.com/department-of-veterans-affairs/va.gov-team/issues/13863

Adds new controller and views for a new type of uploading

specs https://github.com/department-of-veterans-affairs/gibct-data-service/pull/757

## Screenshots
Screen Shot 2020-10-28 at 1 05 45 PM
image
image

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #753, opened 2020-10-26
Refactor CSV type to be Upload types

zurbergram:
## Description

Defining groups csv/file types for upload from multi-tab excel spreadsheets is coming and this is refactor/setup work that can be broken out from https://github.com/department-of-veterans-affairs/gibct-data-service/pull/751

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/config/initializers/upload_types.rb
+++ b/config/initializers/upload_types.rb
@@ -1,0 +1,18 @@
+UPLOAD_TYPES = [
+    *CSV_TYPES_TABLES
+].freeze
+
+def klass_name(klass)
+  return klass if klass.is_a? String
+  klass.name
+end
+
+def klass_names(upload_types)
+  upload_types.map do |upload|
+    klass_name(upload[:klass])
+  end.freeze
+end
+
+UPLOAD_TYPES_REQUIRED_NAMES = klass_names(UPLOAD_TYPES.select { |upload| upload[:required?] })
+UPLOAD_TYPES_ALL_NAMES = UPLOAD_TYPES.map { |upload| klass_name(upload[:klass]) }.freeze
+UPLOAD_TYPES_NO_PROD_NAMES = klass_names(UPLOAD_TYPES.select { |upload| upload[:not_prod_ready?] })
--- a/config/initializers/upload_types.rb
+++ b/config/initializers/upload_types.rb
@@ -2,18 +2,20 @@ UPLOAD_TYPES = [
     *CSV_TYPES_TABLES
 ].freeze
 
-def klass_name(upload)
+UPLOAD_TYPES_ALL_NAMES = UPLOAD_TYPES.map do |upload|
   klass = upload[:klass]
   return klass if klass.is_a? String
   klass.name
-end
+end.freeze
 
-def klass_names(upload_types)
-  upload_types.map do |upload|
-    klass_name(upload)
-  end.freeze
-end
+UPLOAD_TYPES_REQUIRED_NAMES = UPLOAD_TYPES.select { |upload| upload[:required?] }.map do |upload|
+  klass = upload[:klass]
+  return klass if klass.is_a? String
+  klass.name
+end.freeze
 
-UPLOAD_TYPES_REQUIRED_NAMES = klass_names(UPLOAD_TYPES.select { |upload| upload[:required?] })
-UPLOAD_TYPES_ALL_NAMES = UPLOAD_TYPES.map { |upload| klass_name(upload) }.freeze
-UPLOAD_TYPES_NO_PROD_NAMES = klass_names(UPLOAD_TYPES.select { |upload| upload[:not_prod_ready?] })
+UPLOAD_TYPES_NO_PROD_NAMES = UPLOAD_TYPES.select { |upload| upload[:not_prod_ready?] }.map do |upload|
+  klass = upload[:klass]
+  return klass if klass.is_a? String
+  klass.name
+end.freeze
--- a/config/initializers/upload_types.rb
+++ b/config/initializers/upload_types.rb
@@ -2,17 +2,18 @@ UPLOAD_TYPES = [
     *CSV_TYPES_TABLES
 ].freeze
 
-def klass_name(klass)
+def klass_name(upload)
+  klass = upload[:klass]
   return klass if klass.is_a? String
   klass.name
 end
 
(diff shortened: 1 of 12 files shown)


shawkey-daniel (approved):
LGTM

spec/config/initializers/upload_types_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe 'required_table_names' do
+    it 'contains tables' do
+      expect(UPLOAD_TYPES_REQUIRED_NAMES).to eq(REQUIRED_TABLES)
+    end
+  end
+
+  describe 'no_prod_names' do
+    it 'contains tables' do
+      expect(UPLOAD_TYPES_NO_PROD_NAMES).to eq(NO_PROD_TABLES)
```
erluti, line 46: instead of defining a constant for this, could this just be `expect(UPLOAD_TYPES_NO_PROD_NAMES).to be_empty`?

config/initializers/upload_types.rb
```diff
@@ -0,0 +1,19 @@
+UPLOAD_TYPES = [
+    *CSV_TYPES_TABLES
```
erluti, line 2: is this list going to be more than `CSV_TYPES_TABLES`?  `dup` could work here.

config/initializers/upload_types.rb
```diff
@@ -0,0 +1,19 @@
...
+].freeze
+
+def klass_name(upload)
```
erluti, line 5: i think this method would be available in the whole app.  I think the old implementation wasn't too complex to need a method.  Also, there aren't any instances of the klass being a string, is there?  This seems unnecessary

config/initializers/upload_types.rb
zurbergram: Yes, in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/751 there is another array be splatted into this

I'm creating arrays that are merged because while both arrays will be displayed in same location the behavior of the "upload" button is different

spec/config/initializers/upload_types_spec.rb
zurbergram: In https://github.com/department-of-veterans-affairs/gibct-data-service/pull/751/files the `NO_PROD_TABLES` will have a value in it

config/initializers/upload_types.rb
zurbergram: The array being added in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/751 will have String values as there won't be a class to associate with it.

zurbergram: removed the methods

spec/config/initializers/csv_types_spec.rb
```diff
@@ -51,9 +17,13 @@
     end
   end
 
-  describe 'no_prod_names' do
-    it 'contains tables' do
-      expect(CSV_TYPES_NO_PROD_NAMES).to eq(NO_PROD_TABLES)
+  describe 'fields checks' do
+    CSV_TYPES_TABLES.each do |upload|
+      it "#{klass_name(upload)} csv type config has_api? is a boolean" do
+        if upload.respond_to?(:has_api?)
```
erluti, line 23: this line is violating a rubocop `Style/IfUnlessModifier: Favor modifier if usage when having a single-line body.`

spec/config/initializers/upload_types_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    UPLOAD_TYPES.each do |upload|
+      it "#{klass_name(upload)} upload type config has field required?" do
+        expect(upload[:required?]).to be_in([true, false])
+      end
+    end
+
+    UPLOAD_TYPES.each do |upload|
+      it "#{klass_name(upload)} upload type config not_prod_ready? is a boolean" do
+        if upload[:not_prod_ready?].present?
```
erluti, line 65: this line is violating a rubocop `Style/IfUnlessModifier: Favor modifier if usage when having a single-line body.`

spec/config/initializers/upload_types_spec.rb
erluti: What does `NO_PROD_TABLES` represent?  Should the list be in code somewhere instead of as a constant in the test?  This seems like it might trip someone up in the future if they don't know about this list to update when they do add "prod tables"?

philipwritescode (commented):
Looking good, make sure the CI passes

spec/config/initializers/csv_types_spec.rb
zurbergram: fixed

spec/config/initializers/upload_types_spec.rb
zurbergram: fixed

spec/config/initializers/upload_types_spec.rb
zurbergram: If an object in the array has `not_prod_ready?` set to true then at https://github.com/department-of-veterans-affairs/gibct-data-service/blob/master/app/views/dashboards/_latest_upload.html.erb#L1 if will hide the upload type from being visible in production environment. 

This repo does not use feature flags as they are not implemented. 
There is occasionally a need to hide an upload type from production while it is being tested in staging.

erluti: What if the test looked something like this?
```
describe 'no_prod_names' do
    it 'contains tables' do
      UPLOAD_TYPES.each do |upload_type|
        if upload_type[:not_prod_ready?]
          expect(UPLOAD_TYPES_NO_PROD_NAMES).to include(upload_type)
        else
          expect(UPLOAD_TYPES_NO_PROD_NAMES).to_not include(upload_type)
        end
      end
    end
  end
```
My concern is that a future dev would set `:not_prod_ready?` to `true` then the test would fail, and to pass it they would have to change the test put it in `NO_PROD_TABLES`, then when it is prod ready the dev making that update would have to change the test again.  So something like I proposed is testing the functionality and not adding a maintenance task.

zurbergram: changed both unit tests to take this approach for checking the subsets

philipwritescode (approved):
LGTM
thanks for improving those tests!

PR #771, opened 2020-11-18
Update Makefile

zurbergram:
## Description

`make ci` uses `docker-compose.test.yml` but `make rebuild` uses `docker-compose.yml` which causes issues

- adding in the ability to do `env=dev make ci` to have commands run using `docker-compose.yml`
  - Default for commands is to use `docker-compose.test.yml`, this minimizes changes needed
- Setting `CI: 'true'` in `docker-compose.yml` to disable seeding
- add a `make build` command
- add a `make down` command

## Testing done
For all commands testing was done both with `env=dev` before `make ${cmd}` command and without `env=dev`
- `make build` followed by `make ci`
  - When running `make build` the first line in the log should be `docker-compose -f docker-compose.test.yml build`
  - When running `env=dev build` the first line in the log should be `docker-compose build`, 
  - Database seeding should not occur for either one
- `make rebuild`
  - This calls both `make down` and `make build`
- `make db`
- `docker images rm gibct` then `make ci`

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,17 @@
 $stdout.sync = true
+
+ifdef env
+    ENV_ARG  := $(env)
+else
+    ENV_ARG	 := dev
+endif
+
 COMPOSE_DEV  := docker-compose
 COMPOSE_TEST := docker-compose -f docker-compose.test.yml
 BASH         := run --rm gibct bash --login
 BASH_DEV     := $(COMPOSE_DEV) $(BASH) -c
 BASH_TEST    := $(COMPOSE_TEST) $(BASH) -c
+DOWN         := down
 
 .PHONY: default
 default: ci
@@ -45,9 +53,23 @@ up: db
 	@$(COMPOSE_DEV) up
 
 .PHONY: rebuild
-rebuild:
-	@$(COMPOSE_DEV) down
-	@$(COMPOSE_DEV) build
+rebuild: down build
+
+.PHONY: build
+build:  ## Builds the service
+ifeq ($(ENV_ARG), dev)
+	$(COMPOSE_DEV) build
+else
+	$(COMPOSE_TEST) build
+endif
+
+.PHONY: down
+down:  ## Stops all docker services
+ifeq ($(ENV_ARG), dev)
+	@$(COMPOSE_DEV) $(DOWN)
+else
+	@$(COMPOSE_TEST) $(DOWN)
+endif
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ bash:
 	@$(COMPOSE_DEV) $(BASH)
 
 .PHONY: ci
-ci:
+ci: ## requires build to be run first, can do "env=test make ci" to run with docker-compose.test.yml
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -13,8 +21,12 @@ bash:
...
-ci:
+ci: ## requires build to be run first, can do "env=dev make ci" to run with docker-compose.yml
+ifeq ($(ENV_ARG), dev)
+	@$(BASH_DEV) "bin/rails db:setup db:migrate ci"
+else
 	@$(BASH_TEST) "bin/rails db:setup db:migrate ci"
+endif
 
 .PHONY: console
 console:
```
zurbergram, line 32: should i update these commands as well to default to `docker-compose.test.yml` and use an `ifeq` ?

shawkey-daniel (approved):
LGTM

docker-compose.yml
```diff
@@ -33,6 +33,7 @@ services:
       GOVDELIVERY_TOKEN: 'abc123'
       GOVDELIVERY_STAGING_SERVICE: 'True'
       DEPLOYMENT_ENV: 'vagov-dev'
+      CI: "true"
```
LindseySaari, line 36: I think single quotes are preferred here
```suggestion
      CI: 'true'
```

Makefile
```diff
@@ -1,9 +1,17 @@
 $stdout.sync = true
+
+ifdef env
+    ENV_ARG  := $(env)
+else
+    ENV_ARG	 := test
```
LindseySaari, line 6: Super nitpicky, but you may way to align the :=

zurbergram: changed, looks like i copied a tab over instead of spaces

docker-compose.yml
zurbergram: changed

Makefile
```diff
@@ -38,19 +50,42 @@ security: db
...
+	@$(COMPOSE_DEV) $(DOWN)
+else
+	@$(COMPOSE_TEST) $(DOWN)
+endif
 
 .PHONY: clean
 clean:
 	rm -r data || true
+ifeq ($(ENV_ARG), dev)
+	$(COMPOSE_DEV) run gibct rm -r coverage log/* tmp || true
```
LindseySaari, line 86: Do you want the clean command to remove volumes as well? I see it's just removing coverage, log/* and tmp

zurbergram: I'm not sure, this was an existing command that was updated to include the env option

LindseySaari (approved):

Makefile
LindseySaari: Additionally, you may want a true clean command `@$(COMPOSE_DEV) down --rmi all --volumes` which will remove any images or volumes hanging around

zurbergram: I tested out adding that as https://github.com/department-of-veterans-affairs/vets-api/blob/master/Makefile#L86 but because it blew away all images including some of those used by vets-api I decided to not include the command as vets-api is the more used repo and wanted to reduce people affecting their local vets-api when working in gibct-data-service

PR #790, opened 2021-02-18
19571: Create scorecard degree program (DB Specific PR)

elder-joseph:
## Description
As a developer, I need to update the Scorecard API fetch so that data needed for the redesign is available when we begin development of the new functionality.

This is the Database specific PR to add the necessary migration.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/19571

## Testing done
Testing passes locally.

## Screenshots
N/A

## Definition of done


shawkey-daniel (approved):
LGTM

db/schema.rb
```diff
@@ -1287,6 +1287,17 @@
     t.datetime "rated_at", null: false
   end
 
+  create_table "scorecard_degree_programs", force: :cascade do |t|
+    t.integer "unitid"
```
ericboehs, line 1291: Do we need to set a primary key here?

elder-joseph: In this case, I don't believe we would since this field is not required and will end up with many records with the same `unitid`.

ericboehs (approved):

sonntag-adam (approved):

PR #868, opened 2021-11-30
removed geocode mistmatch logic in institutions model and controller,…

PJart:
Needed to remove logic added to location search in the Institutions controller and model that regeocoded and filtered results based off the issue we found with some incorrect coordinates. Upon testing we found that solution was causing an issue with rendering the location search. I removed this logic and added a migration that fixes the incorrect latitude and longitude values we found related to a search on Webster Groves, Missouri and corrects them in the model using the geocoder gem.


bah-manish (approved):

db/migrate/20211130190839_fix_webster_grove_coordiantes.rb
```diff
@@ -0,0 +1,62 @@
+class FixWebsterGroveCoordiantes < ActiveRecord::Migration[6.0]
```
rileyanderson, line 1: It is not best practice to run data migrations this way. These migrations are for database structure/schema changes only, not database records. Please move this to a rake task and run it in the environments where you need to change this data. See here for more info [Data Migrations](https://depo-platform-documentation.scrollhelp.site/developer-docs/vets-api-database-migrations.689832034.html#vets-apidatabasemigrations-DataMigrations)

rileyanderson (changes requested):

db/migrate/20211130190839_fix_webster_grove_coordiantes.rb
PJart: I moved the migration to a rake task

rileyanderson (approved):

PR #871, opened 2021-12-16
Regocode physctry present task

PJart:
Tweek to rake task to regeocode issue coordinates within the institutions model


thilton-oddball (dismissed):
You are likely going to need to do the Rails and Brakeman updates before this will pass. Rails needs to go to 6.1.4.4 and Brakeman to 5.2.0.

Gemfile
```diff
@@ -75,7 +76,7 @@ group :development, :test do
   gem 'rubocop', '0.82.0', require: false
   gem 'rubocop-rails', '2.5.2'
   gem 'rubocop-rspec', '1.38.1'
-  gem 'scss_lint', require: false
+  gem 'scss_lint', '0.59.0', require: false
```
LindseySaari, line 79: Why the specific version here?

config/application.rb
```diff
@@ -41,6 +41,8 @@ class Application < Rails::Application
     config.autoload_paths += Dir["#{config.root}/lib/**/"]
     config.autoload_paths += Dir["#{config.root}/app/utilities/**/"]
 
+    config.action_dispatch.return_only_media_type_on_content_type = false
```
LindseySaari, line 44: Is this a change that belongs with this set of changes?

lib/tasks/fix_coord_mismatch.rake
```diff
@@ -7,22 +7,31 @@ task fix_coord_mismatch: :environment do
...
-      # checks first addresss
-      geocoded = Geocoder.coordinates("#{r.address}, #{r.city}, #{r.state}, #{r.zip}, #{r.country}")
+      address = parse_add(result, result.address)
+      address2 = parse_add(result, result.address_2)
+      address3 = parse_add(result, result.address_3)
+      geocoded = Geocoder.coordinates(address)
+      geocoded2 = Geocoder.coordinates(address2)
+      geocoded3 = Geocoder.coordinates(address3)
       if geocoded.present?
-        # if present will update record, else will check address_2 and adress_3 fields
```
LindseySaari, line 14: It looks like the logic here changed up a bit. It looks like _before_ if geocoded.present?, it would do an update, otherwise it would update geocoded2 AND geocoded3 if present?

Now there's a few more conditionals and the logic is different. I just want to check that this new is correct

PR #877, opened 2021-12-27
refactor of coordinate script, added script to fix issues with phys_c…

PJart:
Refactor of geocoder rake task to update coordinates, added a task to fix issues where the physical_country is outside of the USA and address is geocoded to US address.


rileyanderson:
@PJart Can this be closed?

PR #882, opened 2022-01-04
modified manifest.js file

PJart:
change to manifest.js to fix error in rails 6.1.4.4 upgrade

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,3 +1,3 @@
 //= link_tree ../images
-//= link_directory ../javascripts .js
-//= link_directory ../stylesheets .css
+//= link application.css
+//= link application.js


app/assets/config/manifest.js
```diff
@@ -1,3 +1,3 @@
 //= link_tree ../images
-//= link_directory ../javascripts .js
-//= link_directory ../stylesheets .css
\ No newline at end of file
+//= link application.css
```
LindseySaari, line 2: Have you tested this change as of yet?

LindseySaari: 🤞 Hoping this fixes this issue though! It seems like the correct fix according to the [docs](https://github.com/rails/sprockets/blob/8fc492abde6b20afbe6fa5d5aa7d1085aafd3527/UPGRADING.md#manifestjs):

"You may also find that some files that were not previously compiled as top-level targets are now. For instance, if your existing app has any js files directly at ./app/assets/javascripts or css/scss files ./app/assets/stylesheets, Rails with Sprockets 4 will now compile them as top-level targets. Since they were not previously treated as such, you probably don't mean them to be; if they are .scss partials referencing variables meant to be defined in other files, it may even result in an error message that looks like Undefined variable: $some_variable."

PJart: To the extent that I could, I am developing locally so I ran the rake ci command locally and ran the project locally to do some manual tests, I didn't see any test failures and didn't see anything in the logs. This issue this fixes seems to happen after we merge and it passes the initial ci and gets stuck in a review instance.

LindseySaari (approved):

PR #884, opened 2022-01-05
Application css font face mistype

PJart:
fixed misquote on font-face class in application.css file


LindseySaari (approved):
👏 Let's give it a go!

PR #903, opened 2022-03-11
schedule.rb changes

PJart:
update to schedule.rb, found some issues from testing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ COPY --chown=gi-bill-data-service:gi-bill-data-service . .
 USER gi-bill-data-service
 RUN gem install bundler --no-document -v ${BUNDLER_VERSION}
 RUN bundle install --binstubs="${BUNDLE_APP_CONFIG}/bin" $bundler_opts && find ${BUNDLE_APP_CONFIG}/cache -type f -name \*.gem -delete
-RUN bundle exec whenever --update-crontab --set environment=test
+RUN bundle exec whenever --update-crontab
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -47,6 +47,7 @@ COPY --chown=gi-bill-data-service:gi-bill-data-service . .
 USER gi-bill-data-service
 RUN gem install bundler --no-document -v ${BUNDLER_VERSION}
 RUN bundle install --binstubs="${BUNDLE_APP_CONFIG}/bin" $bundler_opts && find ${BUNDLE_APP_CONFIG}/cache -type f -name \*.gem -delete
+RUN bundle exec whenever --update-crontab --set environment=test
```
ericboehs, line 50: I believe this should be the production environment, yes?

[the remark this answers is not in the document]
PJart: I updated it taking out the env

PJart: I updated it taking out the env

zachclarity: @rileyanderson  can you take a look at this.

zachclarity: @PJart  see comments above.

PJart: I updated the dockerfile to set it to production, by just setting it to bundle exec whenever --update-crontab, it sets the crontab to run without a specified env. The schedule commands run the commands to run the job.

PJart: just to test it I can set the crontab in staging as well, we just need to test that the cron job can be run with the changes I made to schedule.rb

rileyanderson (approved):

PR #906, opened 2022-03-22
Late payments issue update

PJart:
Update to issue seen with late payments not being displayed correct. Needed to add facet in controller to display and tweeked the InstiutionBuilder to only show yes or no for the sec_103_message

### DIFF
--- a/db/migrate/20220322154645_add_default_sec_103_message.rb
+++ b/db/migrate/20220322154645_add_default_sec_103_message.rb
@@ -1,5 +1,5 @@
 class AddDefaultSec103Message < ActiveRecord::Migration[6.1]
   def change
-  	change_column_default(:institutions, :section_103_message, from: nil, to: "No")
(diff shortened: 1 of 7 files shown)


db/migrate/20220322154645_add_default_sec_103_message.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultSec103Message < ActiveRecord::Migration[6.1]
+  def change
+  	change_column_default(:institutions, :section_103_message, from: nil, to: "No")
+  end
```
rileyanderson, line 4:
```suggestion
  def change
    change_column_default(:institutions, :section_103_message, from: nil, to: "No")
  end
```

PJart: does this need to be changed?

PJart: I think its changed now

rileyanderson (approved):

PR #914, opened 2022-05-10
Add geo text search coord script

PJart:
Added call to full text search in geocoder gem to update institutions with bad addresses.


lib/tasks/fix_coord_mismatch.rake
```diff
@@ -54,32 +52,78 @@ def geocode_fields(result, address, address1, address2, address3)
...
+def return_num_text_search(result, geo, city, state, num)
+  text_search = []
+  add_num = result.institution.downcase.split('#').last.to_i
+  text_search << geo if city == result.city.titleize && state == result.state && num == add_num
+end
+
+def return_text_search(result, geo, city, state, zip)
+  text_search = []
+  vil = geo.data['address']['village']
+  if city == result.city.titleize || state == result.state || zip == result.zip || vil == result.city.titleize
```
ericboehs, line 125: If this conditional isn't met should you return an empty array here instead of nil?

lib/tasks/fix_coord_mismatch.rake
```diff
@@ -54,32 +52,78 @@ def geocode_fields(result, address, address1, address2, address3)
...
 end
 
-def parse_add(res, address)
-  if address.present?
-    "#{address}, #{res.city}, #{res.state}, #{res.zip}, #{res.country}"
-  else
-    "#{res.city}, #{res.state}, #{res.zip}, #{res.country}"
+def text_search(result, geocode_name)
+  geocode_name.each do |geo|
+    next if ck_srch_fields(result, geo).blank?
```
ericboehs, line 97: You should memoize ck_srch_fields since you're calling it multiple times.

lib/tasks/fix_coord_mismatch.rake
```diff
@@ -6,15 +6,14 @@ task fix_coord_mismatch: :environment do
   results = Institution.approved_institutions(version)
```
ericboehs, line 6: This rake file is getting a bit large. It would make sense to extract the logic into a class so that it could be tested easier.

ericboehs (changes requested):

app/models/geocode_logic.rb
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  def return_num_text_search(result, geo, city, state, num)
+    text_search = []
+    add_num = result.institution.downcase.split('#').last.to_i
+    text_search << geo if city == result.city.titleize && state == result.state && num == add_num
+    text_search
+  end
+
+  def return_text_search(result, geo, city, state, _num)
```
jhwuz, line 86:
```suggestion
  def text_search(result, geo, city, state)
```

jhwuz: Try not to start method names with `return` for readability. Changing the name to `text_search` will still imply that it returns the same result.
 
Is `num` needed here? It's not ever used inside of the method.

app/models/geocode_logic.rb
```diff
@@ -0,0 +1,118 @@
...
+    if geocoded.present?
+      update_mismatch(result, geocoded)
+    elsif geocoded1.present?
+      update_mismatch(result, geocoded1)
+    elsif geocoded2.present?
+      update_mismatch(result, geocoded2)
+    else
+      check_bad_address(result, geocoded3)
+    end
+  end
```
jhwuz, line 18:
```suggestion
  def geocode_fields(result, addresses)
    geocoded_result = Geocoder.coordinates("#{result.city}, #{result.state}, #{result.zip}")
    addresses.each do |address|
      geocoded = Geocoder.coordinates(address)
      if geocoded.present?
        update_mismatch(result, geocoded)
      else
        check_bad_address(result, geocoded_result)
      end
    end
  end
```

jhwuz: Passing in the addresses as an array could help shorten and clean up this method.

app/models/search_geocoder.rb
```diff
@@ -0,0 +1,34 @@
...
+    @country = results.where.not(physical_country: 'USA')
+  end
+
+  def process_geocoder_address
+    by_address.each do |result|
+      address = parse_add_fields(result, result.address)
+      address1 = parse_add_fields(result, result.address_1)
+      address2 = parse_add_fields(result, result.address_2)
+      geocode_fields(result, address, address1, address2)
+    end
```
jhwuz, line 18:
```suggestion
    by_address.each do |result|
      addresses = [  parse_add_fields(result, result.address),
                               parse_add_fields(result, result.address_1),
                               parse_add_fields(result, result.address_2) ]
      
      geocode_fields(result, addresses)
    end
```

app/models/search_geocoder.rb
```diff
@@ -0,0 +1,34 @@
...
+    country.each do |result|
+      if result.state.present? && result.physical_country.present?
+        geocoded_ct = Geocoder.coordinates(result.physical_country)
+        update_mismatch(result, geocoded_ct)
+      else
+        address = parse_address(result, result.address)
+        address1 = parse_address(result, result.address_1)
+        address2 = parse_address(result, result.address_2)
+        geocode_fields(result, address, address1, address2)
+      end
```
jhwuz, line 31:
```suggestion
      else
        addresses = [ parse_add_fields(result, result.address),
                                 parse_add_fields(result, result.address_1),
                                 parse_add_fields(result, result.address_2) ]
        geocode_fields(result, addresses)
      end
```

jeremyhwu (commented):
All of your methods in GeocodeLogic are public but some of them don't seem to be used outside the class. Maybe methods like `check_bad_address` and `search_bad_addr` should be private helper methods since they are only used within the class?

jeremyhwu (approved):

ericboehs (approved):

PR #928, opened 2022-08-17
fixed ownership name in api

PJart:
Needed to add ownership name to serializers, not showing in payload when testing. Also added an ownership model and ownership model archive fore future use.

### DIFF
--- a/spec/factories/institutions.rb
+++ b/spec/factories/institutions.rb
@@ -18,6 +18,7 @@ FactoryBot.define do
     approved { true }
     bad_address { false }
     high_school { false }
+    ownership_name { nil }
 
     trait :in_nyc do
       city { 'NEW YORK' }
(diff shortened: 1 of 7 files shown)


spec/factories/institutions.rb
```diff
@@ -18,13 +18,15 @@
     approved { true }
     bad_address { false }
     high_school { false }
+    ownership_name { nil }
```
RachalCassity, line 21: Is there a reason why this isn't ownership_name { 'test' } instead of redefining this variable with the same value?

PJart: Its just because of the test, it looks at shared_examples_for_exportable_by_version which compares a step that creates a new instance of institution and compares it to whats being imported from the csv. The ownership field is updated in another process so that is why it is nil.

RachalCassity (approved):

PR #929, opened 2022-08-24
Implemenent rating system updates

PJart:
Added ability to upload rating spreadsheet and calculates rating for institution


Gemfile.lock
```diff
@@ -445,9 +438,7 @@ GEM
     zeitwerk (2.6.0)
 
 PLATFORMS
-  ruby
```
RachalCassity, line 448: Gemfile.lock needs to be cleaned up.

PJart: i did but it is now breaking the build

RachalCassity (approved):

PR #930, opened 2022-09-02
fixed build deploy

PJart:
Fix to build deploy script, removed step that connects https://github.com/department-of-veterans-affairs/vsp-infra-application-manifests/tree/main/apps


.github/workflows/build-deploy.yaml
```diff
@@ -12,66 +12,4 @@ jobs:
         uses: actions/checkout@v2
       - name: Run Tests
         run: |
-          make ci        
-  build-docker-image:
```
ericboehs, line 16: Since it's not building or deploying it should be renamed to something indicated that it runs ci or tests.

ericboehs (changes requested):
Rename workflow please.

ericboehs (approved):

PR #1081, opened 2024-04-04
Remove whenever gem and unneeded cron stuff

ryan-mcneil:
## Description
Cron jobs are no longer used in gibct, and the `schedule.rb` file had already been commented out. This cleans up extra gems and dockerfile packages/commands

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/80059

## Testing done
- passing pipelines

## Definition of done
- [ ] Builds properly
- [ ] No remnants of cron/whenever


nfstern02 (approved):
Always a good thing when you can remove gems from the Gemfile that are not needed.

PR #1091, opened 2024-04-10
Replace import-va-certs script with local mirror

ryan-mcneil:
## Description
As described in the [README](https://github.com/department-of-veterans-affairs/platform-va-ca-certificate/blob/main/README.md) of [platform-va-ca-certificate](https://github.com/department-of-veterans-affairs/platform-va-ca-certificate), occasionally certs will be unreachable, so cloning this repo provides a local method of CA signing certs

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/79342


nfstern02 (approved):

nfstern02:
Let me know when you squash & merge so I can monitor.

Dockerfile
```diff
@@ -80,9 +80,13 @@ ENV PATH="/usr/local/bundle/bin:${PATH}"
...
 RUN whoami
 
-# Download VA Certs
-COPY ./import-va-certs.sh .
-RUN ./import-va-certs.sh
+# Clone platform-va-ca-certificate and copy certs
+WORKDIR /srv
+RUN git clone --depth 1 https://github.com/department-of-veterans-affairs/platform-va-ca-certificate && \
+    cp platform-va-ca-certificate/VA*.cer /tmp && \
+    /bin/bash platform-va-ca-certificate/debian-ubuntu/install-certs.sh
```
ericboehs, line 87: Should we remove the repo when we're done here?

ryan-mcneil: oh, sure. The intent was for this to be done in an earlier throwaway stage in my other dockerfile... Perhaps I can move this to the builder stage for this case? Nothing needs to be kept (including /tmp), right?

ryan-mcneil: JK moved it into the base instead to simplify permissions a bit. Ultimately it will probably land in an earlier stage, so this should be good for now

ericboehs (approved):

PR #1218, opened 2024-10-17
added models for L&C and exams

binq:
added ability to uplaod CSVs for L&C and Exams.


nfstern02:
It looks like you have some failing tests. I tried resubmitting the Jenkins a couple of times and it appears to be failing on some tests of yours.

PR #1336, opened 2025-03-03
EDM-480/async upload refactor

jefftmarks:
## Description
Revert async upload logic after it was determined large files were causing staging instances to terminate themselves. Fallback to plan b, of automating the multiple file upload process where it looks like one request, but really large file is broken into smaller files on client side and each processed as individual upload

## Original issue(s)
https://jira.devops.va.gov/browse/EDM-593


binq (approved):
LGTM

PR #1337, opened 2025-03-04
EDM-480/revert async upload migrations

jefftmarks:
Remove columns on uploads table that were added when implementing async upload


binq (approved):
LGTM

PR #1341, opened 2025-03-05
EDM/disable no store for lcpe

jefftmarks:
Remove 'no-store' from 'Cache-Control' header when handling lcpe/lac or lcpe/exam request to enable proper caching with etag


binq (approved):
LGTM

PR #1349, opened 2025-03-11
EDM/reduce upload size and allow retries

jefftmarks:
Refactor sequential upload to allow for retries if one of the uploads fails. In addition, increase chunk size of individual uploads to secure against timeout issues


binq (approved):
LGTM

PR #1359, opened 2025-03-20
EDM-852/remove special char from enriched id

jefftmarks:
Change enriched idea delimiter from '@' to 'v' (for version)
Update etag to use double quotes instead of single

Per staging review: https://github.com/department-of-veterans-affairs/va.gov-team/issues/105704
https://jira.devops.va.gov/browse/EDM-852


binq (approved):
LGTM

PR #1390, opened 2025-04-22
VEBT-1390/dynamically fetch IPEDS by most recent

jefftmarks:
## Description
Replace hardcoded download sources for IPEDS files with logic that dynamically grabs most recent IPEDS-type file from [this page](https://nces.ed.gov/ipeds/datacenter/DataFiles.aspx?year=-1)

**Added HTTParty gem** which handles redirects (necessitated by IPEDS link above) more smoothly. Otherwise have to manually handle redirects with Net::HTTP it looks like

## Original issue(s)
https://jira.devops.va.gov/browse/VEBT-1390


Gemfile
nfstern02, line 1: Can you use the Faraday gem instead of httpparty? Faraday is already installed.

nfstern02 (commented):
Where are the specs? Need to see specs that clearly show how the branches changed work and an example where there are multiple links on the site and it picks the most recent one. You can use mocking and/or stubbing for the external call.

app/controllers/dashboards_controller.rb
nfstern02, line 1: **conversion.try(:call, @downloader.url)** || conversion || "tmp/#{params[:csv_type]}s.csv"

Is there a more understandable way to express this?

Gemfile
jefftmarks: Looks like we need to update faraday to 2.x in order to use the faraday-follow_redirects gem. We can't update faraday though because it's not compatible with faraday_middleware gem. We actually don't use the middleware gem but govdelivery-tms requires it. The govdelivery gem seems like action mailer gem which maybe isn't in use....... long story short we might need to tackle updating faraday separately. Keeping httparty after we discussed

app/controllers/dashboards_controller.rb
jefftmarks: No longer necessary. Removed

GcioGregg (approved):

PR #1401, opened 2025-05-07
VEBT-1391 add dynamic download source for vsoc

thomaskrenzke:
## Description

Following on to the work done by Jeff in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/1390, this PR adds a dynamic source for Vsoc data. It follows the same pattern and fetches some html, then parses the html to find the latest url for downloading vsoc data.

### DIFF
--- a/app/utilities/no_key_apis/vsoc_download_source.rb
+++ b/app/utilities/no_key_apis/vsoc_download_source.rb
@@ -4,8 +4,6 @@ module NoKeyApis
   class VsocDownloadSource < DynamicDownloadSource
     URL = 'https://vbaw.vba.va.gov/education/job_aids.asp'
 
-    CACHE_KEY = 'vsoc_html'
-
     def initialize
       super(URL)
     end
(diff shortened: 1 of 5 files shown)


app/utilities/no_key_apis/vsoc_download_source.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module NoKeyApis
+  class VsocDownloadSource < DynamicDownloadSource
+    URL = 'https://vbaw.vba.va.gov/education/job_aids.asp'
+
+    CACHE_KEY = 'vsoc_html'
```
jefftmarks, line 7: I know we discussed whether or not to cache vsoc so just want to confirm you'll be caching it. Also I set the expires_in TTL in the parent class sort of arbitrarily to one hour. After talking with Gregg and Noah they suggested each subclass could define it individually. What do you think? Could be implemented here or saved for later

thomaskrenzke: Ah, good call. I don't need caching for Vsoc since it's only really fetching one piece of data from a single html page. Cache key removed.

As for the per-class expires_in, that makes sense to me, but can likely wait until a later date when we have more examples of cacheable html.

nfstern02 (approved):
Looks like you (Tom) opted not to use caching at all.

PR #1429, opened 2025-06-10
VEBT 2213 institution public archives

thomaskrenzke:
## Description

There is a need to make the data used to power the comparison tool easily accessible to the general public. Ideally, it should be as simple as going to the comparison tool and clicking a link to download a file. The approach used in this PR is to create a new model, `VersionPublicExport`, which will be created when a new version is built (via the `InstitutionBuilder`). This new model will take all the serialized data for each active institution and zip it into a binary blob that can be stored in the database and served on request.

Another approach would be to write all of this data to a file and find some convenient public place to host the file (e.g. S3). But this repo doesn't currently have any S3 type stuff set up, so that would be a lot more work. Hopefully the current approach is a good tradeoff between performance (the data only needs to be compiled once when the version is generated, not every time a user wants to download it), size (the data blob being stored in the database is in the 30Mb range), and complexity (it reuses most of the existing institution builder patterns).

## Original issue(s)
[VEBT-2213](https://jira.devops.va.gov/browse/VEBT-2213)

## Testing done
Rspec, local

## Acceptance criteria
- [ ] New `VersionPublicExport` model is created and populated with correct data when a version is created
- [ ] `VersionPublicExport` data is downloaded via public url

## Definition of done


app/models/version_public_export.rb
```diff
@@ -0,0 +1,34 @@
...
+    zip_writer.close
+
+    record = find_or_initialize_by({ version_id: version.id })
+    record.update({ file_type: 'application/x-gzip', data: io.string })
+  end
+
+  def self.write_compressed_institution_data(version, writer, progress_callback: nil)
+    total_count = Institution.approved_institutions(version).count
+    i = 0
+    Institution.approved_institutions(version).limit(10).find_each do |institution|
```
thomaskrenzke, line 24: OOPs

nfstern02 (changes requested):
Drop limit 10 and someone will approve.

nfstern02 (approved):

PR #1494, opened 2025-08-08
Vebt 1682 yrp degree levels

thomaskrenzke:
## Description
We need to be able to translate the many strings used to describe the `degree_level` in the yellow ribbon program csv upload into a curated set of 'good' degree levels. E.g. an entry in the yellow ribbon csv upload might be "undregraduate/graduate" (sic) and we want to map that to the 'Undergraduate' and 'Graduate' degree levels. In this PR I've added a new model (`YellowRibbonDegreeLevelTranslation`) that handles this mapping: it stores a `raw_degree_level` as found in the csv file (downcased) and an array of known/good strings we want to use to describe degree levels. An new UI has been added to able to perform the necessary CRUD operations on this model. In addition, when a new yellow ribbon file is uploaded, we automatically scan for any new/unknown degree levels strings and take a best guess at mapping it to our curated levels.

Basic Flow:
0. Seed the `YellowRibbonDegreeLevelTranslation` table with a bunch of known values from the vets-website repo (one time only)
1. Someone uploads a yellow ribbon csv file. After upload, we see that a new degree level value (e.g. 'aas & professional doctoral') is present in the file, but we don't have a YellowRibbonDegreeLevelTranslation for it yet. We guess what it should map to and create the corresponding YellowRibbonDegreeLevelTranslation model
2. A version is generated, at which point we delete any YellowRibbonDegreeLevelTranslation models we no longer need. I.e. any which contain a `raw_degree_level` that is not present in any yellow ribbon program model.
3. Brian or whomever is able to go to /degree_levels and fiddle with any existing translation and assign whatever mapping they want
4. When the frontend uses our api, for each yellow ribbon program they get the `raw_degree_level` as they always have, in addition to an array of curated degree levels.

The final 'output' of this whole exercise is to include these curated degree levels in the `InstitutionProfileSerializer` so that the front end can more readily group yellow ribbon programs by degree level and avoid having to [do this](https://github.com/department-of-veterans-affairs/vets-website/blob/9cf7eaf8bcffb9f122e66544bc5bdca45c322c6e/src/applications/gi/constants.js#L446-L809)

This PR also add the `better_errors` gem, as it seems a bunch of errors (especially syntax errors in view templates) were not producing a backtrace and just showing the standard 500.html page.

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000
](https://jira.devops.va.gov/browse/VEBT-1682)
## Testing done
Rspec, local playing around

## Screenshots
Screenshot from 2025-08-08 12-35-14

Screenshot from 2025-08-08 12-35-43

## Acceptance criteria
- [ ]

### DIFF
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,1 +1,0 @@
-gibct-data-service
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,0 +1,1 @@
+gibct-data-service
--- a/Gemfile
+++ b/Gemfile
@@ -100,9 +100,9 @@ group :development do
   gem 'web-console', '~> 4.2', platforms: :ruby
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
+  gem 'better_errors'
+  gem 'binding_of_caller'
   gem 'spring', platforms: :ruby
-  gem "better_errors"
-  gem "binding_of_caller"
 end
 
 gem 'solid_queue', '~> 1.1'
--- a/Gemfile
+++ b/Gemfile
@@ -101,6 +101,8 @@ group :development do
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring', platforms: :ruby
+  gem "better_errors"
+  gem "binding_of_caller"
 end
 
 gem 'solid_queue', '~> 1.1'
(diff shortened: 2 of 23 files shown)


nfstern02 (commented):
Can you move the migration file into it's own request? Other than that, it looks good. I'm happy to see you're using turbo in your code.

nfstern02 (approved):

PR #1505, opened 2025-08-20
remove spec that makes real network call

thomaskrenzke:
## Description

One of the specs was making a real network request to http://example.com. This has been causing intermittent failures (especially during Jenkins CI builds). The test doesn't really seem to _do_ anything other than check for a 200. We could stub the network call with WebMock or VCR or similar, but it seems like removing such a trivial spec might be the easier way to go.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#0000

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/spec/lib/common/client/base_spec.rb
+++ b/spec/lib/common/client/base_spec.rb
@@ -104,11 +104,4 @@ describe Common::Client::Base do
       end
     end
   end
-
-  describe '#get' do
-    it 'returns status 200' do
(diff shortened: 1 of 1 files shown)


spec/lib/common/client/base_spec.rb
nfstern02, line 1: This seems like a lame test as it stands.

nfstern02 (approved):

PR #1508, opened 2025-08-22
VEBT-2483/convert version generation status to stimulus

jefftmarks:
## Description
Moving js script that reloads page to check for status updates during preview generation to a stimulus controller. Use polling and turbo streams instead of page reloads

**TO-DO**: convert the modal to bootstrap modal so we don't need to use jquery modal

FYI found typo where it was called Perform**Insitution**TablesMaintenanceJob so I fixed that in few spots as well (also removed Perform from job name because that seemed redundant and not rails-y)

## Original issue(s)
https://jira.devops.va.gov/browse/VEBT-2483


app/javascript/controllers/preview_status_controller.js
nfstern02, line 1: Consider changing res to response for clarity.

nfstern02 (approved):

PR #1521, opened 2025-08-29
VEBT-2483/integrate polling with solid queue

jefftmarks:
Due to transition to solid queue for active job adapter, status reporting during version generation was not behaving as expected (status update jobs all enqueued after entire institution builder transaction completes). Instead, dedicate separate db connection to PreviewGenerationStatusInformation model so that new statuses are committed instantly despite parent transaction


nfstern02 (approved):

nfstern02:
Looks good!

